PR #1, opened 2017-02-02
Add template feature

dominikschulz:
This PR adds a handy template / preset feature for your password store.

### DIFF
--- a/action/edit.go
+++ b/action/edit.go
@@ -8,11 +8,10 @@ import (
 	"os"
 	"os/exec"
 
-	"github.com/justwatchcom/gopass/pwgen"
-	"github.com/justwatchcom/gopass/tpl"
-
 	"github.com/justwatchcom/gopass/fsutil"
 	"github.com/justwatchcom/gopass/password"
+	"github.com/justwatchcom/gopass/pwgen"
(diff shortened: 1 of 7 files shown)


action/edit.go
```diff
@@ -8,6 +8,9 @@ import (
 	"os"
 	"os/exec"
 
+	"github.com/justwatchcom/gopass/pwgen"
+	"github.com/justwatchcom/gopass/tpl"
+
```
metalmatze: This new line is one too much.

action/edit.go
```diff
@@ -27,11 +30,21 @@ func (s *Action) Edit(c *cli.Context) error {
...
+	var changed bool
 	if exists {
 		content, err = s.Store.Get(name)
 		if err != nil {
 			return fmt.Errorf("failed to decrypt %s: %v", name, err)
 		}
+	} else if tmpl, found := s.Store.LookupTemplate(name); found {
+		changed = true
+		// load template if it exists
+		content = pwgen.GeneratePassword(24, false)
```
metalmatze: Maybe the default `24` should be a const shared across packages or passed down. Aren't we setting that somewhere else too?

dominikschulz: Yes, indeed. Will change that.

Issue #4, opened 2017-02-02
Add support to disable colorful output


metalmatze:
https://news.ycombinator.com/item?id=13552756

> I'm pretty excited about this actually. Thank-you so much for your efforts. I've been using pass for awhile now, and I really love what it does, but it's a case where it feels 90% finished.
I have one desperate request; colour output as an option. Every time there is an update to pass (or I need to reinstall) I need to edit the file and change the options from " tree -C " to " tree -n "  
This is a pain in the ass. I am visually impaired. The 'default' dark-blue that tree uses for directories is unreadable to me.
My two choices for dealing with this are to use DIRCOLORS or edit the pass executable. I'd prefer to not muck about with my environment settings. (as I do not normally see any colour output)
Anyway; awesome project!

This should be doable in the future with not that much refactoring.

dominikschulz:
We support doing `export GOPASS_NOCOLOR=true` to disable colors. Is that sufficient?

VA3FXP:
Yes, that will work.

dominikschulz:
Cool. Feel free to reopen if this isn't working as expected.

Miserlou:
I suggest that this feature isn't complete until it's a) available as part of the [standard config](https://github.com/justwatchcom/gopass#edit-the-config) b) documented. Currently it's neither.

dominikschulz:
Fixed by #33

PR #15, opened 2017-02-03
Adding ability for colour to be disabled

mrcrilly:
I've added the ability for colour to be disabled as per #4.

The config flag is simply `nocolor`. Note the lack of 'u' from my British English spelling. This was because I assumed the developers would likely be using American English.

Checking for the debugging environment variable was moved down a few lines for no particular reason at all. I moved it whilst trying to work out how-to integrate debugging into a config flag as well, but decided to just focus on the one flag instead. In the zone, I forgot to move it back to the top, but it has little difference.

For the `RootStore` struct key, I chose `NoColor` instead of `NoColors` (plural) to reflect the fact internally we're setting `color.NoColor` - I was looking to keep things consistent.

I felt the need to write the configuration to disk here on line 57 of `action.go`. When I was using the command to test what happened if I deleted my configuration and then checked the config, a new file wasn't written. This seemed odd to me, so I figured it was best we persistent what the user sees as I believe that would be a good user experience.

Slight typo in another file. Haven't checked all the code, though.

New config has been shown to work on my local compiled copy of `gopass`.

Versions:
- macOS 10.12
- gpg (GnuPG) 2.0.30 / libgcrypt 1.7.6 (gpg installed via brew)
- git version 2.11.0
- gopass HEAD (n/a ) go1.7.5

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 # gopass
 
+[![Build Status](https://travis-ci.org/justwatchcom/gopass.svg?branch=master)](https://travis-ci.org/justwatchcom/gopass)
 [![Go Report Card](https://goreportcard.com/badge/github.com/justwatchcom/gopass)](https://goreportcard.com/report/github.com/justwatchcom/gopass)
 
 The slightly more awesome Standard Unix Password Manager for Teams. Written in Go.
--- a/README.md
+++ b/README.md
@@ -242,6 +242,8 @@ path: /home/tex/.password-store
 persistkeys: false
 ```
 
+Note that if a configuration file isn't present on disk, it will be written when you first run `gopass`.
+
 ### Managing Recipients
 
 You can list, add and remove recpients from the commandline.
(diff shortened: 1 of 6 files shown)


mrcrilly:
I'm going to assume those failures are unrelated? Looking at the output from TravisCI, the errors pertain to missing tests?

action/action.go
```diff
@@ -53,13 +52,16 @@ func New(v string) *Action {
 	cfg.ImportFunc = askForKeyImport
 	cfg.FsckFunc = askForConfirmation
 	cfg.Version = v
+
+	// write the config to disk for persisten
+	writeConfig(cfg)
```
dominikschulz: We discussed that prior to release already and agreed to only write the config to disk when running "gopass init" or changing any config values, so I'd have to ask to to remove this line.

README.md
```diff
@@ -242,6 +242,8 @@ path: /home/tex/.password-store
 persistkeys: false
 ```
 
+Note that if a configuration file isn't present on disk, it will be written when you first run `gopass`.
```
dominikschulz: Please remove.

mrcrilly:
Changes made.

dominikschulz:
The patch itself looks good, but the branch still seems to include unrelated commits. Could you please rebase onto master or cherry-pick your commits onto a new branch?

mrcrilly:
Hi,

The results you see here are the results of me doing: `git rebase develop master`. Did I do it wrong?

I can create a new branch and cherry pick commits over to that. Cheers.

mrcrilly:
I'm closing this and starting a new one. I'm confused as to how to fulfil the requests being made here, so I need to start from a new branch in order to comply.

PR #16, opened 2017-02-03
Add the packaging metadata to build the gopass snap

elopio:
This is a package for the secure installation of apps that works in most Linux distributions.
Landing it upstream will enable builds that then can be distributed to many users through the Ubuntu store.


elopio:
You can find more information about snapcraft here: http://snapcraft.io/

To test it in an Ubuntu 16.04 machine:

    $ sudo apt install git snapcraft
    $ git clone --recursive https://github.com/elopio/gopass
    $ cd gopass
    $ git checkout snapcraft
    $ snapcraft
    $ sudo snap install *.snap --dangerous --classic

(dangerous because the snap you build yourself doesn't come signed from the trusted store. classic to not use the confinement for now. Once you push it to the store, your users will just have to do $ sudo snap install gopass)

This is an interesting use case for us. Moving from classic to strict confinement would require the users to trust you with their gpg keys. Which is something that should be possible, our philosophy is that we should make it easier from the users to trust in the upstream developers that are providing the software. But the users should still be in control, and it should be clear to what they are agreeing. It would be great to have your opinions in this discussion.

If you have any questions or comments, please let me know.

elopio:
Hey @dominikschulz ! thanks for checking it out.

It's not necessary to have the metadata in master. However, the workflow and tools are optimized to get a continuous delivery from upstream developers directly to users. With the yaml in this branch it would be really simple to configure something like travis to push to the edge channel in the store every time you make a change in master. Then your users in edge will get an automatic update and will be able to give you early feedback before you make a stable release. You get that and many other things just with a simple yaml file.

snap/snapcraft.yaml
```diff
@@ -0,0 +1,20 @@
...
+  gopass:
+    command: gopass
+    plugs: [home]
+
+parts:
+  gopass:
+    source: .
+    plugin: go
+    go-importpath: github.com/justwatchcom/gopass
+    stage-packages: [gnupg]
```
dominikschulz: I think we need `git` in here as well.

elopio: Done.

elopio:
I've moved the yaml to dist/snap. Just take into account that now snapcraft has to be called from dist/snap.

Thanks a lot for the review. Please share your thoughts about snaps :D

dominikschulz:
Thank you. We will test the snap package and merge this if all works well.

dominikschulz:
@elopio Was just testing your PR. Unfortunately I can't seem to build the snap on Ubuntu 16.04. It fails with an error like this:
```
Pulling gopass 
go get -t -d ./github.com/justwatchcom/gopass/...
package github.com/justwatchcom/gopass/action: /home/user/go/src/github.com/justwatchcom/gopass/dist/parts/gopass/go/src/github.com/justwatchcom/gopass exists but /home/user/go/src/github.com/justwatchcom/gopass/dist/parts/gopass/go/src/github.com/justwatchcom/gopass/.git does not - stale checkout?
package github.com/justwatchcom/gopass/fsutil: cannot find package "github.com/justwatchcom/gopass/fsutil" in any of:
        /usr/lib/go-1.6/src/github.com/justwatchcom/gopass/fsutil (from $GOROOT)
        /home/user/go/src/github.com/justwatchcom/gopass/dist/parts/gopass/go/src/github.com/justwatchcom/gopass/fsutil (from $GOPATH
[...]
Command '['/bin/sh', '/tmp/tmp3ni27wzn', 'go', 'get', '-t', '-d', './github.com/justwatchcom/gopass/...']' returned non-zero exit status 1
```

elopio:
@dominikschulz it seems that you are running snapcraft from a weird git directory.

I've updated the steps to build in my comment above to make sure that your repo is clean, and to take into account the new location of the yaml in dist. I've also rebased with the recent master changes. If that doesn't work for you, please give me the steps you are following so I can try to reproduce in my 16.04.

pura vida.

dominikschulz:
Sorry, still no luck building it and we're not sure we want this add complexity in. Not merging for the time being.

elopio:
That's sad, but thanks for trying it.

It would be very useful for us to understand what went wrong in your environment, to support it too. So if in the future you want to get gopass in the Ubuntu store, please let me know. We are also close to get https://build.snapcraft.io/ out of the beta, so in the future building and publishing in a clean and reproducible environment will just require a couple of clicks.

PR #18, opened 2017-02-03
Accepting args for EDITOR command

franciscocpg:
First of all I'd like to say that `gopass` is pretty awesome.

### About this PR
So i've been playing a while with `gopass` and at a certain time I tried to edit my passwords using some GUI tools like Sublime (`subl`) and Visual studio code (`code`).

By default these tools run in background. Fortunately both has the `-w` option which instructs them to "wait" for process. So one can set `EDITOR` variable like `EDITOR="subl -w"` or `EDITOR="code -w"`.

But in doing so and running `gopass edit somepassword` causes the error `Error: failed to run subl -w with /dev/shm/gopass-editsomehash file`.
That's because currently `gopass` is using the `EDITOR` variable content as the `name` argument on call to `exec.Command`. This PR solves this by splitting the `EDITOR` variable content to pass the command to the `name` argument appending the arguments from `EDITOR` command along with `tmpfile.Name()`.


action/edit.go
```diff
@@ -69,7 +70,10 @@ func (s *Action) editor(content []byte) ([]byte, error) {
 		return []byte{}, fmt.Errorf("failed to create tmpfile to start with %s: %v", editor, tmpfile.Name())
 	}
 
-	cmd := exec.Command(editor, tmpfile.Name())
+	cmdArgs := strings.Split(editor, " ")
```
metalmatze: In the `tests` package we use https://godoc.org/github.com/kballard/go-shellquote#Split to do this. We should probably do the same here.

metalmatze:
Please also update the target branch in this PR to `master` as we should add this as a bug fix to the 1.0.1 release. Thanks! :blush:

action/edit.go
franciscocpg: @metalmatze 
I changed it accordingly and because https://godoc.org/github.com/kballard/go-shellquote#Split returns an error I added the error message ``failed to parse EDITOR command `%s` `` if error not nil.
So for an given invalid command, eg `EDITOR="subl '"`, the error message is 
``Error: failed to parse EDITOR command `subl '` ``.
Do you like this approach?

franciscocpg:
@dominikschulz @metalmatze
I've made some mess rebasing and have to open another PR https://github.com/justwatchcom/gopass/pull/30.
Sorry for the inconvenience.

PR #21, opened 2017-02-03
Only copy the first line to the clipboard

### LINKED ISSUE #11: gopass copies entire file into clipboard
When you run `pass -c <filename>`, pass's behavior is to copy the first line of the file, allowing you to get the password easily while using the rest of the file in a freeform fashion. gopass seems to copy the whole contents of the file, which makes it hard to use multiple lines for things like login and URL.

metalmatze:
Closes #11.

I don't think we need a buffer to get the first line, we already have everything in memory in this `func` anyway.


metalmatze:
Travis fails, because we don't have the `.travis.yml` on the master branch yet...

PR #47, opened 2017-02-08
Add dmenu support

### LINKED ISSUE #43: Support for dmenu
@notandy and me just talked and we need this for i3 & dwm, etc...

metalmatze:
Closes #43 

The passmenu works but doesn't support mounts.

This add the script into the binary, just like auto completion for bash & zsh.
Some parts are still WIP and are probably going to change.

Usage: `eval "$(pass completion dmenu)"`

What do you think @notandy & @fortytw2?

### DIFF
--- a/action/completion.go
+++ b/action/completion.go
@@ -51,3 +51,33 @@ source <(gopass completion bash)
 
 	return nil
 }
+
+// CompletionDMenu returns a script that starts dmenu
+// Usage: eval "$(pass completion dmenu)"
+func (s *Action) CompletionDMenu(
+	c *cli.Context) error {
+	out := `#!/usr/bin/env bash
+
+shopt -s nullglob globstar
+
+typeit=0
+if [[ $1 == "--type" ]]; then
+	typeit=1
+	shift
+fi
+
+password=$(gopass list --raw | dmenu "$@")
+
+[[ -n $password ]] || exit
+
+if [[ $typeit -eq 0 ]]; then
+	pass show -c "$password" 2>/dev/null
+else
+	pass show "$password" | { read -r pass; printf %s "$pass"; } |
+		xdotool type --clearmodifiers --file -
+fi
+`
+	fmt.Println(out)
+
+	return nil
+}
--- a/action/completion.go
+++ b/action/completion.go
@@ -1,7 +1,10 @@
 package action
 
 import (
+	"bytes"
 	"fmt"
+	"os/exec"
+	"strings"
 
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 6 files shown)


action/completion.go
```diff
@@ -51,3 +51,33 @@ source <(gopass completion bash)
...
+	typeit=1
+	shift
+fi
+
+password=$(gopass list --raw | dmenu "$@")
+
+[[ -n $password ]] || exit
+
+if [[ $typeit -eq 0 ]]; then
+	pass show -c "$password" 2>/dev/null
```
fortytw2: does `gopass` install as `pass`?

metalmatze: Nope, that's by mistake and worked for me because I `alias pass=gopass`... :smile: Thankfully still WIP and will be update, thanks for pointing out anyway! :blush:

fortytw2: I think it might make more sense overall to switch on `os.Args[0]` as `gopassmenu` or `passmenu` and just run `xdotool` from within the program

dominikschulz: `os.Args[0]` :+1:

action/list.go
```diff
@@ -8,8 +8,16 @@ import (
 
 // List all secrets as a tree
 func (s *Action) List(c *cli.Context) error {
+	raw := c.Bool("raw")
 	filter := c.Args().First()
 
+	// Don't show a tree only new lines
+	if raw {
+		//TODO(metalmatze): Support filtering
```
dominikschulz: You can do:
```
list, err := s.Store.List()
for _, v := range list {
 if smth(v) {
  fmt.Println(v)
 }
}
```

metalmatze:
I reworked the PR and got completely rid of the bash script. We simply don't need it.
It's now all within the binary itself.

Tested on my system. :blush:

action/completion.go
```diff
@@ -51,3 +54,47 @@ source <(gopass completion bash)
...
+		return err
+	}
+
+	content, err := s.Store.First(name)
+	if err != nil {
+		return err
+	}
+
+	if typeit {
+		return exec.Command("xdotool", "type", "--clearmodifiers", string(content)).Run()
```
metalmatze: So `xdotool` is a _fake keyboard/mouse input, window management, and more_. It is used to type the secret into the focused element on your screen. This can be a terminal, Slack Desktop App. Everything that has a keyboard input. It types the secret's chars into that field one after another.

action/show.go
```diff
@@ -20,36 +19,34 @@ func (s *Action) Show(c *cli.Context) error {
 		return s.List(c)
 	}
 
+	if c.Bool("clip") {
```
metalmatze: Not a first sight, but clip and dmenu use the clipboard in the background. We only want to copy the first line of the secret into the clipboard with `gopass show -c` and also with `gopass completion dmenu`. So instead of doing the first line extraction inside the `copyToClipboard` func is moved it to the store with a `func First(name string)` that returns only the first line of the secret.

password/root_store.go
```diff
@@ -303,6 +304,21 @@ func (r *RootStore) Get(name string) ([]byte, error) {
 	return store.Get(strings.TrimPrefix(name, store.alias))
 }
 
+// First returns the first line of the plaintext of a single key
```
metalmatze: Please see the other comment.

PR #50, opened 2017-02-12
Rename LDFLAGS to GOLDFLAGS

### LINKED ISSUE #42: Rename Makefile LDFLAGS
Hi,

Gentoo and Arch inject custom LDFLAGS targeted for binutils-ld that will fail with the golang compiler, please rename the LDFLAGS variable in the Makefile to mitigate this.

Thanks :)

metalmatze:
Closes #42


metalmatze:
That's right. Fixed it.

PR #53, opened 2017-02-14
Create .codeclimate.yml

larkinscott:
Created a Code Climate config with gofmt, golint, and govet enabled.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,13 @@
+engines:
+ gofmt:
+   enabled: true
+ golint:
+   enabled: true
+ govet:
+   enabled: true
+   
+ratings:
+ paths:
+ - "**.go"
+ 
+exclude_paths:


metalmatze:
Are there any benefits by using codeclimate over our own Makefile approach which runs the same stuff?

larkinscott:
Hi @metalmatze! Good to meet you! Happy to list a few of the benefits of running Code Climate:

* Our build is a comparison. This allows us to show you only newly introduced and fixed issues (not all issues)
* You can view the results in GitHub as you're browsing PRs (and overlayed with test coverage results)
* We provide ratings for each file and let you know when grades have changed
* Code Climate provides you with the ability to use an "approvals-based" workflow
* Running locally is done easily with our CLI, which, in some instances, can negate the need to have contributors install a host of dependencies 
* Lastly, we provide the information needed to resolve the issue inline

Glad to dive a little deeper on any (or all) of the above points if you'd like. Appreciate you taking the time!

PR #64, opened 2017-02-18
Implement QR Code generation

### LINKED ISSUE #44: qrencode functionality
qr code generation for passwords, cool would be output on the console like here: https://www.npmjs.com/package/qrcode-console

dominikschulz:
Fixes #44


metalmatze:
Maybe in a following PR we can add `gopass --qr foo/bar` to the root action if one is missing that. But we can discuss this later.

PR #67, opened 2017-02-20
Use XDG_CONFIG_HOME for gopass config

### LINKED ISSUE #46: (linux) Use $XDG_CONFIG_HOME for gopass config yml
Since this is the default location for config file according to freedesktop configuration specification: 
https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

> $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used.

dominikschulz:
Fixes #46

### DIFF
--- a/action/action.go
+++ b/action/action.go
@@ -36,12 +36,14 @@ func New(v string) *Action {
 	pwDir := pwStoreDir("")
 
 	// try to read config (if it exists)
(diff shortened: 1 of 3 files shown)


metalmatze:
Reviewed on ArchLinux locally. LGTM

PR #74, opened 2017-03-02
Accept args for dmenu

### LINKED ISSUE #72: custom arguments for dmenu completion
When running ```gopass completion dmenu --type```, ```dmenu``` is called by gopass and by default it displays entries horizontally. I'd like it to be displayed vertically and for that I would need to be able to call ```dmenu``` with custom arguments, for example ```dmenu -i -l 20 -p '>'```.

We could maybe add a command line switch to specify arguments that would be passed to ```dmenu```. For example ```gopass completion dmenu --type --extra-dmenu-args '-i -l 20 -p \'>\''```.

metalmatze:
Closes #72

`gopass completion dmenu --args "-i -l 20 -p >"`

Should solve the problems you have, right @amietn?

### DIFF
--- a/action/completion.go
+++ b/action/completion.go
@@ -3,9 +3,11 @@ package action
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 
+	shellquote "github.com/kballard/go-shellquote"
 	"github.com/urfave/cli"
 )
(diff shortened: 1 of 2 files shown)


amietn:
Yes this looks like what I needed. Thanks!

PR #76, opened 2017-03-05
adding rofi completion

deadc0de6:
This is a simple copy of the dmenu block for rofi.

It will call ```rofi --dmenu``` to choose which entry to display


deadc0de6:
First time in go. Hope it helps

metalmatze:
Yes. I saw the duplication too. I will merge this and then try to work on making it DRY. Ok?

deadc0de6:
ok, thanks for your help !

metalmatze:
I'm not going to merge this as it it. We're currently thinking about making the approach more general, so it really doesn't matter if one uses dmenu or rofi, or whatever.

Issue #39, opened 2017-02-06
does not work when gpg-agent is enabled


FlorinAndrei:
New to gopass, just test driving it.
- macOS 10.12.3
- gopass 1.0.0 installed from binary tarball
- gpg 1.4.20 installed from brew
- gpg-agent 2.0.30 installed from brew
- git 2.9.2

I've configured gpg-agent to start automatically in `~/.bash_profile`:

```
[ -f ~/.gpg-agent-info ] && source ~/.gpg-agent-info
if [ -S "${GPG_AGENT_INFO%%:*}" ]; then
  export GPG_AGENT_INFO
else
  eval $( gpg-agent --daemon --write-env-file ~/.gpg-agent-info )
fi
```

```
$ cat ~/.gnupg/gpg-agent.conf 
use-standard-socket
default-cache-ttl 28800
max-cache-ttl 86400
```

Tested the agent, and it works. If I do `gpg -d path/to/some/encrypted/file` repeatedly, I only need to enter the GPG pass once.

I've installed gopass and done `gopass init`:

```
$ gopass init
Please select a private Key for encryption:
[0] 0xXXXXXXXXXXXXXXXX - Florin Andrei <florin@XXXXXXXXXXXXXXXX>
Please enter the number of a key (0-0) [0]: 
Password store initialized for: 0xXXXXXXXXXXXXXXXX - Florin Andrei <florin@XXXXXXXXXXXXXXX>

Please select a key for signing Git Commits
[0] 0xXXXXXXXXXXXXXXXX - Florin Andrei <florin@XXXXXXXXXXXXX>
Please enter the number of a key (0-0) [0]: 
Initialized empty Git repository in /Users/XXXXXXXXXXXXX/.password-store/.git/
[master (root-commit) e0dcdef] Add current contents of password store.
 2 files changed, 53 insertions(+)
 create mode 100644 .gpg-id
 create mode 100644 .gpg-keys/XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
[master d984775] Configure git repository for gpg file diff.
 1 file changed, 1 insertion(+)
 create mode 100644 .gitattributes
Git initialized
```

Then I've tried to insert some fake password, for testing:

```
$ gopass insert web/google.com
Enter password for web/google.com: 
Failed to restore terminal: errno 0
Retype password for web/google.com: 
Failed to restore terminal: errno 0
gopass: Encrypting web/google.com for these recipients:
[...]

Issue #13, opened 2017-02-02
Able to store secrets, but with errors


mrcrilly:
I saw `gopass` on Hacker News and thought I would check it out. I actually plan on doing a PR for #4, but I have experienced an issue I want to raise here. Perhaps someone can expand further on this.

```
$ gopass insert world
Enter password for world:
Failed to restore terminal: errno 0
Retype password for world:
Failed to restore terminal: errno 0
gopass: Encrypting world for these recipients:
 - 0xB058C2BF639F134B - Mike C (Hi) <mc@sent.com>

Do you want to continue? [Y/n]:
error: gpg failed to sign the data
fatal: failed to write commit object

Error: failed to committ files to git: exit status 128

$ gopass show world
hello
```

So I insert a secret and it seems to fail, but then I'm able to retrieve it. The secret is present in the `.password-store` as a `.gpg` file and is indeed encrypted. Any thoughts on the errors here?

**Versions:**
- macOS 10.12
- gpg (GnuPG) 2.0.30 / libgcrypt 1.7.6 (gpg installed via brew)
- git version 2.11.0
- gopass HEAD (n/a ) go1.7.5

metalmatze:
Please take a look at the FAQ. 
https://www.justwatch.com/gopass/faq/

I think that's exactly the problem, right?! 😊

ksophocleous:
@metalmatze the FAQ solves part of the problem, thanks for that.

but the error `Failed to restore terminal: errno 0` still remains

metalmatze:
I can confirm this bug. It happend to me right now on ArchLinux.

```
$ gopass insert foo
Enter password for foo: 
Failed to restore terminal: errno 0
Retype password for foo: 
Failed to restore terminal: errno 0
gopass: Encrypting foo for these recipients:
 - 0xB1C7DF661ABB2C1A - Matthias Loibl <mail@matthiasloibl.com>

Do you want to continue? [Y/n]: 
[master bd0cf1f] Save secret to foo.
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo.gpg
```

gzugmeyer:
The bug is the same y discover on #17 

Change the .git/config solve the probleme

hartzell:
I was seeing the same problem:

> Error: failed to committ files to git: exit status 128

(ps, #27 fixes the typo...)

[...]

PR #80, opened 2017-03-12
Fix for "Failed to restore terminal: errno 0"

hartzell:
Aka: Handle borked return value for terminal.Restore

Until recently (go 1.8 still seems to be doing it the old way) `terminal.Restore` returned an error that was non-go-like.

This resulted in this error message whenever the terminal was set to no-echo and then restored:

> Failed to restore terminal: errno 0

This error has been reported in #13 and #39 (and perhaps others).

`terminal.Restore` has actually been returning the `syscall.Errno` returned by `syscall.Syscall6(..)`.  When it's 0 everything was successful, otherwise there's a problem.

`terminal.Restore` is returning it as a value of type `error`, which makes it hard to test for zero-ness.

This inconsistent behavior has been [fixed in this commit][fix].  I still see this problem when I build gopass with go@1.8, so apparently the [fix] did not make the cutoff.  I'm not sure how to code something that will work consistently with the pre- and post-correction versions of terminal.Restore().

[fix]: https://github.com/golang/crypto/commit/c3b1d0d6d8690eaebe3064711b026770cc37efa3

### DIFF
--- a/action/clihelper.go
+++ b/action/clihelper.go
@@ -8,6 +8,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"syscall"
 
 	"github.com/justwatchcom/gopass/gpg"
 	"golang.org/x/crypto/ssh/terminal"
@@ -188,7 +189,11 @@ func promptPass(prompt string) (pass string, err error) {
 	}
 	defer func() {
 		if err := terminal.Restore(fd, oldState); err != nil {
-			fmt.Printf("Failed to restore terminal: %s\n", err)
+			if e, ok := err.(syscall.Errno); ok {
+				if e != 0 {
+					fmt.Printf("Failed to restore terminal: %s\n", err)
+				}
+			}
 		}
 	}()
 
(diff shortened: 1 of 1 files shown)


hartzell:
I've realized that this bit of the crypto library is not part of the main go distribution.

A more proper fix would be to update the vendored copy so that it includes the [fix for underlying problem](https://github.com/golang/crypto/commit/c3b1d0d6d8690eaebe3064711b026770cc37efa3).

I'm not sure how you're managing the vendor dir, so I'll either need a hint or I'll leave it to you.

zachgersh:
@hartzell - looks like `govendor` to me:

https://github.com/kardianos/govendor

metalmatze:
Yes, exactly. We use govendor. I've created a PR with the updated version of that dependency.

metalmatze:
This PR itself LGTM, updating the dependency shouldn't affect this change, looking at it from GitHub.

dominikschulz:
Looks good, but need to test this first.

hartzell:
If you have a different PR that updates the vendor'ed library that you should not use this change.  It's either or.

I *think* that the correct fix is to update the vendor'ed library.

metalmatze:
I can confirm that updating the vendored library indeed fixed the problem. So this workaround isn't needed anymore. Thanks for the help to resolve this issue nevertheless, @hartzell!

PR #81, opened 2017-03-12
Make `-c` clip *all* the secret, not first line

hartzell:
If you create a multiline secret using e.g. `gopass edit` and then show
it, you see all of the lines.

If you copy it to the clipboard via `-c`, only the first line gets
copied to the clipboard.

This makes with and without `-c` behave identically.

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -23,7 +23,7 @@ func (s *Action) Show(c *cli.Context) error {
 	}
 
 	if clip || qr {
-		content, err := s.Store.First(name)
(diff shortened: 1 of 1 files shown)


metalmatze:
That -c only copies the first line of a secret is a feature not a bug. That behavior is identical to the one in the original pass and needs to be kept this way. 
What we can talk about might be another flag that copies the whole secret and not the first line.

metalmatze:
https://github.com/justwatchcom/gopass/pull/21

PR #87, opened 2017-03-20
Improve mounts and init commands

### LINKED ISSUE #75: Not being able to use mount
Whenever I try to run `gopass mount` I get `Error: Entry is not in the password store`.
I already have a root store.
What are the steps to add a new store and mount?

dominikschulz:
This commit adds an option to init new stores at arbitrary locations
using gopass init --store /path.

It also tries to clarify the existing mount and clone behaviour.

Fixes #75


metalmatze:
I like the new way of handling with mounts! :+1: 
I'm good with merging, but please check: https://travis-ci.org/justwatchcom/gopass/jobs/212943643#L201

PR #89, opened 2017-03-20
Improve recursive deletes

### LINKED ISSUE #55: gopass rm -r fails if argument is a file
Actual:

```
gopass rm -f some/file

Are you sure you would like to recursively delete some/file? [y/N]: y

Error: Entry is not in the password store
```

Expected:

```
gopass rm -f some/file

Are you sure you would like to recursively delete some/file? [y/N]: y

Deleted some/file
```

### LINKED ISSUE #70: error messages on deleting empty folders
This could be possibly connected to #55 

deleting an empty folder with gopass rm -r yields an error message, but deletes the folder.

> $ pass edit test/test
> ...
> $ pass rm test/test
> Are you sure you would like to delete test/test? [y/N]: y
> $ pass rm -r test
> fatal: pathspec '/home/andy/.password-store/test' did not match any files
> 
> Error: failed to add files to git: exit status 128

dominikschulz:
This commit improves handling of 'rm -r' on files as well as
removing empty dirs.

Fixes #55
Fixes #70 

This will need some testing before release, as it may change `rm` in subtle ways.

### DIFF
--- a/password/git.go
+++ b/password/git.go
@@ -134,6 +134,9 @@ func (s *Store) gitAdd(files ...string) error {
 	if !s.isGit() {
 		return ErrGitNotInit
 	}
+	for i := range files {
+		files[i] = strings.TrimPrefix(files[i], s.path+"/")
+	}
 
 	args := []string{"add", "--all"}
 	args = append(args, files...)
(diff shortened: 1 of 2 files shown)


metalmatze:
I can only reproduce that #55 is fixed, but #70 seems to be a bug still.

dominikschulz:
Hmm, need to further investigate this for 1.0.3

Issue #100, opened 2017-04-20
gopass completion dmenu --type fails with passwords starting with a dash character


amietn:
Running 

```gopass completion dmenu --type``` 

and selecting a password entry where the encrypted password has a cleartext value starting with a "-" makes gopass crash and the password is never typed:

```
Error: exit status 1
```

PR #101, opened 2017-04-21
specify end of flags for xdotool

amietn:
We don't want passwords starting with a dash to be interpreted as xdotool arguments


amietn:
This fixes the issue mentioned in #100

PR #103, opened 2017-04-25
Adding an option to hide the passwords on show action

AnomalRoil:
When user-names or comments in password entries are often read, it may be **dangerous** to always display the passwords on the terminal/stdout, since terminal output is easily read and/or logged.

Before this PR, when one did `gopass foo`, if foo existed, the whole entry was displayed, first line included.
After this PR, when one does `gopass foo`, only "safe content" is displayed, that is all content excepted for the first line. It is thought to be the password, so this allows to easily read the notes we may have stored in our password entries, without exposing our password on stdout.

In order to display the password, one can now either set `gopass config safecontent false` or occasionally use the new `gopass show -f foo` flag. 

So, this PR add a setting to handle this: "safecontent", as well as a way to force current behaviour using the `-f` or `--force` flag, even when the setting is set to true.

The **current _default_ behaviour** is not changed to a more _sane_ and _secure_ one, since this setting is set to **false** by default, otherwise it would divert from pass default behaviour.
Dmenu and xdotool completion _is not impacted_ since it already uses `s.Store.First` to get its content.

**Unit tests** for the feature are provided and passing. 
Tests also now include the whole `show` action, including QRcode generation (although that one may be a bit much.)

PS: the `safecontent` default setting could also be changed to `true` in order to have more security by default, but this would not be like what pass does.


AnomalRoil:
I'm glad you liked it. :)

PR #109, opened 2017-05-17
If entry was not found start search

### LINKED ISSUE #106: Initiate search instead of displaying "Error: Entry is not in the password store"
If you execute `gopass asdf.com` but `asdf.com` is not in the password store it should instead execute `gopass search asdf.com`.

(And maybe print an error message if then still nothing was found, as `gopass search` normally doesn't do anything if nothing was found. This, however, would be slightly confusing.)

dominikschulz:
Fixes #106

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -48,7 +48,7 @@ func (s *Action) Show(c *cli.Context) error {
 		if err != password.ErrNotFound {
 			return err
 		}
-		color.Yellow("Entry '%s' not found. Starting search ...", name)
+		color.Yellow("Entry '%s' not found. Starting search...", name)
 		return s.Find(c)
 	}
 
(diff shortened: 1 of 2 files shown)


action/show.go
```diff
@@ -44,7 +45,11 @@ func (s *Action) Show(c *cli.Context) error {
 
 	content, err := s.Store.Get(name)
 	if err != nil {
-		return err
+		if err != password.ErrNotFound {
+			return err
+		}
+		color.Yellow("Entry '%s' not found. Starting search ...", name)
```
metalmatze: While using it I saw that there's one space to much.

```diff
- "Entry '%s' not found. Starting search ..."
+ "Entry '%s' not found. Starting search..."
```

Issue #112, opened 2017-05-18
Storing structured files


bumi:
The readme mentions "Storing structured files" as ToDo. Because I have not found any related issue I am opening this one. 
Are there already plans on how to support structured data like credit card information and still be compatible with `pass`? 

Storing more structured information is an extension I am currently looking for, not sure if my go skills will be good enough but would be happy to look into it.

dominikschulz:
The general idea is to support some kind of structured data, e.g. yaml, inside secrets. Of course you can do that right now, but the plan is to give easy read/write access to parts of these files so you can do e.g. `gopass my/credit/card number` on a file like

```
---
name: John Doe
Number: 1234 5678 9012 1234
CV2: 123
```

And get an output like `1234 5678 9012 1234`. Also this should support updates like `gopass insert my/credit/card name`.

In a regular `pass` you would just get the YAML content.

We will try to provide a timeline soon.

dominikschulz:
Implemented in #125

PR #114, opened 2017-05-22
Add support for YAML metadata parsing

bumi:
This allows loading a specified key from a as yaml formatted metadata part of a secret. (related issue: #112 )

## Usage: 

assuming the following secret:

```
password

yaml: mammal
name: Johne Doe
```

### Display the value of a specified key:
```bash
gopass show yaml-example --key [yaml key]
```
### Copy the value of a specified key:
```bash
gopass show yaml-example -k [yaml key] -c 
```

When the `-k/--key` option is given the first line (the password) of a secret will be ignored and the rest will be loaded using [simpleyaml](https://github.com/smallfish/simpleyaml/). The `--clip` and `--qr` flags are still supported and the value the `--key` will be used. 

Editing the metadata part is currently not supported. In my usecase editing would mostly be manual and yaml is easy enough to write. So I don't think that this is super important. 

I am opening this PR for discussion on how this feature should be implemented. 
This is my first piece of go code that I am writing... so I don't really know what I am doing yet :) 

## ToDo:

- [ ] support any type of value in the YAML (currently strings are assumed)

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -7,6 +7,7 @@ import (
 	"github.com/atotto/clipboard"
 	"github.com/fatih/color"
 	"github.com/justwatchcom/gopass/qrcon"
+	"github.com/smallfish/simpleyaml"
 	"github.com/urfave/cli"
 )
 
@@ -16,6 +17,7 @@ func (s *Action) Show(c *cli.Context) error {
 	clip := c.Bool("clip")
 	force := c.Bool("force")
 	qr := c.Bool("qr")
+	key := c.String("key")
 
 	if name == "" {
 		return fmt.Errorf("provide a secret name")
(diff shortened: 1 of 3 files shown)


dominikschulz:
Thanks a lot for the PR.

At first sight this looks like going into the right direction, but still lacking a lot in terms of UX. Will give this some more thought / comments soon.

bumi:
thanks, looking forward to your thoughts.

dominikschulz:
Closing in favour of #125 

Thank you very much for this PR. If this was your first attempt in Go it's definitely very good. We wanted to some things a little different and also include write support so I thought it would be easier if we just pick your approach and implement it directly.

bumi:
whooot! awesome! will check it out. thanks!

PR #121, opened 2017-06-08
Improve error message in case the directory to clone to already exists

flosell:
Trying `gopass clone --path` the first time, we were surprised by the error message: "~/foo is a directory? - yes, thats where we want to clone to!". 

This PR tries to make it a bit more clear what happens: that gopass expects a path that doesn't exist so it can create it.

### DIFF
--- a/action/clone.go
+++ b/action/clone.go
@@ -58,7 +58,7 @@ func (s *Action) Clone(c *cli.Context) error {
 
 func gitClone(repo, path string) error {
 	if fsutil.IsDir(path) {
-		return fmt.Errorf("%s is a directory", path)
+		return fmt.Errorf("%s is a directory that already exists", path)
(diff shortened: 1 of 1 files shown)


metalmatze:
That's the default git behavior you're describing. git needs a non existing folder to clone to. 
However we can still merge this to make it more obvious. 👍

flosell:
Well, we assumed it would create a new folder underneath the path specified, similar to how git creates a new folder in the current working directory if you aren't specifying a path. 
Not saying that's the best assumption but since we made this mistake, a better error message would have saved a minute of confusion :)

dominikschulz:
Thanks for the PR. We'll merge this.

I don't think we have strong opinions about the current behavior. We may consider doing it the way you suggest.

PR #125, opened 2017-06-13
Add YAML read and write support

dominikschulz:
Thanks to @bumi / #114 for some inspiration.

### DIFF
--- a/password/store.go
+++ b/password/store.go
@@ -1,10 +1,13 @@
 package password
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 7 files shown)


password/store.go
```diff
@@ -1,11 +1,14 @@
 package password
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"gopkg.in/yaml.v2"
```
metalmatze: Don't we want to use https://github.com/ghodss/yaml when using yaml? :thinking:

PR #136, opened 2017-06-15
Implement binary support

dominikschulz:
This PR adds a binary command with several subcommands that streamline
working with binary data. To retain compatability with other password
managers the data is store base64 encoded.

We may want to add confirmation interaction for some of the destructive operations in the future, but right now I wanted to keep this PR to the point. Those can easily and cleanly added later on.

### DIFF
--- a/action/binary.go
+++ b/action/binary.go
@@ -1,0 +1,185 @@
+package action
+
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/base64"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/fatih/color"
+	"github.com/justwatchcom/gopass/fsutil"
+	"github.com/urfave/cli"
+)
+
+const (
+	// BinarySuffix is the suffix that is appended to binaries in the store
+	BinarySuffix = ".b64"
+)
+
+// BinaryCat prints to or reads from STDIN/STDOUT
+func (s *Action) BinaryCat(c *cli.Context) error {
+	name := c.Args().First()
+	if name == "" {
+		return fmt.Errorf("need a name")
+	}
+	if !strings.HasSuffix(name, BinarySuffix) {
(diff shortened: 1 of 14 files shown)


action/binary.go
```diff
@@ -0,0 +1,185 @@
...
+	default:
+		return fmt.Errorf("ambiquity detected. Unhandled case. Please report a bug")
+	}
+}
+
+func (s *Action) binaryValidate(buf []byte, name string) error {
+	h := sha256.New()
+	_, _ = h.Write(buf)
+	fileSum := fmt.Sprintf("%x", h.Sum(nil))
+
```
zwopir: I suggest extracting the hash summing, so we can change or extend it easily on one place

action/binary.go
```diff
@@ -0,0 +1,185 @@
...
+	var err error
+	buf, err = s.binaryGet(name)
+	if err != nil {
+		return err
+	}
+	_, _ = h.Write(buf)
+	storeSum := fmt.Sprintf("%x", h.Sum(nil))
+
+	if fileSum != storeSum {
+		return fmt.Errorf("Hashsum mismatch (file: %s, store: %s)", fileSum, storeSum)
```
zwopir: if we plan to expose this function (via wrapper) to the UI, we should return a custom error type, so we can handle error in retrieving a store secret and a validation mismatch differently

action/binary.go
dominikschulz: Hmm, I considered it, but in this case I'm just comparing two hashsums in one method. The hashum doesn't leave this method, so I'm inclined to leave it as is. We can easily change this later without affecting our API.

action/binary.go
dominikschulz: Indeed, but we'll have to do a big round of error handling / propagation refactoring anyway so I'd like to postpone this. I don't want to introduce one kind of error propagation here that may differ from the way we decide it later.

But definitely a good point, we should keep this in mind.

PR #144, opened 2017-06-16
Create secure in-memory directory on macOS

### LINKED ISSUE #29: `gopass edit` should create an in-memory temporary directory on macOS
MacOS doesn't have a `/dev/shm` directory so `gopass` falls back to `ioutil.TempDir` which will write plaintext to the disk. `pass` does a lot of work on [darwin] to avoid this.

[darwin]: https://git.zx2c4.com/password-store/tree/src/platform/darwin.sh#n18

dominikschulz:
Fixes #29

Note: Untested on MacOS so far. May need some tweaking.

### DIFF
--- a/fsutil/tempdir_darwin.go
+++ b/fsutil/tempdir_darwin.go
@@ -1,0 +1,93 @@
+// +build darwin
+
+package fsutil
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff"
+)
+
+func tempdirBase() string {
+	return ""
+}
+
+func (t *tempfile) mount() error {
+	// create 16MB ramdisk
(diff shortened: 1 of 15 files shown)


dominikschulz:
Tested on macOS. After a few small fixes it works fine for me.

fsutil/tempdir_darwin.go
```diff
@@ -0,0 +1,93 @@
...
+package fsutil
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/cenkalti/backoff"
```
zwopir: we probably want to vendor github.com/cenkalti/backoff?

dominikschulz: Yes, definitely. Good catch.

dominikschulz:
Signal handling is IMHO out of scope for this PR, but we should have a look at that later.

PR #148, opened 2017-06-16
Fix sorting of mount points

### LINKED ISSUE #147: sort errors in gopass mounts
The output of `gopass mounts` is sometimes incomplete due to errors in the sort/tree handling.

dominikschulz:
Fixes #147


action/mount.go
```diff
@@ -31,7 +33,11 @@ func (s *Action) MountsPrint(c *cli.Context) error {
 		return nil
 	}
 	root := simple.New(color.GreenString(fmt.Sprintf("gopass (%s)", s.Store.Path())))
-	for alias, path := range s.Store.Mounts() {
+	mounts := s.Store.Mounts()
+	mps := s.Store.MountPoints()
+	sort.Sort(store.ByPathLen(mps))
+	for _, alias := range mps {
+		path := mounts[alias]
```
zwopir: this returns "" for the path, if alias isn't found within mounts. Shouldn't happen (tm), as s.Store.Mounts() and .MountPoints() use the same underlying map. Maybe anyways because of the (tm)?

dominikschulz: Unlikely, possible, but it will be caught by AddMount: https://github.com/justwatchcom/gopass/blob/master/tree/simple/folder.go#L200

zwopir: ok

PR #151, opened 2017-06-19
Add goreleaser config

dominikschulz:
This PR adds goreleaser to simplify our release process. Homebrew integration hasn't been fully tested so far.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,11 +20,11 @@ const (
 
 var (
 	// Version is the released version of gopass
-	Version string
+	version string
 	// BuildTime is the time the binary was built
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -20,11 +20,11 @@ const (
 
 var (
 	// Version is the released version of gopass
-	Version string
+	version string
```
metalmatze: Did you test that having them private still works? Not sure right now.

dominikschulz: Me neither, thats how goreleaser whats them. We should test that to be sure.

PR #166, opened 2017-06-26
support 2FA time based tokens

pakohan:


pakohan:
copy feature missing

pakohan:
I've added it, can you check?

Issue #35, opened 2017-02-06
Gopass options per-mount?


passcod:
It would be nice to be able to set options per-mount instead of only having them global. For example, `autopush` and `autopull` set to true on my local (root) store, but set to false on another mount.

dominikschulz:
This is definitely possible but we'd have to offset the added feature against the added complexity (code and UI-wise). We'll consider this.

PR #169, opened 2017-06-27
Avoid running pager if not set

### LINKED ISSUE #167: pager not found
after installing **gopass** on macOS I couldn't run it because of the following error.

```
$ gopass

Error: exec: "pager": executable file not found in $PATH
$
```

It's really hard to find out what this means, Google does not yield any good results since I didn't know what it's supposed to do. After reading the source I found out the binary name can be overwritten by the PAGER env variable, so I thought it's a bash builtin. This showed me that a program like `less` is meant. Then it's simple to run gopass. Nevertheless, not the best UX

dominikschulz:
This PR avoids starting a pager if $PAGER is not set or the, newly introduced,
config option nopager is true.

Fixes #167

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,12 @@ $ gopass cp emails/example.com emails/user@example.com
 
 ## Advanced Features
 
+### Auto-Pager
+
+Like other popular open-source projects `gopass` automatically pipe the output
+to `$PAGER` if it's longer than one terminal page. You can disable this behaviour
+by unsetting `$PAGER` or `gopass config nopager true`.
+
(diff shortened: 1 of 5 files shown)


pakohan:
LGTM, although it makes the code a little less readable that the flag is already inverted with the **no** prefix. Wouldn't it be possible to introduce **pager** default true?

dominikschulz:
Right now I'd like to keep it as is, but for #35 we'll need to clean up the kinda messy configs anyway. Would probably make sense to takle this there for all existing options.

Issue #99, opened 2017-04-16
Completion should respect escaping


SevereOverfl0w:
For example if your pass name contains a `(` then it should complete to `\(` not `(`

dominikschulz:
This should be fixed. Can someone confirm this @SevereOverfl0w @i1skn @metalmatze ?

SevereOverfl0w:
Just built against master. Works at least for the case mentioned above ^^

PR #178, opened 2017-06-30
Completion respects escaping

i1skn:
Related to #99.
Cause `compgen` is actually pretty bad in handling spaces, escaping part was implemented in `go`.

### DIFF
--- a/action/completion.go
+++ b/action/completion.go
@@ -2,9 +2,22 @@ package action
 
 import (
 	"fmt"
+	"regexp"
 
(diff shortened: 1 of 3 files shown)


dominikschulz:
Very nice, but one failing lint check and one failing integration test.

Could you have a look at those, please?

Let me know if you need any help with that.

i1skn:
@dominikschulz Updated the PR. Now all the checks are passed!

PR #180, opened 2017-07-02
Check for already existing entries, when add mount

i1skn:
When a user adds a new mount there is a possibility, that entry with this name already exist.
Example:
```
gopass generate test 24
gopass mount add test ~/.test-password-store
```
After that one could not reveal `test.gpg` in the root store, until mount test is unmount.

### DIFF
--- a/action/mount.go
+++ b/action/mount.go
@@ -63,7 +63,11 @@ func (s *Action) MountAdd(c *cli.Context) error {
 	if k := c.String("init"); k != "" {
 		keys = append(keys, k)
 	}
-	if err := s.Store.AddMount(c.Args()[0], c.Args()[1], keys...); err != nil {
+	alias, localPath := c.Args()[0], c.Args()[1]
+	if s.Store.Exists(alias) {
+		fmt.Printf(color.YellowString("WARNING: shadowing %s entry\n"), alias)
+	}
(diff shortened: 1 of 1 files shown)


dominikschulz:
Thanks for your PR. I'm not yet sure if should merge it.

When we designed `gopass` we modeled the `mounts` feature after regular Unix filesystem mounts. These act similar to how `gopass` does right now: If you mount a device to a mountpoint everything that was at the mountpoint before is invisible and inaccessible until the device is unmounted. To me this feels very natural and is a powerful feature.

But I admit that it may confuse some users. I think we should warn the users if they are attempting to `shadow` a mountpoint, but I wouldn't prevent it.

i1skn:
Hello, thanks for the detailed feedback! What do you think if we change to a warning and just tell a user that he/she shadowing another entry? Does that sound good?

dominikschulz:
Yes, if you change the error to a warning I'd merge it.

zwopir:
IMHO that sounds perfect. We shouldn't deny shadowing, as it allows users to build up their secrets tree individually. But a warning when shadowing is definitely a good thing.

metalmatze:
I approve @zwopir's comment! 👍

i1skn:
@dominikschulz Changed to a warning :)

dominikschulz:
Thanks a lot!

PR #184, opened 2017-07-09
README: correct .deb download URL

marthjod:
Hi, I came across a 404 while following along with the README. You could perhaps generally use a _/latest_ or _/stable_ URL so you don't have to change docs for version bumps?


dominikschulz:
I'm not sure if providing a `/latest` URL would help so much in the end, IMHO we should rather fix our release tooling.

Issue #179, opened 2017-06-30
gopass init fails to initialize git repo.


james-lawrence:
```
Initialized empty Git repository in /home/james/.password-store/.git/

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <james@localhost.localdomain>) not allowed
Failed to init git: exit status 128
```

maybe check git config first and ask what information is wanted for those fields? I like to initialize my repository information differently depending on the project so I don't set the global.

dominikschulz:
This should have been fixed by #68

PR #185, opened 2017-07-11
configure user.name, user.email before initial git commit (#179)

marthjod:
Avoid potential failure on initial git commit:
Prompt user for necessary git config values and
provide private key identity names/emails
as suggestions for convenience.

### DIFF
--- a/action/clihelper.go
+++ b/action/clihelper.go
@@ -197,8 +197,8 @@ func (s *Action) askForPrivateKey(prompt string) (string, error) {
 // git config user.name and git config user.email, respectively.
 func (s *Action) askForGitConfigUser() (string, string, error) {
 	var (
-		name string = ""
-		email string = ""
+		name string
+		email string
 	)
 
 	if !s.isTerm {
--- a/action/clihelper.go
+++ b/action/clihelper.go
@@ -216,16 +216,17 @@ func (s *Action) askForGitConfigUser() (string, string, error) {
 
 			if !s.isTerm {
 				return identity.Name, identity.Email, nil
-			} else {
-				useCurrent, err = s.askForBool(
-					fmt.Sprintf("Use %s (%s) for password store git config?", identity.Name, identity.Email), false)
-				if err != nil {
-					return "", "", err
-				}
-				if useCurrent {
-					return identity.Name, identity.Email, nil
-				}
 			}
+
+			useCurrent, err = s.askForBool(
+				fmt.Sprintf("Use %s (%s) for password store git config?", identity.Name, identity.Email), false)
+			if err != nil {
+				return "", "", err
+			}
+			if useCurrent {
+				return identity.Name, identity.Email, nil
+			}
+
 		}
 	}
 
--- a/action/clihelper.go
+++ b/action/clihelper.go
@@ -192,18 +192,19 @@ func (s *Action) askForPrivateKey(prompt string) (string, error) {
 	return "", fmt.Errorf("no valid user input")
 }
 
-// askForGitConfigUser will iterate over GPG private key identities and return one identity's name
-// and/or one identity's email address if selected by the user for being used as values for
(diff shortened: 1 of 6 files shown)


marthjod:
If this is a way to implement this, I would look into adding tests as well.

dominikschulz:
This looks pretty good, thank you!

If you could add some tests that would be very nice.

action/git.go
```diff
@@ -29,7 +29,19 @@ func (s *Action) gitInit(store, sk string) error {
 		}
 	}
 
-	if err := s.Store.GitInit(store, sk); err != nil {
+	// for convenience, set defaults to user-selected values from available private keys
+	userName, userEmail, err := s.askForGitConfigUser()
```
marthjod: **TODO** handle `err`

action/clihelper.go
```diff
@@ -192,6 +192,66 @@ func (s *Action) askForPrivateKey(prompt string) (string, error) {
...
+	keyList = keyList.UseableKeys()
+	if len(keyList) < 1 {
+		return "", "", fmt.Errorf("No usable private keys found")
+	}
+
+	for _, key := range keyList {
+		for _, identity := range key.Identities {
+			useCurrent = false
+
+			if s.isTerm {
```
dominikschulz: These nested loops are a little hard to read.
Could you try to unroll them a bit?

Maybe something like
```
if !s.isTerm {
  return identity.Name, identiy.Email
}
useCurrent, err := s.askForBool("Do you want to use name/email for password store x?")
if useCurrent {
  return identity.Name, identiy.Email, nil
}
```
Right now it looks like you could select Name and Email from different identities, which doesn't really make sense.

action/clihelper_test.go
```diff
@@ -0,0 +1,41 @@
...
+	keyList, err := s.gpg.ListPrivateKeys()
+	if err != nil {
+		t.Error(err.Error())
+	}
+
+	name, email, _ := s.askForGitConfigUser()
+
+	// tests cannot know whether keyList returned empty or not.
+	// a better distinction would require mocking/patching
+	// calls to s.gpg.ListPrivateKeys()
```
dominikschulz: For our integration tests we use a `canned` keyring with a fixed set of keys.

action/clihelper.go
marthjod: I thought this _might_ be an (although not very likely) scenario but it definitely simplifies things to not split identities up, agreed.

action/clihelper_test.go
marthjod: This keyring is used indirectly in [init_test.go#22](https://github.com/justwatchcom/gopass/pull/185/commits/ff0860ed75e819ce1c684070e46969e3c3900817#diff-9e21dcdd5a0a12483a17f3cd3057ef75R22), but I couldn't figure out how to also leverage it in the unit test here.

dominikschulz: Yes, that's probably too much overhead here. That's why we made part of the tests as integration tests.

action/clihelper.go
```diff
@@ -218,38 +214,22 @@ func (s *Action) askForGitConfigUser() (string, string, error) {
...
-					useCurrent, err = s.askForBool(fmt.Sprintf("Use %q as user name for password store git config?", identity.Name), false)
-					if err != nil {
-						return "", "", err
-					}
-					if useCurrent {
-						name = identity.Name
-					}
+			if !s.isTerm {
+				return identity.Name, identity.Email, nil
+			} else {
```
dominikschulz: Almost perfect, but could you please outdent the else branch? Better style.

```
if !s.isTerm {
 return ...
}
useCurrent, ...
```

dominikschulz:
@marthjod If you could rebase real quick, we can merge this. Thanks for your effort!

marthjod:
Done. I could also squash commits for a more concise history if you want.

dominikschulz:
Awesome, thank you very much!

PR #193, opened 2017-07-15
Add YAML document marker

### LINKED ISSUE #192: Can't access the first key/value pair within structured secrets
The scenario is that I want to create a secret with _just_ structured input, i.e. something like:

```yaml
some_secret: surprise
another_secret: even_more_surprises
```

Let's say we added this kind of structured secret under `secret/structured_secret`, when I run gopass to read it I get the following:

```
$ gopass secret/structured_secret some_secret

Error: key not found
```

However, accessing the second key works perfectly fine:

```
$ gopass secret/structured_secret another_secret
even_more_surprises
```

My expectation would be that I can access the structured secret's key/value no matter the position they're in within the object.

dominikschulz:
This PR changes an important heuristic: It is no longer legal to have an YAML document without an introductory `document marker` (an optional start marker in the YAML spec).

On the one hand this makes YAML handling more robust and deterministic, but on the other hand this makes it a little less convenient.

Please help me decide if we want to enforce them or not. I'd like to get this PR merged in any case, but we could decide to **not enforce** the markers (i.e. support docs with and without markers and just add them on any change).

Fixes #192

### DIFF
--- a/gpg/mock/gpg.go
+++ b/gpg/mock/gpg.go
@@ -1,0 +1,65 @@
+package mock
+
+import (
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/justwatchcom/gopass/gpg"
+)
+
+// Mocker is a no-op GPG mock
+type Mocker struct{}
+
+// New creates a new GPG mock
+func New() *Mocker {
+	return &Mocker{}
+}
+
+// ListPublicKeys does nothing
+func (m *Mocker) ListPublicKeys() (gpg.KeyList, error) {
+	return gpg.KeyList{}, nil
+}
+
+// FindPublicKeys does nothing
+func (m *Mocker) FindPublicKeys(...string) (gpg.KeyList, error) {
+	return gpg.KeyList{}, nil
+}
+
+// ListPrivateKeys does nothing
+func (m *Mocker) ListPrivateKeys() (gpg.KeyList, error) {
+	return gpg.KeyList{}, nil
+}
+
+// FindPrivateKeys does nothing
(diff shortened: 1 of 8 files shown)


metalmatze:
So this is only for YAML files, right? The normal behavior of having the first line as password doesn't break? (Just to make sure).
Examples would be:
```
asdfhasdfh123
---
key: value
```

with no password:
```
---
key: value
```

A few examples to decide would be nice, because I'm not sure I really understand this right now.

dominikschulz:
I've put the examples in some test cases: https://github.com/justwatchcom/gopass/pull/193/files#diff-d84e2d685e1c463ca6de12364c991ddc

Will extract them to make it more readable.

dominikschulz:
Tested use-cases:

1.) Setting key to empty secret, will create the following document:
```
---
foo: bar
```

2.) Getting a key from an password-only secret will complain about a missing document marker:
```
password
```

3.) Adding a key to a password-only secret will add a document marker:
```
password
```
add `foo: bar`
```
password
---
foo: bar
```

4.) Trying to read a key from a YAML document w/o a document marker will complain about the missing marker
```

foo: bar
```

martinhoefling:
@dominikschulz  is there a migration for missing markers? Maybe I just missed a simple solution to list all secrets to (semi-) automatically edit them :-)

martinhoefling:
`for sec in $(gopass ls -f); do gopass edit $sec; done`

dominikschulz:
Actually no, documents w/o markers are just handled as regular plain text. Since we can't know if plain text is supposed to be YAML or not I didn't want to mess anything up by introducing those markers.

Issue #98, opened 2017-04-11
Improve gopass init


dominikschulz:
`gopass init` is currently confusing and lacking some features. We need to improve this.

## Use cases

1. Initialize the primary local store: `gopass init` - **OK**
1. Initialize a secondary local store: `gopass init ???` - *Confusing*
1. Initialize the primary local store and push it to a remote: ??? - *Confusing*
1. Initialize a secondary local store and push it to a remote: ??? - *Confusing*
1. Initialize Setup a new store (primary or secondary) and set it up for use by a team (moving around keys, adding recipients, ...): ??? - *Confusing*

PR #194, opened 2017-07-17
add hint for adding secondary stores (#98)

marthjod:
...in case an already initialized primary store is found

Does this cover use case 2 (#98)?


dominikschulz:
This definitely improves the UX. Thank you!

PR #198, opened 2017-07-19
Return with an error code when searching after not finding password

### LINKED ISSUE #197: gopass show exits with status 0 when no entry is found
Current behaviour:

``` bash
$ gopass show this_does_not_exist
Entry 'this_does_not_exist' not found. Starting search...
$ echo $?
0
```

This behaviour seems a bit strange, especially if you trying use gopass in a script in which case you end up with an error message in the secret. It would be nice if gopass exited with a non-zero exit status in this instance, or if this is not possible a flag that would enable this behaviour.

OS: OS X 10.11.6
gopass: 1.2.0_2
go: 1.8.3

metalmatze:
Fixes #197 

I'm not really sure if `1` is a correct or right code number to return with. :thinking:

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -2,6 +2,7 @@ package action
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/atotto/clipboard"
 	"github.com/fatih/color"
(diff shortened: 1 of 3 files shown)


metalmatze:
Just fixed.

PR #204, opened 2017-07-20
Also Import missing keys on recipients show

dominikschulz:
This PR also calls the import missing key func when requesting a recipient listing - an operation that takes some time anyway. This was we can try to ensure that no missing recipients keys are show in the recipients output.

### DIFF
--- a/action/recipients.go
+++ b/action/recipients.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


metalmatze:
Seems like a very good addition!

Only golint doesn't like it so far:
`store/sub/recipients.go:95:1: exported method Store.ImportMissingPublicKeys should have comment or be unexported`
:blush:

PR #205, opened 2017-07-20
Add version check

dominikschulz:
This PR adds a version check to the `gopass version` subcommand.
It will make an unauthenticated call to the GitHub REST API v3 every time the user runs `gopass version` unless `CHECKPOINT_DISABLE` is non-empty.

### DIFF
--- a/action/version.go
+++ b/action/version.go
@@ -29,7 +29,7 @@ func (s *Action) Version(c *cli.Context) error {
 	}
 
 	if r.Name != s.version {
-		fmt.Println(color.YellowString("\nYour version of gopass is out of date!\nThe latest version is %s.\nYou can update by downloading from www.justwatch.com/gopass", r.Name))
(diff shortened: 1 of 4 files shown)


action/version.go
```diff
@@ -1,9 +1,31 @@
...
+	}
+
+	r, err := ghrel.FetchLatestStableRelease("justwatchcom", "gopass")
+	if err != nil {
+		fmt.Println(color.RedString("\nError checking latest version: %s", err))
+		os.Exit(1)
+	}
+
+	if r.Name != s.version {
+		fmt.Println(color.YellowString("\nYour version of gopass is out of date!\nThe latest version is %s.\nYou can update by downloading from www.justwatch.com/gopass", r.Name))
```
metalmatze: I'd like to at least mention package managers to avoid confusion.
```diff
- You can update by downloading from www.justwatch.com/gopass.
+ You can update by downloading from www.justwatch.com/gopass or via your package manager.
```

PR #218, opened 2017-07-23
RFC: Use dep

dominikschulz:
This PR changes the dependency management to use the semi-official `dep` tool instead of `govendor`.

RFC: Should we switch to `dep` (yet)?


dominikschulz:
After some experiements with `dep` I don't think it's ready, yet.
We'll revisit this later.

PR #227, opened 2017-08-01
Added `gopass check` command

muesli:
`gopass check` validates known passwords against cracklib for common flaws,
like being too short or coming from a dictionary/word-list.

This currently requires linking against libcrack2, as I'm not aware of a Golang
implementation of cracklib.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ sudo: required
 before_install:
   - go get -v github.com/golang/lint/golint
   - sudo apt-get update -qq
-  - sudo apt-get install -y libcrack2
+  - sudo apt-get install -y libcrack2-dev
 
 after_success:
   - gem install fpm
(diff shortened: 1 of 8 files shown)


muesli:
Not sure what's going on with Travis here, but I'm fairly confident this should build fine once it's not trying to target a PR.

metalmatze:
Thanks for you contribution. 
The library you use requires CGO. We don't want any CGO (and thus C) in gopass, because it makes cross compiling hard to impossible.

Why not simply use a pipe to do want you want? We do that with a few other dependencies too.
`gopass foo/bar | cracklib-check`

Closing this, but feel free to comment if there's anything coming up without CGO.

muesli:
From a technical point of view I'd much prefer linking to a C library over piping & parsing text output.

I do understand your reasoning though. Happy to maintain it in my fork until there's a cracklib implemented in Golang. If someone else wants to implement the pipe, I'm happy to test and review it.

metalmatze:
The pipe already works. :wink:

PR #232, opened 2017-08-04
`gopass audit` now also detects dupe/shared passwords

muesli:
`gopass audit` will now display which entries share the same password and warn you about it.

### DIFF
--- a/action/audit.go
+++ b/action/audit.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"github.com/muesli/crunchy"
 	"github.com/urfave/cli"
@@ -20,6 +21,7 @@ func (s *Action) Audit(c *cli.Context) error {
 	var out io.Writer
 	out = os.Stdout
 
+	dupes := make(map[string][]string)
 	foundWeakPasswords := false
 	for _, secret := range t.List(0) {
 		content, err := s.Store.Get(secret)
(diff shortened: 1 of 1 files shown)


metalmatze:
Not sure if it makes sense to maybe parallelize with a worker pool or something like that. 
It takes a long time going through all 400 passwords. 
But it's not too bad in the end, because one doesn't run this every day, I guess.

Just wanted to add my experience in case someone feels the same.

dominikschulz:
Decrypting the secrets takes quite some time and generates a lot of noise. IMHO this is OK for now, but if anyone steps up to improve this that would be nice.

muesli:
I agree, the bottleneck here isn't actually inside crunchy's dictionary lookups (even with fairly large dictionaries it seems to manage >10 passwords per second on a regular i5 here), but spawning gpg decrypt processes one after another. This could probably be done in parallel.

What would be nice (but doesn't work): we could also store an audit-state for each secret somewhere. The problem here is, that users might change their passwords with tools other than gopass. At that point our cached audit-state becomes incorrect and meaningless.

FWIW, I have a patch ready that passes newly generated / inserted passwords through crunchy to spit out a warning at that stage. This means most users (who solely rely on gopass for their password storage needs) would only ever `gopass audit` once - while the slightly more paranoid user could still run it automatically via cron in the background and mail the results to themselves.

In the bigger picture, though, I'd rather have a meaningful and slow audit than a fast but sloppy one. That doesn't mean there's plenty of room for improvements, of course.

PR #233, opened 2017-08-04
Use fmt.Print without an output buffer in action/audit

muesli:
Got rid of the io.Writer, see remarks in #228.


muesli:
My pleasure. Thanks for merging all the PRs!

PR #234, opened 2017-08-04
Improve gopass audit UX

dominikschulz:
This PR improves output of the `gopass audit` command and improves handling on invalid entires.

### DIFF
--- a/action/audit.go
+++ b/action/audit.go
@@ -2,10 +2,17 @@ package action
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
+	"github.com/cheggaaa/pb"
+	"github.com/fatih/color"
 	"github.com/muesli/crunchy"
 	"github.com/urfave/cli"
+)
+
+const (
+	maxCheckLength = 256
 )
 
 // Audit validates passwords against common flaws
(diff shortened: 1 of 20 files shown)


action/audit.go
```diff
@@ -19,29 +26,49 @@ func (s *Action) Audit(c *cli.Context) error {
...
+		content, err := s.Store.GetFirstLine(secret)
 		if err != nil {
-			return err
+			fmt.Println(color.RedString("Failed to retrieve secret '%s': %s", secret, err))
+			continue
 		}
 
 		pw := string(content)
+		if len(pw) > maxCheckLength {
+			fmt.Println(color.RedString("Warning: Skipped long secret"))
```
muesli: Might be something we could even do upstream in crunchy. Dictionary lookups with that length certainly make no sense.

muesli:
Heads up: vendoring for new deps is missing.

As a sidenote: personally I had such a bad experience with pb in the past (it might have improved by now), I ended up writing my own (imo slightly saner solution) here: https://github.com/muesli/goprogressbar

You may or may not be interested in it, if pb does the job as you wanted it, I'd also be happy ;-)

dominikschulz:
Good ideas @muesli . Feel free to put those in another PR if you like. I just picked pb because I've used it in the past and it worked. But if you think your lib is considerably better, I'd be OK with replacing pb.

If you integrate the length check in crunchy we can drop the check here. I did some tests with some of my stores and stumbled upon `audit` being very slow with big/huge secrets

muesli:
@dominikschulz I'll add that extra check in crunchy. The WagnerFischer algorithm it uses becomes very slow if the strings you compare start diverging in length too heavily. Once the entropy / length becomes great enough, a password is fairly secure by definition, anyway.

That being said, when it comes to an audit command, I will always prefer accuracy over speed.

PR #238, opened 2017-08-06
Dropped cheggaaa/pb for muesli/goprogressbar

muesli:
I hope this doesn't come across as me having a bit of an NIH-syndrome.

I do find the original pb implementation a bit overly complicated and lacking though. I would encourage everyone to give this version a spin and compare the looks / behaviour of the two different progress-bars.

The only downside is goprogressbar assuming ANSI support on Windows, which only recently returned in Windows 10. pb is supposed to behave a bit nicer there, but I ran into various issues with it there, too (this may or may not be fixed by now, I don't have a Windows machine to test this right now). I will promise to look into this in the next few days, though.


muesli:
I've also introduced a bit of consistency in the audit's output's lingo.

muesli:
I can rebase / squash that if you want - unless you're already on it. Just let me know!

dominikschulz:
@muesli Please rebase.

muesli:
@dominikschulz Done!

muesli:
@dominikschulz Now we're good to merge!

dominikschulz:
Awesome, thank you.

PR #241, opened 2017-08-06
Fix GitVersion on uninitialized stores

### LINKED ISSUE #230: `gopass version` can't retrieve git version for uninitialized password stores
When the password-store hasn't been initialized yet, `gopass version` prints git version 0.0.0. While this might be a good indicator that the store just hasn't been initialized yet, it would still be useful to know the actual git version in case `gopass init` fails.

dominikschulz:
Fixes #230


muesli:
I can confirm this fixes #230 on the machine I originally found the issue.

Issue #206, opened 2017-07-20
Remove configuration options


dominikschulz:
`gopass` supports lots of configuration options.
This offers a lot of flexbility but adds lot of cognitive overhead to users and developers and makes testing all combinations quite hard.

We should try to reduce the number of configuration options to a bare minimum.

I propose the following changes:

Option | Action | Default | Comment
----------|----------|-----------|--------------
alwaystrust | remove | true | When encrypting the trust of a recipient shouldn't matter
askformore | keep | true | Nice feature, but we should not ask for more info by default
autoimport | keep | true | Missing recipient keys are super annoying and they virtually don't hurt
autopull | remove | true | If autopush is enabled we should always do autopull as well
autopush | keep | true | There must be a way to disable automatic git pushes
checkrecipients | remove | ? | Should we drop invalid recipients by default?
cliptimeout | keep | 60
debug | remove | false | Use GOPASS_DEBUG instead
gitrecurse | remove | true | Provide a command line flag to disable this per invocation
loadkeys | remove | true | Missing recipient keys are super annoying
nocolor | remove | false | Use GOPASS_NOCOLOR instead
noconfirm | keep | false | Confirming recipients per encryption is helpful, but sometimes annoying
nopager | remove | ? | Should we drop this?
path | keep
persistkeys | remove | true | We should always persist keys on adding recipients as it doesn't hurt in any way
safecontent | keep | false

zwopir:
nopager: can be dropped, if we use a `isTTY` or whatever check
autopull and autopush are not independent config parameters. Maybe we can put them in one config option `autorepoaccess = (on|off|pull)`. If we do, we need a better name than `autorepoaccess`.
`checkrecipients` should take into account if the secret can be encrypted for a users own gpg identity. It should at least encrypt with that identity, even if other recipients fail.

metalmatze:
I agree that we can get rid of most of them. Providing them via env seems way easier.
Combining autopull & autopush is :100: Maybe call it `gitsync` `gitautosync`?!

frntn:
@dominikschulz does the old configuration file will be compatible when updating gopass ?
If not how can we migrate without expected behavior of our own ?

Thanks

dominikschulz:
The old configuration will be compatible, but some behavior may change depending on the values you used.

If you have specific concerns we will try to address them before the next release.

frntn:
As of written, my concern was about `autopull`/`autopush` not appearing in the source code anymore.
[...]

PR #242, opened 2017-08-06
Disable checkRecipients by default as it contradicts alwaysTrust

dominikschulz:
If CheckRecipients is true AlwaysTrust doesn't matter anymore, since invalid recipients won't even be passed to gpg.

Also see #206 
@frntn

### DIFF
--- a/store/sub/store.go
+++ b/store/sub/store.go
@@ -36,15 +36,16 @@ type gpger interface {
 
 // Store is password store
 type Store struct {
-	alias      string
-	autoImport bool
-	autoSync   bool
-	debug      bool
-	fsckFunc   store.FsckCallback
-	importFunc store.ImportCallback
-	path       string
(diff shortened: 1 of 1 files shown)


frntn:
I have successfully build, install & test this commit against my setup : 
```
bash$ git remote add dominikschulz https://github.com/dominikschulz/gopass.git
bash$ git fetch dominikschulz
bash$ git checkout fix/checkrecp
bash$ make
bash$ sudo make install
bash$ /usr/bin/gopass edit team-devel/foo/bar
gopass: Encrypting /foo/bar for these recipients:
 - 0xAD01D1EDC256A009 - John Doe <john.doe@anon.com>
 - 0xD385289FA809234F - Matthieu Fronton <m@tthieu.fr>

Do you want to continue? [Y/n]: n
```

Thanks :+1:

Issue #154, opened 2017-06-20
Remove external dependencies - gpg


dominikschulz:
It would be nice if we could get rid of our external dependencies some day.
This would allow for a nicer workflow, better error and cross platform support.
This issue is meant to track any progress / discussions on that matter for gpg.

dominikschulz:
* https://godoc.org/golang.org/x/crypto/openpgp - The "official" PGP package by the Go authors, seems to be a little neglected and has some outstanding bugs, e.g. Secret Keyring parsing is flaky at best

yml:
I did exactly this in [my own password management lib](https://github.com/yml/keep) but now I am in dead end since golang openpgp does not support modern version of gnupg version above 2.1.

* https://github.com/mozilla/sops/issues/189
* https://github.com/oleiade/serrure/issues/6
* https://github.com/yml/keep/issues/1

bkcsoft:
For Gitea we use https://github.com/keybase/go-crypto/tree/master/openpgp instead of `x/crypto/openpgp` since we wanted support for new and shiny keys 🙂

yml:
@bkcsoft do you have an example on how to make it work with gpg agent ?

jungle-boogie:
It would still be an external dependency, but it's much more of a lightweight dependency:
https://github.com/tarsnap/scrypt

PR #245, opened 2017-08-08
Rewrite auditing to run concurrently with a worker pool

metalmatze:
I went ahead and re-implemented the audit command.
It's now using a worker pool to decrypt quicker. Sadly it doesn't improve the speed as much as I thought, but still a bit. For 400 secrets it now takes 55s instead of 80s.

I also improved the output because I needed to handle that differently due to the concurrency introduced:

```
Detected weak secret for 'foo/bar1': Password is mangled, but too common / from a dictionary
Detected weak secret for 'foo/bar2': Password is too short
Detected weak secret for 'foo/bar3': Password is mangled, but too common / from a dictionary
Detected weak secret for 'foo/bar4': Password is too short
```
becomes:
```
Password is too short:
        - foo/bar2
        - foo/bar4
Password is mangled, but too common / from a dictionary:
        - foo/bar1
        - foo/bar3
```
@muesli

### DIFF
--- a/action/audit.go
+++ b/action/audit.go
@@ -11,8 +11,8 @@ import (
 	"github.com/urfave/cli"
 )
 
-// AuditedSecret with its name, content a warning message and a pipeline error.
-type AuditedSecret struct {
+// auditedSecret with its name, content a warning message and a pipeline error.
+type auditedSecret struct {
 	name string
 
 	// the secret's content as a string. Needed for checking for duplicates.
@@ -39,11 +39,11 @@ func (s *Action) Audit(c *cli.Context) error {
 	jobs := make(chan string)
 
 	// Secrets that have been audited.
-	checked := make(chan AuditedSecret)
+	checked := make(chan auditedSecret)
 
 	// Spawn workers that run the auditing of all secrets concurrently.
 	validator := crunchy.NewValidator()
-	for worker := 1; worker <= runtime.NumCPU(); worker++ {
+	for worker := 0; worker < runtime.NumCPU(); worker++ {
 		go s.audit(validator, jobs, checked)
 	}
 
@@ -76,9 +76,10 @@ func (s *Action) Audit(c *cli.Context) error {
 		bar.LazyPrint()
 
 		if i == len(list) {
-			close(checked)
+			break
 		}
 	}
+	close(checked)
 	fmt.Println("") // Print empty line after the progressbar.
 
 	foundDuplicates := false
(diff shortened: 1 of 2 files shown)


action/audit.go
```diff
@@ -3,76 +3,133 @@ package action
...
+	"runtime"
 
 	"github.com/fatih/color"
 	"github.com/muesli/crunchy"
 	"github.com/muesli/goprogressbar"
 	"github.com/urfave/cli"
 )
 
+// AuditedSecret with its name, content a warning message and a pipeline error.
+type AuditedSecret struct {
```
dominikschulz: Could you make this private?

action/audit.go
```diff
@@ -3,76 +3,133 @@ package action
...
+	jobs := make(chan string)
+
+	// Secrets that have been audited.
+	checked := make(chan AuditedSecret)
+
+	// Spawn workers that run the auditing of all secrets concurrently.
 	validator := crunchy.NewValidator()
-	dupes := make(map[string][]string)
-	foundWeakPasswords := false
+	for worker := 1; worker <= runtime.NumCPU(); worker++ {
```
dominikschulz: `for worker := 0; worker < X; worker++` would be the usual idiom.

action/audit.go
```diff
@@ -3,76 +3,133 @@ package action
...
 		bar.Text = fmt.Sprintf("%d of %d secrets checked", bar.Current, bar.Total)
 		bar.LazyPrint()
 
-		content, err := s.Store.GetFirstLine(secret)
-		if err != nil {
-			bar.Clear()
-			fmt.Println(color.RedString("Failed to retrieve secret '%s': %s", secret, err))
-			continue
+		if i == len(list) {
+			close(checked)
```
dominikschulz: Closing the channel you're reading from looks odd. Couldn't you just `break` here?

dominikschulz:
The improved output looks so much nicer. Thank you!

action/audit.go
metalmatze: True. I wasn't sure where and how to close the channel. But I probably can simple `break` there, true.
It's a bit tricky, because you can't really close the channel within the workers themselves.

action/audit.go
metalmatze: Whoops. True.

action/audit.go
metalmatze: Yep.

muesli:
Note: you don't _have_ to close channels, they will be garbage-collected in Go, just like everything else.

muesli:
What I've noticed sine just merging this change: now GPG asks me for my key's passphrase twice.

metalmatze:
Hmm. Do you have a Dual Core CPU? :smile: 
It's probably the 2 goroutines running concurrently asking for the password at the same time. I'm glad you don't have 16 cores.

I wanted to close the channels just as a good practice.

muesli:
Actually I have a couple more cores, but I don't think it's related to that. Unlike before it also doesn't cache my keyphrase for a period anymore, but instead asks me on every `audit` run, at least once. Even weirder: at that point it already has like 30 passwords audited.

dominikschulz:
Ok, this is very strange. We worked on this and related features over the last days and never noticed this. When I run `gopass audit` `gpg` is asking the Passphrase `once per private key`, exactly like it should.

muesli:
I'll revert and make sure it's not something unrelated on my system.

muesli:
Nope, definitely related to this change. Will try to debug it here, since I can at least reliably reproduce it.

muesli:
Ok, this definitely is an issue with multiple gpg processes running parallely. Setting the number of parallel workers to one, the issue disappears.

dominikschulz:
Could you provide some details about your environment (OS, GPG Version, ...)?

muesli:
Latest ArchLinux
GPG: 2.1.22
i5 quadcore

Further findings: the more I increase the number of workers, the more often will I get to see the gpg keyphrase dialog pop up.

Also, if I cancel the popup, ranging over the checked secrets forgets to check for an error being returned. That causes an empty password to be detected for multiple services. No worries, got that fixed locally and will provide a PR, but still debugging the GPG spawning issues right now.

muesli:
You could probably reproduce this with this change:
```
	for worker := 0; worker < 32; worker++ {
```

dominikschulz:
Yes, indeed. This change makes it reproducible.

muesli:
Did you ever check out https://github.com/golang/crypto/tree/master/openpgp?

muesli:
Error handling is fixed, see: #249

dominikschulz:
Wrt. https://github.com/golang/crypto/tree/master/openpgp please see #154

PR #249, opened 2017-08-08
Fixed handling decryption errors during audit

muesli:
Only add passwords to duplicates if there wasn't an error during decryption.
Reuse the decryption error for grouping them in messages.

### DIFF
--- a/action/audit.go
+++ b/action/audit.go
@@ -64,8 +64,9 @@ func (s *Action) Audit(c *cli.Context) error {
 
 	i := 0
 	for secret := range checked {
-		duplicates[secret.content] = append(duplicates[secret.content], secret.name)
-
+		if secret.err == nil {
+			duplicates[secret.content] = append(duplicates[secret.content], secret.name)
+		}
 		if secret.message != "" {
 			messages[secret.message] = append(messages[secret.message], secret.name)
 		}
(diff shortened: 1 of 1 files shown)


action/audit.go
```diff
@@ -122,7 +123,7 @@ func (s *Action) audit(validator *crunchy.Validator, secrets <-chan string, chec
 	for secret := range secrets {
 		content, err := s.Store.GetFirstLine(secret)
 		if err != nil {
-			checked <- auditedSecret{name: secret, content: string(content), err: err}
+			checked <- auditedSecret{name: secret, content: string(content), err: err, message: err.Error()}
 			continue
 		}
```
metalmatze: An error in this case is an error in the pipeline. You're now adding  messages as errors too.  I'm not really sure that this actually does what it is supposed to do. 
Will check this tomorrow.

muesli: I'm adding messages for errors, so they show up in the output, grouped like the other messages, e.g. "Failed to decrypt".

Which is indeed an error in the pipeline, but that's worth highlighting in an audit. If you dislike warnings and errors being mixed together in the output, we could separate the two groups or color-highlight them.

muesli: I do see another tiny fix required now, though. PR coming up.

PR #250, opened 2017-08-08
Handle pipeline errors independently of audit errors

muesli:
Pipeline errors will now be printed in red at the bottom of the audit report.


dominikschulz:
My mistake, grouping is fine.

PR #259, opened 2017-08-12
Implement terminal selection

### LINKED ISSUE #258: Improve gopass find workflow
We would like to improve the `gopass find` workflow.

Currently a common workflow looks like this:
```
gopass find foo
# fiddle around with the mouse and/or Cmd+C/Cmd+V
gopass -c ...
```

It would be nice to remove the need to use a mouse in this case, e.g.:
```
$ gopass find foo
> xy/dev/foodb
   xy/dev/foobardb
   xy/dev/foobazdb
<↑/↓> to change the selection, <→> to show, <←> to copy, <ESC> to quit
```

dominikschulz:
Fixes #258


action/find.go
```diff
@@ -18,11 +19,28 @@ func (s *Action) Find(c *cli.Context) error {
...
+	if !s.isTerm {
+		for _, value := range choices {
 			fmt.Println(value)
 		}
+		return nil
 	}
 
-	return nil
+	act, sel := termwiz.GetSelection(choices)
+	fmt.Printf("User selected to %s %d\n", act, sel)
```
metalmatze: This seems more like a debugging message to me. Is it intended to always say: `User selected to show 20`? I don't really care about that as a user.

dominikschulz: Oh, yes indeed. This is left over debug code. Thanks for double checking. Will be gone in a sec.

dominikschulz:
@metalmatze Updated PR

PR #265, opened 2017-08-13
If find yields only one results display directly

### LINKED ISSUE #263: gopass find should directly display result if only one match
If running gopass find (explicit or implicit) and there is only one match it should directly display the result.

dominikschulz:
Fixes #263

### DIFF
--- a/action/find.go
+++ b/action/find.go
@@ -34,7 +34,6 @@ func (s *Action) Find(c *cli.Context) error {
 	}
 
 	act, sel := termwiz.GetSelection(choices)
-	fmt.Printf("User selected to %s %d\n", act, sel)
 	switch act {
 	case "copy":
 		return s.show(c, choices[sel], "", true, false, false)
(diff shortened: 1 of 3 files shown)


action/find.go
```diff
@@ -18,11 +20,33 @@ func (s *Action) Find(c *cli.Context) error {
...
 	needle := strings.ToLower(c.Args().First())
+	choices := make([]string, 0, 10)
 	for _, value := range l {
 		if strings.Contains(strings.ToLower(value), needle) {
+			choices = append(choices, value)
+		}
+	}
+
+	if len(choices) == 1 {
+		fmt.Println(color.YellowString("Found exact match in '%s'", choices[0]))
```
metalmatze: I think it would be better to have this not print a yellow string too.
Everything returned by this is yellow and the different messages are hard to distinguish.
What about making this line green?

dominikschulz: Ok, sounds good. Will change.

dominikschulz:
Changed the color and rebased.

Btw. this branch is based upon #259, so that would need to be merged first.

PR #266, opened 2017-08-13
Default to not include symbols on generate

### LINKED ISSUE #264: Do not default to symbols in generated passwords
The new [NIST guidelines](https://github.com/usnistgov/800-63-3) no longer recommend special characters in passwords and I always run `gopass generate -n`, so I think we might as well change the default to *not include symbols*.

dominikschulz:
Fixes #264


dominikschulz:
Updated. Now it should print a nice deprecation warning.

PR #276, opened 2017-08-20
Check newly inserted secrets with crunchy

muesli:
I decided to keep the dupe check out of this quick password audit, so inserting new secrets remains as fast as possible.


dominikschulz:
Should we also check passwords inserted with `gopass edit`?

muesli:
@dominikschulz: You're right, will file another PR for that.

PR #286, opened 2017-08-23
Add default editors for linux, windows and others

### LINKED ISSUE #279: gopass edit fails if $EDITOR is not set
jimb0@local ~]$ gopass edit Foo/Bar/foobar/console/info

You need a passphrase to unlock the secret key for
user: "Jim"
4096-bit RSA key, ID 9999999, created 2017-08-21 (main key ID 9999999)

Error: failed to run editor with /dev/shm/gopass-edit943276169/tempfile file

gopass 1.3.1 (8284376) go1.8 linux amd64
CentOS Linux release 7.3.1611 (Core)

I think this might related to the other issue I submitted. But if I do the above with "pass" instead of "gopass" works just fine.

dominikschulz:
Fixes #279

### DIFF
--- a/action/edit.go
+++ b/action/edit.go
@@ -61,10 +61,7 @@ func (s *Action) Edit(c *cli.Context) error {
 }
 
 func (s *Action) editor(content []byte) ([]byte, error) {
-	editor := os.Getenv("EDITOR")
-	if editor == "" {
-		editor = "editor"
-	}
+	editor := getEditor()
 
 	tmpfile, err := fsutil.TempFile("gopass-edit")
 	if err != nil {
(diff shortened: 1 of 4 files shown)


dominikschulz:
On Linux the user should always get his favourite editor because we check `$EDITOR` and fall back to `editor` which should be as symlink to the chosen/default editor.

I have no idea about windows, but I figure that notepad.exe should be always available. I'v read that there is no command line editor on x64 windows by default.

For all other OS, including Mac, I assume that `vi` should be always available (after all it's around since 1976). vim is much newer and may not be available.

We may also implement a more sophisticated editor selection/fallback, but I'd like to put this into another PR.

PR #296, opened 2017-08-28
Trim any trailing newlines or carriage returns in show output

frobware:
Instead of getting an additional blank line in the show output you now get:

```
$ gopass foo.com
bar
$ ls
<...>
```

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -3,6 +3,7 @@ package action
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/atotto/clipboard"
 	"github.com/fatih/color"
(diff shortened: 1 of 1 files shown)


action/show.go
```diff
@@ -91,7 +92,7 @@ func (s *Action) show(c *cli.Context, name, key string, clip, force, qr bool) er
 		}
 	}
 
-	fmt.Println(color.YellowString(string(content)))
+	fmt.Println(color.YellowString(strings.Trim(string(content), "\r\n")))
```
dominikschulz: unsing `strings.TrimSpace` would be more idiomatic

frobware:
> unsing strings.TrimSpace would be more idiomatic

I'm sure I pushed a commit that only considered space at the end.

action/show.go
frobware: On reflection I reduced this to trimming trailing \n's or \r's in case leading whitespace is part of the password.

PR #303, opened 2017-08-30
Implement gopass sync

### LINKED ISSUE #281: autosync - provide clear feedback about pushes
The current git push/pull output is very _technical_. We should try to make it easier to grasp for less technical users.

Example:
```
gopass insert foo/bar
...
[git] Changes have been pushed to git@github.com:my/store.git
```

dominikschulz:
Fixes #281

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,16 +193,10 @@ by unsetting `$PAGER` or `gopass config nopager true`.
 ### git auto-push and auto-pull
 
 If you want gopass to always push changes in git to your default remote (origin)
-enable autopush:
+enable autosync:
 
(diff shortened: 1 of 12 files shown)


dominikschulz:
I'll rebase and add the special case for the root store. Was already thinking about doing that anyway.

What do you mean with your remark about `Output ...`? This is an error case, isn't it?

dominikschulz:
@metalmatze Rebased and added the special case for the root store.

dominikschulz:
I'd prefer to keep the error ouput like it is for now. OK?

PR #314, opened 2017-09-06
Add secret type

dominikschulz:
This PR introduces a new type Secret which aims to
better encapsulate handling of secrets, especially
one containing YAML.


dominikschulz:
Yes, of course. It's probably not obvious why this makes sense w/o having dug too deep into the code/history of the project.

In the past we introduced several features on top of regular plain text secrets. However many of those features, like YAML, needed to be implemented spread all over the place. This made testing and modifications very hard and introduced a whole bunch of bugs over time. This PR is a first attempt to better manage these kinds of features by putting an abstraction on top of the plain text secrets. I'm not yet fully happy with the interface design, but it's an IMHO very good first step upon we can easily build further improvements.

PR #325, opened 2017-09-08
Omit newline on gopass show when out goes to a terminal

### LINKED ISSUE #324: Feature proposal: only print newline when output goes to a terminal
When gopass shows a secret it appends a newline character. This makes sense when gopass is invoked from a terminal and a human consumes the output. However, when the output is sent to another command, the newline character usually has to be stripped off using `tr` or similar. Doable, but a bit annoying.

The proposal is to detect where the output is sent to and only append the newline character if the output goes to a terminal.

I believe the change needs to be made in `show.go` at the end of the `show` function: https://github.com/justwatchcom/gopass/blob/2d46fce7a0b8227f7f986f38b6a4fd2a3e2c41f6/action/show.go#L96

There is a Go package to determine whether the output is going to a terminal or not: https://godoc.org/github.com/mattn/go-isatty

Sorry, I would do a PR but I have zero experience with Go and its tooling.

dominikschulz:
Fixes #324

### DIFF
--- a/action/show.go
+++ b/action/show.go
@@ -98,7 +98,10 @@ func (s *Action) show(ctx context.Context, c *cli.Context, name, key string, cli
 		}
 	}
 
-	fmt.Println(color.YellowString(strings.TrimRight(string(content), "\r\n")))
+	fmt.Print(color.YellowString(strings.TrimRight(string(content), "\r\n")))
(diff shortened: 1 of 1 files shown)


dominikschulz:
No, thsi is intentional. See #296

PR #327, opened 2017-09-12
Introduce global yes flag

### LINKED ISSUE #289: Skip  [y/N] prompt?
Is it possible when doing "gopass recipients add 99999 --store Foo" to skip the prompt for [y/N]. Like "gopass recipients add 99999 --store Foo --yes"? 

Goal is to allow scripting of adding many users at once.

dominikschulz:
This commit introduces a global flag '--yes' and a way to translate this into a context value.

Fixes #289


dominikschulz:
:rofl:

PR #330, opened 2017-09-13
Implement config options per mount

### LINKED ISSUE #35: Gopass options per-mount?

dominikschulz:
Fixes #35


dominikschulz:
Yes, you can!
Actually I did not (yet) test if this works properly with `gopass sync`, but I've tested most of the other flags (like cliptimeout, noconfirm, ...).

PR #331, opened 2017-09-14
Added Message/Response for querying hosts

martinhoefling:
This message allows querying for hostnames. If no hostname was found, hostname and subdomains are removed from the left hand side until a match occurs or only domain and tld remain.

E.g. _www.amazon.de_ would match _"something/amazon.de"_ but not _"li.de"_

For safety reasons, it is only matched again strings between / , at the beginning or at the end.

E.g. _tamazon.de_ would not match

_something/amazon.de 
amazon.de/something 
something/amazon.de/something_

Further, documentation for the error message was added


action/jsonapi.go
```diff
@@ -24,6 +26,10 @@ type queryMessage struct {
 	Query string `json:"query"`
 }
 
+type queryHostMessage struct {
```
dominikschulz: I start to wonder if we should move the implementation of the `jsonapi` to it's own package, e.g. `github.com/justwatchcom/gopass/utils/jsonapi` and only invoke it from this place?

action/jsonapi.go
```diff
@@ -107,13 +113,43 @@ func respondMessage(ctx context.Context, s *Action, msgBytes []byte) error {
...
+		return err
+	}
+
+	l, err := s.Store.List(0)
+	if err != nil {
+		return err
+	}
+	choices := make([]string, 0, 10)
+
+	reQuery := fmt.Sprintf("(^|.*/)%s($|/.*)", regexSafeLower(message.Host))
```
dominikschulz: This would definitely warrant some in-place documentation.

action/jsonapi.go
```diff
@@ -107,13 +113,43 @@ func respondMessage(ctx context.Context, s *Action, msgBytes []byte) error {
...
+	if err := searchAndAppendChoices(reQuery, l, &choices); err != nil {
+		return err
+	}
+
+	for len(choices) == 0 && strings.Count(message.Host, ".") > 1 {
```
dominikschulz: What exactly is this block doing?

action/jsonapi.go
martinhoefling: You mean invoke the jsonapi from `actions/jsonapi.go`? Sounds good - that would allow to easily split up the current stuff in `jsonapi.go` into multiple files. E.g. extracting the message struct spec.

martinhoefling: see latest commit. Thanks for the suggestion, it's much better now 👍

action/jsonapi.go
martinhoefling: I refactored that part into a function "isPublicSuffix". I hope it is more clear now. Let me know if it still needs a comment / further refactoring.

tests/config_test.go
```diff
@@ -17,7 +18,11 @@ func TestConfig(t *testing.T) {
 	assert.Contains(t, out, "autosync: true")
 	assert.Contains(t, out, "cliptimeout: 45")
 	assert.Contains(t, out, "noconfirm: true")
-	assert.Contains(t, out, "path: /tmp")
+	if runtime.GOOS == "darwin" {
```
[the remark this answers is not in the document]
martinhoefling: Yeah, it has nothing to do with this PR, I opened a separate one (#341). However, I do not get why this will fail on Windows?

action/jsonapi.go
martinhoefling: Comment added.

PR #335, opened 2017-09-15
one step closer to working on Windows

Zamiell:


Zamiell:
I tested this out locally and it seemed to fix the bug.

However, the tests failed for me. Since this is a somewhat trivial change, I suspect that the testing itself is just broken on Windows.

Zamiell:
What does this part of the code do? https://github.com/justwatchcom/gopass/blob/a65b17fd790ca98befa9117c099313f57bd947b1/action/fsck.go#L29-L34
Is it related? Will that code also fail on Windows?

alexmironof:
I suggest to use `HomeDir` in `User` struct from package `os/user`

dominikschulz:
Good idea, but AFAIK `os/user` needs CGO. Since we use cross-compilation to build the releases, this is a no-go. I'd suggest https://github.com/mitchellh/go-homedir instead.

config/location.go
```diff
@@ -3,20 +3,38 @@ package config
...
+	// (which is part of the XDG Base Directory Specification for Linux and
+	// other Unix-like operating sytstems)
 	if xch := os.Getenv("XDG_CONFIG_HOME"); xch != "" {
 		return filepath.Join(xch, "gopass", "config.yml")
 	}
+
+	// Third, check to see if we are running on a Windows platform
+	// We can check for platform via the "runtime.GOOS" variable:
+	// https://stackoverflow.com/questions/19847594/how-to-reliably-detect-os-platform-in-go
+	if runtime.GOOS == "windows" {
```
dominikschulz: Have you considered using https://github.com/mitchellh/go-homedir ?

PR #359, opened 2017-09-28
gopass recipients add/remove selection wizard

### LINKED ISSUE #351: `gopass recipients add` with no arguments should give the user a wizard in the exact same way that `gopass init` does
`gopass recipients add` without any additional parameters will just fail.

It should instead show a wizard. Specifically meaning that it should shows the user a list of all of the currently installed PGP public keys on the system and lists them by number along with the name/email attached to them so that the user can easily just pick one from the list (which is what happens when you do a `gopass init`).

Finally, `gopass recipients remove` should also have the same functionality.

dominikschulz:
Fixes #351

### DIFF
--- a/action/recipients.go
+++ b/action/recipients.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/fatih/color"
 	"github.com/justwatchcom/gopass/utils/ctxutil"
+	"github.com/justwatchcom/gopass/utils/termwiz"
 	"github.com/urfave/cli"
 )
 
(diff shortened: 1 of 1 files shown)


dominikschulz:
Not sure why you "requested changes"?!
Obviously there is an issue with your private key, but without any further info no one but you will be able to reproduce it.

dominikschulz:
We just confirmed that this works for us. Just raise an issue if you gather the info to reproduce your problem.

metalmatze:
Sorry to cause confusion. Not sure what the problem was, but I've not seen it again so far. :+1:

PR #367, opened 2017-10-01
Setup method for native messaging wrapper and manifest

martinhoefling:
I've split `gopass jsonapi` into `gopass jsonapi listen` and `gopass jsonapi configure`. Here, I implemented the manifest configuration for chrome, chromium and firefox on OSX and Linux. Global installation is supported as well as specifying an alternate lib path on linux.

Currently, the wrapper is installed in /usr/local/bin which requires superuser privileges on linux but not on OSX with brew. What are your thoughts on this? Installing them alongside with the gopass config would also be an option.

Tests: ... any ideas what would be meaningful tests? They should be platform independent right? So I've tested the functionality by hand on osx and linux.

I've also updated the gopassbridge plugins for firefox and chrome to 0.0.5 and published them so that with that changes - a normal user should be able to set up its browser.

### DIFF
--- a/utils/jsonapi/manifest/location.go
+++ b/utils/jsonapi/manifest/location.go
@@ -1,0 +1,69 @@
+package manifest
+
+import (
+	"fmt"
+	"path"
+	"runtime"
+)
+
+func getLocation(browser, libpath string, globalInstall bool) (string, error) {
+	switch platform := runtime.GOOS; platform {
+	case "darwin":
+		{
+			switch browser {
+			case "firefox":
+				{
+					if globalInstall {
+						return "/Library/Application Support/Mozilla/NativeMessagingHosts/%s.json", nil
+					}
+					return "~/Library/Application Support/Mozilla/NativeMessagingHosts/%s.json", nil
+				}
+			case "chrome":
+				{
+					if globalInstall {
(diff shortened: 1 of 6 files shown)


utils/jsonapi/manifest/location.go
```diff
@@ -0,0 +1,69 @@
...
+
+func getLocation(browser, libpath string, globalInstall bool) (string, error) {
+	switch platform := runtime.GOOS; platform {
+	case "darwin":
+		{
+			switch browser {
+			case "firefox":
+				{
+					if globalInstall {
+						return "/Library/Application Support/Mozilla/NativeMessagingHosts/%s.json", nil
```
gburtini: All these paths should probably be stored in an easily configurable dictionary. Perhaps in a config. file? Especially the linux ones where you cannot make such a strong assumption about the shape of people's filesystems.

martinhoefling: Storing in a dictionary was actually my first approach, but you end up with at least three nesting levels and dynamic nesting levels (more like a tree), which needs special handling too.

Not sure about the config file: Those are typically written by gopass on initial run/configuration. If we ship a separate file, wouldn't that deviate from the single binary contains it all principle?

One other solution I could think of is making heavy usage of [templating](https://golang.org/pkg/text/template/) but I am not sure if that is more readable at the end.

Issue #368, opened 2017-10-01
Add English-like/pronouncable password generation.


gburtini:
**Motivation**: often, a password has to be communicated through a side-channel or remembered (either for the short or long run), despite being held in a password manager. For example, some of my passwords get entered on iOS where I do not currently have a method to securely use my GPG hardware token. Despite this, there are techniques to improve the communicability of passwords while still maintaining a high-degree of security.

There's a few models for discussion here.

1. The simplest and most common model, further constrain the character set to mostly alternate between vowels and non-vowels. This can be augmented with special characters infrequently. 

2. The best model, the XKCD approach. We take an English (or maybe, _any language_) dictionary and select M words from it (with a minimum total character length of N), then assemble those words separated by a small class of characters, perhaps `_` and `-` by default.

3. [2], but we also mutate the English words to only be English-like. Ideas here include transmuting vowels, swapping common morphemes (this would be English-specific likely, due to the need to define a dictionary of morphemes), etc. Done right, this would generate non-words that still have communicative and memorable value, e.g., `Welok` and `Bomloy`, combined together to produce a password like `Welok_Bomloy-Twolith.` 

Each decision here (e.g., how frequently to have non-pronounceable characters) can have its effective entropy calculated. We should define a target entropy before embarking on any of this.

To whatever extent possible, we should defer to a password generation library that will do this for us. If a satisfactory library cannot be found, I would be interested in participating in that. I have not yet searched.

dominikschulz:
Thank you very much for bringing this up. IMHO this is a very good idea and we'd like to include this feature.

So far we didn't find any usable password generation libraries for Go, this is why we even wrote the simple `pwgen`-like generator ourselves.

If you would like to contribute work on an advanced password generation package, that would be awesome.

martinhoefling:
@dominikschulz  I work on a vendorable xkcd password generator lib / standalone tool. So you can assign that to me if you want. 

@dominikschulz would it be ok for you guys if I reuse init() and randomInteger() from pwgen.go? I guess it's not "Substantial portion" in the sense of MIT license - so are you fine with a reference to gpass and the source file? I would put my stuff under MIT license as well.

dominikschulz:
@martinhoefling Please go ahead with both proposals. Unfortunately I can't seem to assign the issue to you (https://help.github.com/articles/assigning-issues-and-pull-requests-to-other-github-users/). No need to ask for code reuse as long as you use it in open source code.

[...]

PR #373, opened 2017-10-03
Vendorized and integrated xkcd password generation (#368)

martinhoefling:
see #368 - exact (default) parameters to be discussed.


dominikschulz:
@martinhoefling Please rebase and add a switch for the delimiter.

gburtini:
That was fast! Great work @martinhoefling.

action/generate.go
```diff
@@ -10,18 +10,22 @@ import (
...
+	defaultXKCDLength = 4
 )
 
 // Generate & save a password
 func (s *Action) Generate(ctx context.Context, c *cli.Context) error {
 	force := c.Bool("force")
 	edit := c.Bool("edit")
 	symbols := c.Bool("symbols")
+	xkcd := c.Bool("xkcd")
+	xkcdo := c.Bool("xkcdo")
```
gburtini: This could use a better variable name. I am not sure what distinguishes it from `xkcd` above.

gburtini: Looking further down, I see it causes the separator to be a space. I would instead prefer if you offered an argument for the separator. Perhaps, if you're feeling up to it, a list of separators to be randomly used.

e.g., `--xkcd --separators=",-_"` to randomly select one of comma, dash and underscore as a separator.

main.go
```diff
@@ -433,6 +433,14 @@ func main() {
 					Name:  "symbols, s",
 					Usage: "Use symbols in the password",
 				},
+				cli.BoolFlag{
+					Name:  "xkcd, x",
+					Usage: "Use multiple random english words as password, separated by space",
+				},
+				cli.BoolFlag{
+					Name:  "xkcdo, xo",
+					Usage: "Use multiple random english words as password, no separator but CamelCased",
```
gburtini: While it is certainly debatable, this is not what I recognize as `camelCase` (lower first letter), but rather `ClassCase` or `PascalCase`: https://en.wikipedia.org/wiki/Camel_case#Variations_and_synonyms

Perhaps not an important comment given the discussion of adding a separator flag. 

Additionally, you should capitalize English.

action/generate.go
```diff
@@ -54,8 +58,14 @@ func (s *Action) Generate(ctx context.Context, c *cli.Context) error {
 	}
 
 	if length == "" {
-		length = strconv.Itoa(defaultLength)
-		if l, err := s.askForInt("How long should the password be?", defaultLength); err == nil {
+		def := defaultLength
```
gburtini: Why is this variable called `def`? What does it mean? Can you use a more descriptive name? `candidateLength` or `lengthAsString` perhaps, if that is what it represents here?

action/generate.go
dominikschulz: These variable names are indeed confusing. I was rather thinking about an switch like `delimiter` or `separator` which contains the delimiter. You could also support multiple separators like @gburtini suggested. I'm not sure if that is a good idea, because having to remember the exact placement of each different separator makes the generated password much harder to remember, but we could offer it for users who want it.

main.go
dominikschulz: I think the name Camel Case is OK. So far I never met someone making a difference between `camelCase` and other variations like `ClassCase`.

I agree that we should credit the original XKCD comic. Probably somewhere in the Readme.

action/generate.go
dominikschulz: At the second look this whole length calculation block looks a little odd to me. First converting from int to string and later back to int doesn't feel right.

Maybe you could untangle this a bit?

main.go
gburtini: Yeah, fair enough. If we add a separator flag instead, this comment is likely to go away anyhow, so any confusion about what the case means will be a non-issue.

action/generate.go
gburtini: You're right - I think it disproportionately impacts memorability. Perhaps we should _not_ add it.

It does not add significant bits of entropy [1] if you can assume users are only going to use natural separators.

[1] 4 words from Martin's 7777 wordset is 7777^4 options. Allowing the ASCII character set of separators would add 255^3 more to that set while greatly decreasing memorability.

martinhoefling: So do you have a strong opinion about a "delimiter" switch? From a user perspective, I see two cases: 
* Easier to type: lowercase + space and 
* Easier to copy / paste 1 Word-> no spaces but readable by inserting caps (you're right - it's not camelcase).

With two explicit switches, I would always need only one switch. With a delim switch, you need two for one of the two cases.

What about these two switches.

xkcdnosep, xn Use multiple random english words combined to a password without spaces/separator. First character of words is capitalized.
xkcdsep, xs Use multiple random english words combined to a password with separator. If no separator is given, the default is space.

The default is chosen if empty string is given in the option.

martinhoefling: Hmm, unfortunately urfave/cli does not support optional arguments. So I would then go for:

xkcd, x Use multiple random english words combined to a password. If no separator is specified, the words are combined without spaces/separator and the first character of words is capitalised.
xkcdsep, xs Word separator for generated xkcd style password.

martinhoefling:
Review comments addressed: See latest commit f4f2044

PR #374, opened 2017-10-04
Add Data() method to retrieve all secret data

raphink:
There is currently no way to retrieve all the password data of a secret (unless you parse it as a string or []byte).


raphink:
Thank you!

PR #375, opened 2017-10-04
Do not enforce YAML parsing

raphink:
Try to parse in YAML and give up silently if it fails

Some tools (e.g. [the lastpass importer])https://git.zx2c4.com/password-store/tree/contrib/importers/lastpass2pass.rb#n60) create invalid YAML which makes `gopass` crash. It seems easier and safer to just try and parse in YAML, and give up without crashing otherwise.

### DIFF
--- a/store/secret/secret.go
+++ b/store/secret/secret.go
@@ -45,13 +45,10 @@ func Parse(buf []byte) (*Secret, error) {
 
 // decodeYAML attempts to decode an optional YAML part of a secret
 func (s *Secret) decodeYAML() (bool, error) {
-	if !strings.HasPrefix(s.body, "---\n") {
-		return false, nil
-	}
 	d := make(map[string]interface{})
 	err := yaml.Unmarshal([]byte(s.body), &d)
 	if err != nil {
-		return true, err
+		return false, nil
 	}
 	s.data = d
 	s.body = ""


dominikschulz:
We've introduced the enforcement of the document sperator on purpose and I'm not sure if lifting this limitiation is a good idea. We had some discussion when we introduced it and I still think it's better to explicitly require well formed YAML than relying on undefined results.

raphink:
@dominikschulz I understand. However, this means documents generated with lastpass2pass cannot be edited (even to be fixed) by gopass, because it crashes as it sees the `---` prefix but the YAML document is invalid. Could the error be turned into a warning maybe?

dominikschulz:
Only the YAML decoding should fail. Editing the document itself should still be possible.

Could you post a redacted secret that allows us to reproduce this issue?

raphink:
```
 gopass edit perso/Test/test.com 

Error: failed to decrypt perso/Test/test.com: yaml: line 2: mapping values are not allowed in this context
github.com/justwatchcom/gopass/action.(*Action).Edit
	/home/raphink/go/src/github.com/justwatchcom/gopass/action/edit.go:34
main.main.func24
	/home/raphink/go/src/github.com/justwatchcom/gopass/main.go:367
github.com/justwatchcom/gopass/vendor/github.com/urfave/cli.HandleAction
	/home/raphink/go/src/github.com/justwatchcom/gopass/vendor/github.com/urfave/cli/app.go:485
github.com/justwatchcom/gopass/vendor/github.com/urfave/cli.Command.Run
	/home/raphink/go/src/github.com/justwatchcom/gopass/vendor/github.com/urfave/cli/command.go:193
github.com/justwatchcom/gopass/vendor/github.com/urfave/cli.(*App).Run
	/home/raphink/go/src/github.com/justwatchcom/gopass/vendor/github.com/urfave/cli/app.go:250
main.main
	/home/raphink/go/src/github.com/justwatchcom/gopass/main.go:832
runtime.main
	/home/raphink/goroot/src/runtime/proc.go:185
runtime.goexit
	/home/raphink/goroot/src/runtime/asm_amd64.s:2197
```

with the entry containing something like:

```
somepasswd
---
Test / test.com
username: myuser@test.com
password: somepasswd
url: http://www.test.com/
```

dominikschulz:
Ok, thank you. This must not fail.

Will investigate.

dominikschulz:
Please see #380

PR #380, opened 2017-10-05
Turn YAML errors into warnings

### LINKED ISSUE #379: YAML errors should be only warnings
From https://github.com/justwatchcom/gopass/pull/375#issue-262779363:

> Some tools (e.g. [the lastpass importer])https://git.zx2c4.com/password- store/tree/contrib/importers/lastpass2pass.rb#n60) create invalid YAML which makes gopass crash. It seems easier and safer to just try and parse in YAML, and give up without crashing otherwise.

dominikschulz:
Fixes #379


raphink:
Thanks

PR #383, opened 2017-10-05
better docs

Zamiell:


Zamiell:
erm, the above comments are meant to be seen [inline on the commit](https://github.com/Zamiell/gopass/commit/acd8101e6d7640fbb99f62bdd980af6893e17a46)

Zamiell:
I was thinking of doing a second PR. Two questions:

1) There is no consistent line wrapping standard throughout the documentation. I propose there should either be no line wrapping at all, or there should be an 80 character limit. See [this discussion](https://about.gitlab.com/2016/10/11/wrapping-text/). Which is it? I personally feel more comfortable with no wrapping, but either way, I can fix it.
2) There is no consistent stylization of the word gopass throughout the documentation. Sometimes it is stylized as gopass and other times it is stylized as `gopass`. I think it actually reduces clarity to always stylize it as the latter, as it can cause confusion as to whether the docs are talking about the program or instructing the user to execute the gopass command. Which is it?

dominikschulz:
Please do so! That would be very welcome.

1. Personally I also avoid line wrapping and rely on my editor to do that. No hard feelings though. If anyone comes up with a case for line wrapping that'd be also OK for me.
2. I've tried to style it as `gopass`, but again: no strict preference from my side, I'd be fine with styling it as plain gopass if you prefer that and reserve the `gopass` styling for verbatim commands.

PR #414, opened 2017-10-22
add config option to toggle colored output

qbit:
`make tests` passes this time. Previously I had only run `make test` and `make test-integration`, because apparently `s` is invisible to me :D

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -42,6 +42,7 @@ func New() *Config {
 			NoPager:     false,
 			SafeContent: false,
 			UseSymbols:  false,
+			NoColor:     false,
 		},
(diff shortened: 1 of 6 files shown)


dominikschulz:
We had this option in the past. We've discussed the removal in #206 and removed it in #213.
Do you have a good use case over using `GOPASS_NOCOLOR`?

qbit:
Having two locations (env var and config file) that dictate the behaviour of gopass seems  to increase the cognitive overhead to me. IMO, the ideal situation would be to have all or none of the config options specified via env var.

dominikschulz:
Right now we need to keep the envvar for nocolor to support the integration tests, but we may consider removing it in the future.

PR #417, opened 2017-10-24
add openbsd install instructions

qbit:
gopass has been imported into the OpenBSD ports tree as [security/gopass](http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/security/gopass/). This means people running `-current` will be able to `pkg_add` gopass as soon as new packages are cut!


falschparker82:
nooooice! 🎉

PR #435, opened 2017-10-30
Add more desktop notifications

dominikschulz:
This commit adds more desktop notifications and introduces
the gopass logo to these notifications.


dominikschulz:
Not exactly. This will cause a desktop notification when your clipboard gets auto-cleared (after the timeout expires).

This has two reasons: 1.) there is no other way to inform the user if this operation succeeded or not, i.e. if an error ocurs 2.) this feature was broken for quite some time and no one noticed due to 1.)

PR #444, opened 2017-11-04
Improved documentation and more useful default for wrapper dir

martinhoefling:
Documentation how the username is determined.

Use the config directory instead of /usr/local/bin as default for wrapper. This has two main advantages:
- guaranteed to be writeable for the user
- not in $PATH, so `gopass_wrapper.sh` won't block autocompletion.

### DIFF
--- a/action/jsonapi.go
+++ b/action/jsonapi.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/fatih/color"
+	"github.com/justwatchcom/gopass/config"
 	"github.com/justwatchcom/gopass/utils/jsonapi"
 	"github.com/justwatchcom/gopass/utils/jsonapi/manifest"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


config/io.go
```diff
@@ -108,7 +107,7 @@ func (c *Config) Save() error {
 		return errors.Wrapf(err, "failed to marshal YAML")
 	}
 	cfgLoc := configLocation()
-	cfgDir := filepath.Dir(cfgLoc)
+	cfgDir := Directory()
```
dominikschulz: There is no need to change this line in this PR, is there?

config/location.go
```diff
@@ -67,3 +67,9 @@ func PwStoreDir(mount string) string {
 	}
 	return filepath.Join(Homedir(), ".password-store")
 }
+
+// Directory returns the configuration directory for the gopass config file
+func Directory() string {
+	cfgLoc := configLocation()
+	return filepath.Dir(cfgLoc)
```
dominikschulz: Why not just `filepath.Dir(configLocation())`?

action/jsonapi.go
```diff
@@ -91,7 +92,7 @@ func (s *Action) getLibPath(ctx context.Context, c *cli.Context, browser string,
 func (s *Action) getWrapperPath(ctx context.Context, c *cli.Context) (path string, err error) {
 	path = c.String("path")
 	if path == "" {
-		path, err = s.askForString(ctx, color.BlueString("In which path should gopass_wrapper.sh be installed?"), manifest.DefaultWrapperPath)
+		path, err = s.askForString(ctx, color.BlueString("In which path should gopass_wrapper.sh be installed?"), config.Directory())
```
dominikschulz: I'm not sure if it's a good idea to place the script in the config folder, but as the user can still choose another location this should be OK.

martinhoefling: Yeah, I am also not fully happy about this, but as far as I know, there's not FHS compliant location to store user scripts that shouldn't be reachable via $PATH

martinhoefling:
Review comments adressed.

action/jsonapi.go
dominikschulz: Yes, I agree.

PR #460, opened 2017-11-13
Don't try to update any HEAD build, regardless of version

muesli:
I think the original check is faulty, as it's never supposed to try and update a HEAD build, regardless of the exact version. This also leads to the problem, that gopass currently thinks that 1.6.0-git+HEAD is updateable to 1.6.0 stable.

This change cancels the update check for all manual HEAD builds.

### DIFF
--- a/action/version.go
+++ b/action/version.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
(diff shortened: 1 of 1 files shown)


dominikschulz:
Good catch, that broke when changing the default version number.

PR #469, opened 2017-11-16
Add pledge(2) support for OpenBSD

qbit:
Pledge (OpenBSD only) is a way to restrict the operations an
application can make use of on OpenBSD. Applications declare
their intent by specifying the system calls they will use,
after which point, any use outside of the pledge'd calls will
result in the kernel killing the process.

More info on pledge can be found here: http://man.openbsd.org/pledge


falschparker82:
Note that this might disable the version check web call in `gopass version` (don't know enough about OpenBSD to verify):
`protect.Pledge("stdio rpath tty proc exec")`

Otherwise, a great idea IMHO, we might want to add `seccomp` as well then: https://godoc.org/github.com/seccomp/libseccomp-golang

falschparker82:
(oh well, libseccomp bindings invoke CGo: that would probably open a whole new can of worms, forget what I just said...)

dominikschulz:
CGo would only be an option if we find a way to cross-compile our releases. CGo for Linux-only files would probably be possible, but restrict building releases to Linux.

If you still want Seccomp it would probably be best to move this to another issue.

dominikschulz:
@qbit If I understand the pledge documentation correctly this breaks the version check as @falschparker82 noticed. IMHO we're missing the `inet` capability.

I'll leave it to your discretion if this is OK or not.

PR #473, opened 2017-11-18
add wpath and cpath to pledges

qbit:
Prior to this, running gopass would abort when running without
an initialized password store.


jungle-boogie:
This prevented the coredump on my openBSD system. Without this, `gopass` simply coredumps.

PR #475, opened 2017-11-24
Adding concurrency for reencrypt

AnomalRoil:
This a PR to add support for concurrent re-encryption. 
Unfortunately, I discovered only once I had it all working that when using GPG2, gpg-agent is blocking and does not support concurrent access to its secring. This means that it does not constitute the significant speed-up one could hope for. (If a too big value is used in the setting it might even slow down the process.)

With GPG1, I tried it on a dummy store with 86 small secrets and it went down from:
```
time ./gopass recipients rm -store "<root>"   0xDA0BDE9DB49B2F71
87% cpu 4.382 total
```
to a mere 
```
time ./gopass recipients rm -store "<root>" 0xDA0BDE9DB49B2F71
329% cpu 1.816 total
```

If you plan to use Go's OpenPGP later on, I still think this is a worthy PR, since it would then allow to fully exploit all cores when re-encrypting. BTW, notice that if you plan to use Go's OpenPGP, you should still keep a fall-back to gpg since it does not support smartcards, while gpg does.

### DIFF
--- a/pkg/action/config_test.go
+++ b/pkg/action/config_test.go
@@ -44,6 +44,7 @@ func TestConfig(t *testing.T) {
   autoimport: true
   autosync: true
   cliptimeout: 45
+  concurrency: 1
   editrecipients: false
   nocolor: false
   noconfirm: false
@@ -87,6 +88,7 @@ foo/nopager: false`
   autoimport: true
   autosync: true
   cliptimeout: 45
+  concurrency: 1
   editrecipients: false
   nocolor: false
   noconfirm: false
@@ -105,7 +107,7 @@ mount 'foo' config:
   notifications: false
 `
 	want += "  path: " + backend.FromPath("").String()
-	assert.Equal(t, want, strings.TrimSpace(buf.String()), "action.setConfigValues")
+	assert.Equal(t, want, strings.TrimSpace(buf.String()), "action.printConfigValues")
 	buf.Reset()
 
 	delete(act.cfg.Mounts, "foo")
@@ -134,6 +136,7 @@ autoclip
 autoimport
 autosync
 cliptimeout
+concurrency
 editrecipients
 nocolor
 noconfirm
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -44,6 +44,7 @@ func New() *Config {
 			AutoImport:    true,
 			AutoSync:      true,
 			ClipTimeout:   45,
+			Concurrency:   1,
 			NoColor:       false,
 			NoConfirm:     false,
 			NoPager:       false,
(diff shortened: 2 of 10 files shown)


dominikschulz:
I'm not sure about this PR. The code loos good (maybe a few nitpicks), but I'm not sure if the added complexity is worth the speedup. Especially since we're officially recommending gpg 2 by now, which doesn't work with this.

We explored this in the past and didn't pursue it any further exactly because of the limited benefit with gpg 2.

And while we want to switch to a pure Go GPG implementation some day, I think this is too far away to justice this PR either.

But I'd be intrested what others think of this: @metalmatze @falschparker82 @martinhoefling @muesli @qbit ?

qbit:
I am all for simplicity! Just out of curiosity what kinda bugs were you guys seeing?

AnomalRoil:
Well, by default, this is currently disabled, since the concurrency setting sets the number of concurrent workers and its default is 1. 
This approach using an unbuffered channel and waitgroups is a good way to perform concurrent exec calls in Go and wait over their return values, so it does not introduce bugs of the sort you might have experienced. (I guess you had cases where the goroutines would terminate before the exec call returns?)
I believe this is a really simple solution to the concurrent execution, which might be useful to the few ones still using gpg1.

On the other hand, it is true that this is not really interesting when one is using gpg2 since gpg2 relies on gpg-agent and the gpg-agent is by design's choice not concurrent, and probably won't be in the next few months/years.

I believe it wouldn't hurt to already have it, even if you do not migrate soon to a pure go GPG solution, but that's your call.

qbit:
I didn't realize this only worked with gpg1 - IMO that's a good reason to not add it. I am not a fan of adding complexity to code that is going to be used less and less.

AnomalRoil:
It does work with gpg2, it's just that it won't be increasing the performance as it does with gpg1. 
The code is completely generic, I wouldn't have proposed it if it was only for gpg1. It's just that gpg-agent used by gpg2 is blocking, so concurrent access to it are not possible.

muesli:
The solution is alright, but I have to agree: it's not going to be an improvement for the vast majority of gopass' users (who run gpg2), so the added complexity outweights the benefits here. I'd vote for keeping the pull request open, though. The situation may change in the future, especially if we manage to directly interface with libgpg.

dominikschulz:
@AnomalRoil Sorry for the long delay on this one, but since we're started to support different crypto backends recently I think it could be worthwhile to merge this after all.
If you're still interested in getting this merged please rebase onto the current master and then we can merge it quickly.

AnomalRoil:
It seems that the refactor from  #715  are getting in my way, plus there were many changes since my previous commit, so it'll take some time to refactor, I can't just do it easily through a rebase, since files moved and it seems git won't follow them.

dominikschulz:
I'm sorry for us taking so long on this. If you decide to refactor this branch, that would be nice.
Otherwise feel free to close it.

AnomalRoil:
Alright, I've re-implemented it.

Note that currently only the reencrypt function is using it, are there other batch operations where I should add it?

dominikschulz:
Thanks a lot! For now I think this is good. There are some other places where we could add concurrency, but reencryption is actually the most important place already.

AnomalRoil:
I'd be curious to do some benchmarks with and without concurrency using another backend, but I'm afraid I've got no time to try new backends right now.

PR #553, opened 2017-12-22
Drop unnecessary type conversions

ferhatelmas:
Reported by unconvert linter.

### DIFF
--- a/action/create.go
+++ b/action/create.go
@@ -151,7 +151,7 @@ func (s *Action) createPrintOrCopy(ctx context.Context, c *cli.Context, name, pa
 	if c.Bool("print") {
 		fmt.Printf(
 			"The generated password for %s is:\n%s\n", name,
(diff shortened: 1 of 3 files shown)


ferhatelmas:
If you want to integrate it, I can add it in a separate PR.

dominikschulz:
Yes, that would be nice.

ferhatelmas:
I just came to add unconvert but it's already done by you :+1: 

How about using gometalinter to run all these linters so as to have smaller make file.

dominikschulz:
Hey, I'm sorry if you were about to open a PR. I was trying to fix a bug in the Makefile and just added this at the same time.

Gometalinter is interesting, but we had some trouble with that at work and I'd prefer to stick to the explicit approach we use right now.

ferhatelmas:
No worries, I will try to make myself useful in other parts.

Issue #557, opened 2017-12-22
gopass zsh completion can't be packaged in its current form.


Foxboron:
When running `make completion`, the resulting `zsh.completion` file should be inserted into `/usr/share/zsh/site-functions/_gopass`. However, the current file expects the user to source the file to enable completion. This isn't really a propper way to provide zsh completion. 

An example solution would be to provide a zsh completion file:
```
λ fox@theia ~ » cat /usr/share/zsh/site-functions/_gopass
#compdef gopass 

_values '' $(gopass --generate-bash-completion)
```

I'm however unsure how you want to solve this, so creating an issue to discuss a propper solution.

dominikschulz:
Hi, thank you for your packaging efforts. Unfortunately I'm not sure if I can follow you here. Shell completion isn't my strength and I've never used zsh personally.

The cli library we use, generates bash completion automatically and we want to provide some hints on how to set this up on a per user basis. So far we didn't really think about packaging this.

What exactly do you need to package the gopass zsh completion? Shall we include your snippet as an alternate system wide zsh completion script or can we just replace the existing one? I'll need some input here. Maybe @metalmatze can, AFAIK he's been using zsh as well.

Foxboron:
I'm not the best at explaining myself, so apologize for any confusion.

zsh currently piggybacks on the bash completion, but the script makes some assumptions that doesn't really hold. It works if you source it inside your `.zshrc`, or that the resulting script will be sourced. zsh has a way of dealing with completions that are installed globally, but it's not compatible with the way gopass currently does things.

What i need is honestly a rewrite of what `gopass completion zsh` returns, something along the lines of what i wrote above with some more vetting as i'm not the strongest in zsh completion scripting myself \o/ This would probably break other peoples setups tho :/

dominikschulz:
Ok, we can definitely include this, but I'll need some help with that.

Foxboron:
I saw the `fish.completion` file in the root just now, i think a PR to just add `zsh.completion` is the propper way to solve this. If you want i can just add the simple version now in a PR, and write a propper completion file whenever i have time over christmas.

PR #560, opened 2017-12-23
Added zsh completion file

Foxboron:
Started working on a better zsh completion file. It's not perfect as it's not able to complete all flags for the second subcommand like `gopass audit hibp`. But it's a lot better then the current state of things.

Please review and check for spelling mistakes or errors :)

Refference: #557

Signed-off-by: Morten Linderud <morten@linderud.pw>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ dist/
 
 *.completion
 !fish.completion
+!zsh.completion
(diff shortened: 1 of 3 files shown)


Foxboron:
I forgot to fix the makefile. I'll get on it when i have time :)

dominikschulz:
This looks ok by itself, but we'd have to maintain it Everytime we change the commands.
Could we maybe just generate this (and the fish completion) somehow?

Foxboron:
I have no idea how that would work sadly.

dominikschulz:
Please see #565 for my current approach to generating zsh (and possily fish) completion scripts.
This needs some more work (wrt. flags), but once completed this should enable us to generate shell completion scripts that neither rely on bash nor have to be kept manually in sync.

Foxboron:
@dominikschulz woah, super neat! I'll take a look later this evening and test the generated zsh completion.

Foxboron:
This is superseded by #565

PR #573, opened 2017-12-28
Add create message for jsonapi and corresponding test

martinhoefling:
I've added a message and corresponding implementation for generating / storing secrets via json message. This would allow simple password creation / storage from the browser.


martinhoefling:
@dominikschulz , I am working on the coverage for the jsonapi part as well. The current tests instrument the binary and thus do not contribute to the coverage. I try to migrate most of these tests to tests of the `api.RespondMessage` method, with appropriate store and mocks set up.

dominikschulz:
Thank you very much. I was inclined to ask about that, but given the overall coverage we have right now I didn't think it'd be warranted just yet ;)

PR #578, opened 2017-12-30
Refactored tests from gopass binary instrumentation towards interface tests

martinhoefling:
see comment in #573


martinhoefling:
see comment in #573

martinhoefling:
Some basic tests remain in tests/jsonapi_test.go to make sure that gopass properly passes stdin/out to the jsonapi code. All other cases are now tested in api_test.go.

PR #598, opened 2018-01-03
Added completion install to makefile and documentation updates

Foxboron:
I don't know Makefiles very well, but this might be a solution to the problem. The question is if we should update the documentation to reflect that all the completions can be installed with `make install-completion`?

CC: @SevereOverfl0w  @dominikschulz 

Signed-off-by: Morten Linderud <morten@linderud.pw>


SevereOverfl0w:
This would mean that installation from source wouldn't work for completion functionality.

I'm not sure if the releases at https://www.justwatch.com/gopass/releases/1.6.7/ would work with this change either.

Foxboron:
@SevereOverfl0w Correct, but that's an effect on how go does it's packaging. Installation from source should either consists of `make && make install`, or just `go get` without the added functionality.

Another solution would be to do what docker does: https://docs.docker.com/compose/completion/

My opinion is that this is something your package manager should solve.

PR #612, opened 2018-01-05
gpg.Decrypt: use os.Stdin to allow user to enter gpg password if prompted to

samzilverberg:
fix for https://github.com/justwatchcom/gopass/issues/350

I've tested this on mac, and so far looks good:

setup for test:
1. `gpg --full-gen-key` : with 5 years validity and a password
2. create some new dir, cd to it, ```gopass init --nogit --path=`pwd` --store=test```
3. choose the new created key for this store
4. `gopass insert test/bla -m` and put in some pass in the editor
5. cd to the gopass source dir
6. `make clean build` to get a fresh gopass bin
7. `GOPASS_DEBUG=true ./gopass test/bla` => will now show the prompt

behavior tested:
- if prompted for pass, the prompt shows and you can put your pass
  - if you put wrong pass in prompt, the re-opened prompt works as expected
  - if you choose "cancel", pass prompt is closed, decryption fails as expected, no weird side-effects
  - right pass => decryption works as expected
- if no pass prompt is needed, decryption works as expected, no weird side-effects

I suggest try it manually on linux as well. 
I don't have an idea how to create an automated test case for this atm (i'm also not very experienced with golang yet).


dominikschulz:
Thank you for your PR.
Unfortunately I can't reproduce this issue (Linux, GPG 2.0.22, pinentry-curses).

Which version of GPG are you using and what is your pinentry configuration?

I don't see any issue with merging this PR right now, but I'd like to be able to reproduce it first.

samzilverberg:
just re-tested on another mac and i can reproduce the problem and the fix works.
versions on the mac
```
$ gpg --version
gpg (GnuPG) 2.2.3
libgcrypt 1.8.1

$ pinentry-curses --version
pinentry-curses (pinentry) 1.1.0
Copyright (C) 2016 g10 Code GmbH
```

out of interest i also tried to reproduce the problem (and the fix) on the docker container the gopass repo uses using the same steps.
at first i couldn't, but then tried another trick and was able to, maybe try it as well?

the trick is that after having everything setup the gpg key, pass store and secret, then do:
- export the gpg key via `gpg -a --export-secret-keys ${key-id} > secret_key.priv`,
- delete the key from your key ring
  - `gpg --delete-secret-keys ${key-id}`
  - `gpg --delete-keys ${key-id}`
- now add it again via `gpg --import secret_key.priv`
- `GOPASS_DEBUG=true ./gopass test/bla` will now show the prompt!

PR #637, opened 2018-01-17
Allow unclip notification to be disabled

ouroboros8:
Personally those notifications bug me 😉

Feedback welcome.


ouroboros8:
Tests failing, my bad. Looking at it.

ouroboros8:
Take 2.

metalmatze:
I can definitely see why people would be annoyed sometimes. 
lgtm

dominikschulz:
The request makes sense and we don't want to force desktop notifications on anyone, but don't you think @ouroboros8 it would make more sense to have a single flag that disables all notifications and not only the unclip notifications?

ouroboros8:
That's a good point @dominikschulz. Personally the thing that bugs me specifically about the unclip notification is that it notifies me when everything is working fine; I would still like to be notified if the unclipping fails! But you're right taht this PR is maybe too specific.

The best solution might be to introduce levels of notification verbosity (INFO and ERR?) and have that be a parameter to Notify. Then users could configure NotifyLevel as INFO, ERR or NONE. What do you think? Happy to have a go at implementing if you think that would be worthwhile.

dominikschulz:
That sounds a little bit like over engineering. I would prefer to keep things simple and personally I like the notification when the clipboard was cleared but I can also understand that others don't like it. Not really sure what the best way forward would be, right now.

ouroboros8:
You're probably right about the over-engineering. Just had a quick look at how one could globally disable notifications. How do you feel about having Notify take ctx as an arg? Doesn't feel amazingly elegant to me to be honest, but much better than just repeating the approach in this PR ad nauseam anywhere a notification happens.

dominikschulz:
That sounds like a good idea. Passing a context to Notify sounds like a better approach.

ouroboros8:
New pull request https://github.com/justwatchcom/gopass/pull/690 creates an option that toggles all notifications.

PR #642, opened 2018-01-27
Make ctxutil.IsTerminal actually check if we are using a terminal.

### LINKED ISSUE #640: Inability to use gopass programmatically
Currently when doing something like `gopass show entry-that-does-not-exist` gopass will drop users to a wizard UI, this prevents other tooling from using gopass programmatically.

For example [this](https://github.com/gcoop-libre/ansible-lookup-plugin-pass/) tool lets ansible use pass compatible tools to manage passwords on remote systems. However, gopass hangs when running `show` on a non-existing password.

qbit:
Best I can tell, this was a stub function that never actually set
decided if a terminal was being used or not. This seems to fix it.

Might fix #640 - but needs some testing to be sure.

### DIFF
--- a/utils/ctxutil/ctxutil.go
+++ b/utils/ctxutil/ctxutil.go
@@ -1,6 +1,11 @@
 package ctxutil
 
-import "context"
+import (
+	"context"
+	"os"
+
+	"golang.org/x/crypto/ssh/terminal"
+)
 
 type contextKey int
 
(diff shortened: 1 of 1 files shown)


dominikschulz:
I'm sorry, but this is not how we use contexts in `gopass`. The method is not a stub, it's just extracting a value from the context and returning a default if none was set. The actual detection is here: https://github.com/justwatchcom/gopass/blob/master/context.go#L62

qbit:
Ya, that was definitely the wrong place for me to add the check, however, the issue still remains. With the check in it’s current location gopass hangs when using tools like absible. I will add more details to #640

Issue #638, opened 2018-01-21
key: value support of data mapping in non-yaml secrets


notandy:
Hey,

I just moved from browserpass to gopassbridge and experienced that login/email/username key: value in the body part of the secret doesn't work anymore. A lot of software like Android/IOS/Browserpass/etc pass compatible software uses the naive key: value metadata description.

I've seen gopass json api only provides metadata if the secret is stored in yaml format. I have successfully implemented a very easy key: value parser to fill the data map of the secret datatype if yaml parsing fails.
see https://github.com/notandy/gopass

But yaml is really tightly integrated, I am not sure how cleanly implement the classical approach in parallel. My implementation breaks tests and expectations.

As of #559 the yaml support itself is quite buggy, even though I appreciate the work invested having yaml support in gopass, I would really like seeing the yaml parser dropped and  instead having a basic key: value parser. (which should also be able to parse the yaml key/value entries)

dominikschulz:
As state in #559 I'm aware of these issues and I would like to improve the situation. Moving to a simple key:value parser would solve some of these issue and I like the idea, but it would break a lot of stuff in other places. So the problem is not so much the implementation but rather the migration path.

notandy:
:+1:

Issue #559, opened 2017-12-23
Yaml parsing assumes nested values for non-string types


martinhoefling:
Assume you have  a secret in the following form:

``` yaml
mypassword
---
user: 1234
url: http://my.url.com
```

The user yaml value cannot be obtained since gopass is complaining about nested values. This is due to the [interface assertion in secret.go](https://github.com/justwatchcom/gopass/blob/c5a8e526ed0066909e47d7985d4675c433378d0a/store/secret/secret.go#L149). Non string types, like bool, float and int are not covered here.

Several fixes are thinkable:

1. parse with regexp instead of yaml. This would be robust but would prevent extending yaml features in the future.
2. implement conversion to string in the yaml parser. That would make it a string independent on the method you're using to access the value. It also works with values like
3. Extend the `Value` method to support other types. That would work out for most cases but e.g. the conversion to float and back to string could suffer from concatenation: `1.23400` would translate to `1.234` which is probably no transparent to the user. Same goes for 0815
4. We can also leave it as it is but make clear to the user that those special values must be put in double quotes.

What are your thoughts? I would go for 2. or 3. and can pick up the issue if desired.

dominikschulz:
The problem is that most people don't understand YAML. And they shouldn't have to. The question I'm asking myself is if it's better have proper YAML support or a proper UX. I'd go for UX here.

IMHO what we should do is to think about dropping the YAML parser. Parsing key-value pairs should be easy and robust even without regexp. I'd only need to think about marshaling and handling of more complex data for a moment, e.g. we use YAML folded style/heresocs in a few secrets.

martinhoefling:
Just as reference:
Problematic cases are 

```yaml
mypassword
---
user: 1234.131100
```
If you parse as number, you end up with `1234.1311` as username which is probably wrong for most users.

dominikschulz:
Thanks for the example. I fully aware of the nature of this problem. I'm just not sure on what the best approach would be.

dominikschulz:
See also #659

PR #648, opened 2018-02-04
Clarify how yaml attributes are specified

hermannsblum:
This may be quite trivial to any existing pass user, but for me just trying your software in combination with gopass-bridge, it was very unclear that the yaml attributes require the `---` separator.
I therefore suggest to add a simple example to the documentation, as proposed.
Thanks for your awesome work 😊

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -250,7 +250,13 @@ gopass jsonapi configure --browser chrome --path /home/user/.local/
 
 ```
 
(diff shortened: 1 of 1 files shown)


dominikschulz:
Thank you very much.

We're not completely happy with the current state of YAML support, as the UX is not great, but it'll probably take some time until we implement a proper solution for #638 and #559 .

PR #651, opened 2018-02-12
Remove blank line after "Copied <secret> to clipboard..." message.

adamheins:
Currently there's an extra blank line printed after the "Copied <secret> to clipboard. Will clear in <x> seconds." message. This PR removes that blank line.


dominikschulz:
Actually the blank line there was intended to make the output more readable.
Why do you want to remove it?

adamheins:
It seems to be inconsistent with other messages, for example if I just print the secret to the terminal. See the image below:

blank_line_example

PR #659, opened 2018-02-18
Implement a simple, robust and predictable KV parser

### LINKED ISSUE #559: Yaml parsing assumes nested values for non-string types

### LINKED ISSUE #638: key: value support of data mapping in non-yaml secrets

dominikschulz:
This commit replaces the YAML parser with a very simple key value parser that doesn't suffer from YAML's weaknesses.

Fixes #559
Fixes #638


dominikschulz:
@notandy I used your code from #638 as a starting point and implemented a new K/V parser with YAML fallback support.

@martinhoefling This should cover #559 as well.

notandy:
@dominikschulz works well with gopass-bridge, thanks!

Issue #653, opened 2018-02-13
Pluggable encryption


rendaw:
I came here after looking at `pass` and seeing `gpg` tossed everywhere.  I have a Trezor which has offloaded encryption and decryption functions exposed via protobuf.  Would this be doable without major refactoring in gopass?  95% of the work I wouldn't want to write myself is git management, bash completion, etc.

dewey:
You could take a look at this PR: https://github.com/justwatchcom/gopass/pull/645

It's adding another crypto backend so gpg wouldn't have to be used. This should give you an idea of the work involved if you want to work on it.

dominikschulz:
@rendaw To do this does actually require a major refactoring, as you can see in the PR @dewey mentioned.

Good things is that it's already being done and should be ready in a few short weeks. After #645 is merged it should be a lot easier (if not trivial) to add a different crypto backend. I have not access to this device but I'd be happy to assist you if you want to contribute a backend for that.

rendaw:
Awesome, I'll watch that.  Thanks @dewey @dominikschulz !

And this is off the cuff so I haven't thought about it much, but if you were interested and had it on an Amazon wishlist I might get it for you.

dominikschulz:
This should be resolved by #645 .

PR #662, opened 2018-02-20
Add Trezor crypto backend

rendaw:
Trezor is a hardware bitcoin wallet that also exposes crypto primitives.  The developer (Satoshi Labs) made a password manager with this for example: https://trezor.io/passwords/

Here's basically how this works:
1. Create a secret
2. The encrypt method sends the encrypt request to the device
3. The trezor requests a pin and shows a keypad on the screen
4. The user enters the pin by position on the computer, the sequence is sent to the trezor
5. The trezor asks the user to confirm the operation on the device (physical button)
6. The trezor encrypts the text and sends the result back

Decryption is mostly the same in terms of flow.  The trezor stays unlocked for a few minutes, so rapid encrypts/decrypts just require the user to press the confirm button.

Some notes on the backend:

* The trezor api doesn't have an easy way to get the stored private key fingerprint, so it's possible if someone has a bunch connected and they start gopass for the first time they'll have to go through and unlock each in turn so gopass can build a device id <-> key id mapping.  I serialize a dictionary of keys with the ciphertext in case they expose a fingerprint later, the fingerprint can be added in a backward compatible way I think.
* The trezor_encrypt dependency (maintained by me) depends on gotk3 for pinentry which is a bit slow to build.  I think this should run fine on systems without gtk3 if the trezor functionality isn't invoked
* I'm not entirely clear what a bunch of the methods I didn't implement do.  I assume they're related to teams?  Trezor keys can't be passed around too easily so multiple encryption is probably not possible, if that's related.
* Devices have a label, a device id, and a private key.  I added the device id to the recipient name with the label since I didn't see another way to figure out which device the label corresponded to in the encrypt function.
* Trezor doesn't add any key id to the encrypted payload so I had to do that myself with the `Encrypted` struct.
* It's possible to set up a Trezor firmware emulator but I thought that was outside the scope of unit tests.  It didn't seem like unit tests would be simple to add here, but if you have suggestions I could try something.

This works AFAICT but it's a first draft.  I plan to try it myself for a bit, but I'd appreciate any comments/suggestions in the meantime (or perhaps it looks mostly good enough to go?)

Also #653 seems to be resolved in my book with the experimental crypto work.


dominikschulz:
Wow, this is huge. Thank you very much. I'll try to have a thorough look as soon as possible.

rendaw:
:+1: 
I'm afraid I don't know what the latest build error means.  I'll look more tomorrow, gotta run now.

rendaw:
So AFAICT there seem to be problems sourcing Darwin libraries for cross compile for the HID library and static linking issues with GTK (on a limb, I think the Ubuntu packages don't have static libraries).

I'm not to experienced in this area, but I have a couple ideas:

1. Conditionally compile trezor support, disabled by default
2. Fork gopass to add trezor support
3. Make trezor support a plugin built separately (https://golang.org/pkg/plugin/ ?)

rendaw:
I was pointed to https://github.com/romanz/trezor-agent which seems to be a thorough gpg Trezor integration I had missed before.  The advantages of using that is it doesn't require a pin to encrypt and it has a standard key format (so you can export the private key it looks like).

This PR would make using the Trezor simpler, but not significantly.  Is there merit to keeping this open?

dominikschulz:
So far I've only managed to review parts of it. The first change that caught my attention was that you enabled CGO. Unfortunately CGO adds lot's of complexity for us, especially when cross-compiling releases. So I'm afraid we won't merge any PR that requires CGO (one of the reasons why we still need to shell out to git). If your PR depends on CGO than I currently don't see a way to merge it.

If CGO is not strictly necessary we could ponder if we still want to merge it after CGO is gone.

rendaw:
Roger.  CGO is necessary both for the GUI bindings and HID bindings for communicating with the trezor.

PR #690, opened 2018-03-02
Allow notifications to be disabled

ouroboros8:


config/store_config.go
```diff
@@ -24,6 +24,7 @@ type StoreConfig struct {
 	StoreBackend  string `yaml:"storebackend"`  // storage backend (e.g. FS, K/V, ...)
 	SyncBackend   string `yaml:"syncbackend"`   // sync backend (e.g. GitCLI, GoGit, ...)
 	UseSymbols    bool   `yaml:"usesymbols"`    // always use symbols when generating passwords
+	Notify        bool   `yaml:"notify"`        // enable desktop notifications
```
dewey: Maybe it's just me but in a config option to just call it Notify (like the internal library) could be confusing. Maybe EnableNotifications, or just Notifications would be clearer to the outside?

ouroboros8: Agree! Let's go with Notifications, as it's a bit more concise.

utils/notify/notify_darwin.go
```diff
@@ -3,13 +3,16 @@
...
 	"os/exec"
+
+	"github.com/justwatchcom/gopass/utils/ctxutil"
 )
 
 // Notify displays a desktop notification using osascript
-func Notify(subj, msg string) error {
-	if nn := os.Getenv("GOPASS_NO_NOTIFY"); nn != "" {
+func Notify(ctx context.Context, subj, msg string) error {
+	if os.Getenv("GOPASS_NO_NOTIFY") != "" || ctxutil.IsNotifications(ctx) != true {
```
dominikschulz: You should not compare to a bool constant, better: `!ctxutil.IsNotifications(ctx)`

utils/notify/notify_linux.go
```diff
@@ -3,14 +3,16 @@
...
 
 	"github.com/godbus/dbus"
+	"github.com/justwatchcom/gopass/utils/ctxutil"
 )
 
 // Notify displays a desktop notification with dbus
-func Notify(subj, msg string) error {
-	if nn := os.Getenv("GOPASS_NO_NOTIFY"); nn != "" {
+func Notify(ctx context.Context, subj, msg string) error {
+	if os.Getenv("GOPASS_NO_NOTIFY") != "" || ctxutil.IsNotifications(ctx) != true {
```
dominikschulz: You should not compare to a bool constant, better: `!ctxutil.IsNotifications(ctx)`

utils/notify/notify_windows.go
```diff
@@ -3,13 +3,16 @@
...
 	"os/exec"
+
+	"github.com/justwatchcom/gopass/utils/ctxutil"
 )
 
 // Notify displays a desktop notification through msg
-func Notify(subj, msg string) error {
-	if nn := os.Getenv("GOPASS_NO_NOTIFY"); nn != "" {
+func Notify(ctx context.Context, subj, msg string) error {
+	if os.Getenv("GOPASS_NO_NOTIFY") != "" || ctxutil.IsNotifications(ctx) {
```
dominikschulz: Missing a negation? This would make more sense: `!ctxutil.IsNotifications(ctx)`

ouroboros8:
😄

PR #699, opened 2018-03-05
fixes missing bash binary

peel:
On some systems /bin/bash might not exist (ie nixos) or point to non-default shell (ie. macos) hence it should be picked from environment. Thus the change.


peel:
@dominikschulz /bin/sh seems to be working fine. Conceptually if not using bash-specific syntax should be fine as well. fixed that.

PR #705, opened 2018-03-09
 Fixes #667 and correct a few details 

### LINKED ISSUE #667: Disable copy to clipboard after generate
The current behaviour (version 1.6.10) seems to be that it copies the generated password to the clipboard after generating it with `gopass generate <name>`. This behaviour is undesirable because it gets picked up and stored by most clipboard managers.

Some alternatives could be:

- only copy to primary selection which most password managers don't (or can be configured to not) store.
- disable copying to clipboard altogether and let the user manage it

AnomalRoil:
This primarily adds the config option requested in #667.

I've added back the notion or erroring in case of a show action on an
entry with only a password and the safe content option enabled.

Also I've fixed the way the String() function of the StoreConfig forgot
to display the Notification value in its output.

I've also handled the merge conflict caused by  #703.

Now, this does NOT work for the generate action, for a reason I do not yet fully understand. I hope you'll be able to help me there.
In the show action, the context seems to correctly use the config values set. 

However, when running the Generate action, `ctxutil.HasAutoClip(ctx)` returns `false`, even if `gopass config autoclip` returns `true`.
As per the default value of AutoClip, `IsAutoClip` will return `true` is `HasAutoClip` is `false`...

This makes no sense to me, it's just like if the Generate action had not loaded the context correctly.

### DIFF
--- a/action/config_test.go
+++ b/action/config_test.go
@@ -39,6 +39,7 @@ func TestConfig(t *testing.T) {
 	assert.NoError(t, act.Config(ctx, c))
 	want := `root store config:
   askformore: false
+  autoclip: true
   autoimport: true
   autosync: true
   cliptimeout: 45
@@ -76,6 +77,7 @@ foo/nopager: false`
 	act.printConfigValues(ctx, "")
 	want = `root store config:
   askformore: false
+  autoclip: true
   autoimport: true
   autosync: true
   cliptimeout: 45
@@ -88,6 +90,7 @@ foo/nopager: false`
 	want += `  safecontent: false
   usesymbols: false
 mount 'foo' config:
+  autoclip: false
   autoimport: false
   autosync: false
   cliptimeout: 23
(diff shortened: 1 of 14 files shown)


AnomalRoil:
It seems that the Generate action does not use the correct context and has not the correct settings.

I suspect this might be related to the newly added #701. I've tried using Store.GetContext, to no avail...

AnomalRoil:
Alright, I've found it: when Setting the password in the store, the correct Store context wasn't retrieved since there was no "SetContext" function doing the same as the "GetContext" does.

dominikschulz:
Yes, I was just about the state this but you beat me by a few minutes ;)

The PR looks good. Thanks for fixing some existing flaws :+1:

AnomalRoil:
I'm afraid I forgot some tests as Codecov said. 
Here they are:
https://github.com/AnomalRoil/gopass/commit/4b4f00efb6275f69463dfa5bc9a4031a8689961a

dominikschulz:
Alright, I'll just pull them from there. Thanks a lot!

PR #707, opened 2018-03-11
Unbreak Makefile

eli-schwartz:
This fixes an issue where distro builds violated an expectation that software should be built when building and installed while installing. Instead, gopass was repeatedly rebuilt during the packaging phase, which is not supposed to modify the release artifacts.

This is probably not very enjoyable even without it violating distro packaging guidelines...

By making proper use of targets, we can also ensure that simply running `make` will rebuild the binary iff it needs to be rebuilt, so this serves the additional utility of avoiding extra work *and* making it obvious if changes to the source files have been made since the last rebuild.

/cc @Foxboron

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ go:
   - '1.10'
 
 script:
-  - make all
+  - make travis
   - make test-integration
   - make install-completion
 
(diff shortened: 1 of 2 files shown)


dominikschulz:
Thank you very much for this PR. While I disagree with the title (technically the Makefile is not broken), the changes make perfect sense!

eli-schwartz:
Eh, probably could have reworded that title, but it was 2AM (3AM if you count daylight savings time issues), and I was fighting with the fact that 1) it doesn't seem to pick up vendor/ without workarounds grabbed from the Arch build script, and 2) the binary instantly segfaults for reasons that eventually turned out to be "using a pie-enabled go compiler".

dominikschulz:
No worries at all, I can relate. Just wanted to document the fact that the Makefile wasn't completely broken before ;)

PR #738, opened 2018-04-06
Bugfix for Secret.String() not adding a blank second line

phillipj:
Previously when having a secret containing only one line of text,
retrieving the entire secret contents with `.String()` appended
a blank empty second line.

These changes fixes that by checking if there in fact is more than one
line / `.Body()` exists before appending it.

The expected `.String()` output before applying a fix:

```
=== RUN   TestParse
--- FAIL: TestParse (0.00s)
    secret_test.go:215:
            Error Trace:    secret_test.go:215
            Error:          Not equal:
                            expected: "password"
                            actual  : "password\n"
            Test:           TestParse
            Messages:       Simple Secret
```

Does this sound reasonable?

### DIFF
--- a/pkg/store/secret/secret.go
+++ b/pkg/store/secret/secret.go
@@ -66,8 +66,12 @@ func (s *Secret) Bytes() ([]byte, error) {
 func (s *Secret) String() string {
 	var buf strings.Builder
 	_, _ = buf.WriteString(s.password)
-	_, _ = buf.WriteString("\n")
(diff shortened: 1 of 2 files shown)


dominikschulz:
Thanks a lot for reproting and fixing this!

Please double check that your files are gofmt'd. If the tests pass we can merge it.

phillipj:
Woopsie, sorry about that! Fixed a gofmt issue in the test file, amended and force pushed.

Issue #740, opened 2018-04-07
use gopass as git credential helper


veecue:
Hi,
I want to use gopass as a backend to automatically store my git credentials.
As an inspiration, there is [pass-git-helper](https://github.com/languitar/pass-git-helper)
According to this [page](https://git-scm.com/docs/api-credentials), I would propose the following:
- Write a parser/compositor for the [git-credential format](https://git-scm.com/docs/git-credential)
- Add a new command `git-credential` with the subcommands `get`, `store` and `erase`, which parses the input credential, matches it against a stored password and calls the specific action
Does anyone have any other suggestions? I would be willing to work on an implementation

Cheers,
veecue

dominikschulz:
Your PR has been merged. Thanks a lot!

PR #743, opened 2018-04-09
implemented git credential caching

veecue:
See #740 
TODO:
- [x] Write some kind of tests?
- [x] Do not overwrite existing values with the same value
- [x] Properly resolve autosync hangup
- [ ] Use a better and more configurable way to create storage paths
- [ ] Add the possibility to use multiple accounts on one host

### DIFF
--- a/pkg/action/git-credential.go
+++ b/pkg/action/git-credential.go
@@ -102,10 +102,14 @@ func parseGitCredentials(r io.Reader) (*gitCredentials, error) {
 
 // GitCredentialBefore is executed before another git-credential command
 func (s *Action) GitCredentialBefore(ctx context.Context, c *cli.Context) error {
+	err := s.Initialized(ctx, c)
+	if err != nil {
+		return err
+	}
 	if !ctxutil.IsStdin(ctx) {
 		return ExitError(ctx, ExitUsage, nil, "missing stdin from git")
 	}
-	return s.Initialized(ctx, c)
+	return nil
 }
 
 // GitCredentialGet returns a credential to git
@@ -119,24 +123,24 @@ func (s *Action) GitCredentialGet(ctx context.Context, c *cli.Context) error {
 	path := "git/" + fsutil.CleanFilename(cred.Host) + "/" + fsutil.CleanFilename(cred.Username)
 	if !s.Store.Exists(ctx, path) {
 		// if the looked up path is a directory with only one entry (e.g. one user per host), take the subentry instead
-		if s.Store.IsDir(ctx, path) {
-			tree, err := s.Store.Tree(ctx)
-			if err != nil {
-				return ExitError(ctx, ExitDecrypt, err, "Error: %v while listing the storage", err)
-			}
-			sub, err := tree.FindFolder(path)
-			if err != nil {
-				return ExitError(ctx, ExitDecrypt, err, "Error: %v while listing the storage", err)
-			}
-			entries := sub.List(0)
-			if len(entries) == 1 {
-				path = "git/" + entries[0]
-			} else {
-				fmt.Fprintln(os.Stderr, "gopass error: too many entries")
-			}
-		} else {
+		if !s.Store.IsDir(ctx, path) {
 			return nil
 		}
+		tree, err := s.Store.Tree(ctx)
+		if err != nil {
+			return ExitError(ctx, ExitDecrypt, err, "Error: %v while listing the storage", err)
+		}
+		sub, err := tree.FindFolder(path)
+		if err != nil {
+			// no entry found... this is not an error
+			return nil
+		}
+		entries := sub.List(0)
+		if len(entries) != 1 {
+			fmt.Fprintln(os.Stderr, "gopass error: too many entries")
+			return nil
+		}
+		path = "git/" + entries[0]
 	}
 	secret, err := s.Store.Get(ctx, path)
 	if err != nil {
@@ -144,7 +148,7 @@ func (s *Action) GitCredentialGet(ctx context.Context, c *cli.Context) error {
 	}
 	cred.Password = secret.Password()
 	username, err := secret.Value("login")
-	if err != nil {
+	if err == nil {
 		// leave the username as is otherwise
 		cred.Username = username
 	}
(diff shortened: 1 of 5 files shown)


pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/justwatchcom/gopass/pkg/store/sub"
```
dominikschulz: Missing gofmt?

pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+
+type gitCredentials struct {
+	Protocol string
+	Host     string
+	Path     string
+	Username string
+	Password string
+}
+
+func (c *gitCredentials) WriteTo(w io.Writer) (int64, error) {
```
dominikschulz: Please add a godoc comment on this method

pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+	return n, nil
+}
+
+func parseGitCredentials(r io.Reader) (*gitCredentials, error) {
+	rd := bufio.NewReader(r)
+	c := &gitCredentials{}
+	for {
+		key, err := rd.ReadString('=')
+		if err == io.EOF {
```
dominikschulz: I'd prefer to write it as ```if err != nil {
  if err == io.EOF { ... }
}```

pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+	return n, nil
+}
+
+func parseGitCredentials(r io.Reader) (*gitCredentials, error) {
+	rd := bufio.NewReader(r)
+	c := &gitCredentials{}
+	for {
+		key, err := rd.ReadString('=')
```
dominikschulz: Why not just do
```
line, err := rd.ReadString('\n')
kv := strings.SplitN(line, "=", 2)
key = strings.TrimSpace(kv[0])
value = strings.TrimSpace(kv[1])
```
I admin that it's not much shorter, but that's the way I usually do this.

pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// GitCredentialGet returns a credential to git
+func (s *Action) GitCredentialGet(ctx context.Context, c *cli.Context) error {
+	cred, err := parseGitCredentials(os.Stdin)
+	if err != nil {
+		return ExitError(ctx, ExitUnsupported, err, "Error: %v while parsing git-credential", err)
+	}
+	path := "git/" + fsutil.CleanFilename(cred.Host) + "/" + fsutil.CleanFilename(cred.Username)
+	if !s.Store.Exists(ctx, path) {
```
dominikschulz: Some comments on this would be nice. I assume you're first checking the root store and then doing a search which should find any matching entry in a mounted substore?

pkg/action/git-credential.go
```diff
@@ -0,0 +1,189 @@
...
+	return nil
+}
+
+// GitCredentialStore stores a credential got from git
+func (s *Action) GitCredentialStore(ctx context.Context, c *cli.Context) error {
+	cred, err := parseGitCredentials(os.Stdin)
+	if err != nil {
+		return ExitError(ctx, ExitUnsupported, err, "Error: %v while parsing git-credential", err)
+	}
+	path := "git/" + fsutil.CleanFilename(cred.Host) + "/" + fsutil.CleanFilename(cred.Username)
```
dominikschulz: This will always use the root store. For me that's ok, but I bet some users will complain ;)

pkg/action/git-credential.go
veecue: I used goimports, which handles it like that...

pkg/action/git-credential.go
veecue: Can't you just mount a substore to git?

pkg/action/git-credential.go
veecue: It just seemed more straightforward to read the format exactly as it is described in the man page

pkg/action/git-credential.go
```diff
@@ -0,0 +1,192 @@
...
+	secret := secret.New(cred.Password, "")
+	if cred.Username != "" {
+		_ = secret.SetValue("login", cred.Username)
+	}
+	// TODO: Skip this step if the same secret is already in the store. How to implement this?
```
veecue: How can comparison to existing keys be done?

veecue: No, I'm looking for a way to check if the secret that would be stored equals the secret already stored by encrypting the current secret and comparing it to the stored secret (no need to decrypt the stored secret here)

dominikschulz: This is not supported. I don't think this edge case would be enough to extend the (already pretty big) store interface. If you want to compare it the old value will need to be decrypted.

pkg/action/git-credential.go
```diff
@@ -0,0 +1,192 @@
...
+	ctx = sub.WithAutoSync(ctx, false)
+	cred, err := parseGitCredentials(os.Stdin)
+	if err != nil {
+		return ExitError(ctx, ExitUnsupported, err, "Error: %v while parsing git-credential", err)
+	}
+	// try git/host/username... If username is empty, simply try git/host
+	path := "git/" + fsutil.CleanFilename(cred.Host) + "/" + fsutil.CleanFilename(cred.Username)
+	if !s.Store.Exists(ctx, path) {
+		// if the looked up path is a directory with only one entry (e.g. one user per host), take the subentry instead
+		if s.Store.IsDir(ctx, path) {
```
metalmatze: Negate and return early:
```go
if !s.Store.IsDir(ctx, path) {
	return nil
}
...
```
Remove the else case below.

pkg/action/git-credential.go
```diff
@@ -0,0 +1,192 @@
...
+			tree, err := s.Store.Tree(ctx)
+			if err != nil {
+				return ExitError(ctx, ExitDecrypt, err, "Error: %v while listing the storage", err)
+			}
+			sub, err := tree.FindFolder(path)
+			if err != nil {
+				return ExitError(ctx, ExitDecrypt, err, "Error: %v while listing the storage", err)
+			}
+			entries := sub.List(0)
+			if len(entries) == 1 {
```
metalmatze: Negate and return early:

```go
if len(entries) != 1 {
	fmt.Fprintln(os.Stderr, "gopass error: too many entries")
}
path = "git/" + entries[0]
```
Not a major issue in this case, but would prefer it that way.

dominikschulz:
@veecue I'd love to see this merged. Do you plan to continue working on this PR?

veecue:
Yes, I do. However, I'm not shure if the last two items on my TODO-List are necessary at all

dominikschulz:
I think addressing our review comments and the first three bullet points should be enough.

dominikschulz:
@veecue That looks pretty good already. Could you rebase (or merge master) again, so we can merge this as an alpha feature?
Thanks a lot!

PR #745, opened 2018-04-11
Fix JSON API wrapper shebang

### LINKED ISSUE #744: JSON API wrapper assumes `bash` path
I know, right? What messed up distro doesn't have `bash` in `/bin`? Well, NixOS does. In fact, you shouldn't assume to know the path of anything. That's just good practice in general, but is basically enforced on NixOS.

The shebang in the wrapper script doesn't work, because `bash` on NixOS is actually at `/run/current-system/sw/bin/bash`. Which is a symlink to the currently system-wide activated `bash` version. But I digress.

You should either call `$(which bash)` to get the correct path (for any executable, in fact), or use `/usr/bin/env bash`.

mkaito:
It's bad practice to assume you know the path of anything when writing
a portable script. That's what `/usr/bin/env` is for.

Fix #744

### DIFF
--- a/pkg/jsonapi/manifest/constants.go
+++ b/pkg/jsonapi/manifest/constants.go
@@ -1,6 +1,6 @@
 package manifest
 
-var wrapperTemplate = `#!/bin/sh
+var wrapperTemplate = `#!/usr/bin/env bash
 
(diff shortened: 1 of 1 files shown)


martinhoefling:
I would also go for that since the script is posix compatible (checked via posh) and `/bin/sh` should be present and posix compatible as well.

PR #749, opened 2018-04-13
Allow running native hosts messaging as default when binary is renamed

martinhoefling:
This change uses `jsonapi listen` as default action if the binary is renamed / copied to gopass_native_host or gopass_native_host.exe . This makes gopass compatible with windows browsers, which require an executable and not a script to be called.

I would test this a bit with some windows users and add automated setup and / or documentation later on. Does that make sense?

### DIFF
--- a/app.go
+++ b/app.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"os"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


dominikschulz:
LGTM, Thanks a lot!

PR #761, opened 2018-05-03
Audit freshly generated passwords

muesli:
Even generated passwords should be audited, since nothing stops the random
generator from creating one called "password" (if you only try long enough or
are particularly unlucky).

### DIFF
--- a/pkg/pwgen/pwgen.go
+++ b/pkg/pwgen/pwgen.go
@@ -8,6 +8,8 @@ import (
 	"math/rand"
 	"os"
 	"time"
+
+	"github.com/muesli/crunchy"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/pwgen/pwgen.go
```diff
@@ -41,12 +43,18 @@ func GeneratePassword(length int, symbols bool) string {
 // GeneratePasswordCharset generates a random password from a given
 // set of characters
 func GeneratePasswordCharset(length int, chars string) string {
-	pw := &bytes.Buffer{}
-	for pw.Len() < length {
-		_ = pw.WriteByte(chars[randomInteger(len(chars))])
+	var password string
+
+	validator := crunchy.NewValidator()
+	for i := 0; i < 3 && validator.Check(password) != nil; i++ {
```
dominikschulz: You rely on the fact that the validator complains if the inital password is empty?

muesli: Correct. Could be done "cleaner" with a couple more lines of code, but then I'm not sure it's actually more readable. What do you think?

dominikschulz: I think it's fine. Just wanted to understand your reasoning.

PR #779, opened 2018-05-14
Do not overwrite local manifest in tests by calling jsonapi configure

### LINKED ISSUE #778: "make test" overwrites chrome app manifest
When running make test, the installed app manifest of the current user is overwritten. It looks like for some reason SetUp in jsonapi/manifest/setup.go is called with default params.

martinhoefling:
Fixes #778


martinhoefling:
Sure, as I said, maybe it makes more sense to not write by default.

PR #798, opened 2018-05-23
Mention minimum required Go version in README

muesli:
Since we use `strings.Builder` we depend on Go 1.10+.


dominikschulz:
Thanks!

PR #803, opened 2018-05-28
WIP: Parallel stream encryption/decryption

dominikschulz:


dominikschulz:
The custom crypto backend is likely going away.

PR #833, opened 2018-06-03
Fix issue with config values missing from app context

belak:
When I started experimenting with gopass I was getting frustrated with having a pager open when I didn't want it to, so I tried setting it in a config file. However, it looks like these values were being ignored because they never made it into the app context.

This loads all config values into the app context which have an available helper in ctxutil.

### DIFF
--- a/context.go
+++ b/context.go
@@ -58,5 +58,19 @@ func initContext(ctx context.Context, cfg *config.Config) context.Context {
 		ctx = ctxutil.WithColor(ctx, false)
 	}
 
+	// now that defaults have been set, we need to set values from the config.
+	ctx = ctxutil.WithAskForMore(ctx, cfg.Root.AskForMore)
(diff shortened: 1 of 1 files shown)


belak:
I'm curious what this breaks since the config itself will already be set or have the default values at this point and this only sets values which come from the config. Where would sub-stores be overriding config values?

dominikschulz:
The WithContext method of a sub store config will only overwrite context values which weren't set before. This would break if you'd populate these with the root store confiy

belak:
I guess my confusion is that the first thing done in `app.go` is that `config.Load()` is called so if I'm understanding this right, all the config values here should be their final values. Are config values overridden elsewhere in the code?

belak:
For context, I was aiming to solve #838 with this PR. This seemed like the right fix because the config is currently passed into initContext but never actually used.

dominikschulz:
Yes ... it's a fair approach, but I don't think this one will work. Let's see if we can find a proper solution in #838

PR #863, opened 2018-06-11
Enable notifications on FreeBSD

aperum:
FreeBSD support has been in go-dbus for some while now and the notifictions are working fine, at least for me. I have a hunch that this is also the case for OpenBSD and DragonflyBSD, as they also have support in go-dbus, but i do not own such systems and can't test this, so only FreeBSD is enabled with this commit.


dominikschulz:
Thanks a lot, the change itself looks good, but I'd like to take a little different approach.
If there are multiple UNIX/Linux derivates that support notifications through DBUS I'd rename the `notify_linux.go` to `notify_dbus.go` and include all relevant build tag platforms there.
WDYT?

Also: Please sign-off your commits with `git commit -s`.

dominikschulz:
@aperum We'd just need to 'Signed-Off' lines now. Please see https://github.com/probot/dco#how-it-works

aperum:
I've amended the last commit so that the correct sign-off address is used, hope the build system detects that.

dominikschulz:
The Bot is a bit picky, it wants that on every commit. I'll just squash them.

Issue #899, opened 2018-07-21
jsonapi configure does not support windows


Project0:
### Summary
As today running `gopass jsonapi configure` fails under windows. There are also no templates available in the code.

### Steps To Reproduce
Run `gopass jsonapi configure` under windows.

### Expected behavior
It should create the manifests for the browser

### Environment
- OS: Windows 10
- gopass Version: 1.8.2
- Installation method: gopass repo release binary download

Project0:
I was able to manage the setup for windows. 
One problem appeared: The Native Messaging json config does not allow to set arguments in the `path` setting, so its not allowed to just set `gopass jsonapi listen`. Unfortunately its impossible to use batch here, it doesnt forward the stdin/stdout messages. I managed to fix it by a simple wrapper binary: 

# Wrapper
```go
package main

import (
	"log"
	"os"
	"os/exec"
)

func main() {
	cmd := exec.Command("gopass", "jsonapi", "listen")
	cmd.Stdin = os.Stdin
	cmd.Stdout = os.Stdout
	cmd.Stderr = os.Stderr
	err := cmd.Run()
	if err != nil {
		log.Fatal(err)
	}
}
```
Build for windows: `GOOS=windows GOARCH=amd64 go build -o wrapper.exe wrapper.go`
wrapper.exe needs to be placed within the same directory than the manifest.

# Chrome manifest
win_chrome_gopassbridge.json:
```json
{
  "name": "com.justwatch.gopass",
  "description": "Gopass wrapper to search and return passwords",
  "path": "wrapper.exe",
  "type": "stdio",
  "allowed_origins": [
    "chrome-extension://kkhfnlkhiapbiehimabddjbimfaijdhk/"
  ]
}
```
Register manifest (current working dir): 
`REG ADD "HKCU\Software\Google\Chrome\NativeMessagingHosts\com.justwatch.gopass" /ve /t REG_SZ /d "%CD%\win_chrome_gopassbridge.json" /f`

Project0:
thanks @martinhoefling for the info, this is what i was looking for ;-). I will create a PR for the changes.

Project0:
Hmm, i encountered some problems with the `native_host.exe` catchup. Seems something is different than running with `jsonapi listen`. The store seems not to initialize, but i am unsure why and where it fails. 

Need to cross-test again with the linux client, but so far these are the errors

running as usual with parameters works:
```batch
[...]

PR #904, opened 2018-07-28
Add windows support for jsonapi configure

Project0:
I am still unsure about handling the gopass_native_host.exe binary. 
I tried first the windows symlink approach (mklink), but this requires higher privileges in windows (https://groups.google.com/forum/#!topic/golang-nuts/shRlhV3Fe2E). 

This means also the global install support does not work without that -> user needs to run gopass (or cmd) explicitly with higher privileges (run as admin). 
While this is ok for global, i decided it should be possible to configure the jsonapi for user only without admin rights.

For now it means, if there are new requirements or changes in the api for the web extension the user needs to re-run the configure to update native binary.


Project0:
#899

Project0:
ready to review/merge

Project0:
actually i added the GOOS hack only once a time ;-), but let me check how much we can improve here

pkg/action/jsonapi.go
```diff
@@ -119,12 +54,12 @@ func (s *Action) getLibPath(ctx context.Context, c *cli.Context, browser string,
 	return c.String("libpath"), nil
 }
 
-func (s *Action) getWrapperPath(ctx context.Context, c *cli.Context) (string, error) {
+func (s *Action) getWrapperPath(ctx context.Context, c *cli.Context, defaultWrapperPath string, wrapperName string) (string, error) {
```
dominikschulz: You could change this to `defaultWrapperPath, wrapperName string`.

colemickens:
Where does the binary live? Gopass is not detecting any of my browsers and I'm trying to configure jsonapi manually. Thanks.

colemickens:
Just found this. Cool trick, I was actually thinking of implementing it since I missed that jsonapi had Windows support.

>Windows: The jsonapi setup copies the current gopass binary as wrapper ( gopass_native_host.exe calls directly the listener).

Issue #753, opened 2018-04-20
Windows: Creating local store fails with invalid path


dl1ely:
Windows 10 Pro, using the onboarding wizard, during the creation of a local store (after key generation), the following error occurs:
```
Error: failed to run onboarding wizard: failed to init local store: failed to init store '' at 'gpgcli-gitcli-fs+file://gpgcli-gitcli-fs+file//C:%5CUsers%5CDE00173%5C.password-store': failed to initialize store: failed to write recipients file: mkdir \\C:: The specified path is invalid.: failed to write recipients file: mkdir \\C:: The specified path is invalid.
```
`gopass config` shows:
```
path: gpgcli-gitcli-fs+file://gpgcli-gitcli-fs+file//C:%5CUsers%5CDE00173%5C.password-store
```

Is that value right? First there is a "file://" part which looks ok again, but the part after that looks broken.
I then had a look at `config.yml`, and there i see (only relevant part quoted):
```
root:
    path: gpgcli-gitcli-fs+file://C:%5CUsers%5CDE00173%5C.password-store
```
This looks more valid. But why is the `gopass config` output broken? I also tried wrapping the `path` in `config.yml` in `"`, but that did not change the value of `path` in the `gopass config` output.

The problem seems to be in some kind of (broken) duplication of parts of the paths. `config.yml` looks okayish, but the resulting config value looks broken (duplicated parts and `file//` instead of `file://`).

I also tried setting the path to only the part after `file://`, because i suspected the `gpgcli-gitcli-fs+` part was added programatically, but that lead to an `No gpg binary found: failed to init root store: need path` error.

Any ideas what is going wrong here?

dominikschulz:
Thank you very much for reporting this issue. This may be related to some path conversions for the Windows platform. Unfortunately I still don't have access to a Windows test instance, but I'll try to reproduce on Linux or Mac.

These are probably multiple issues (wrong config output, wrong mkdir args, gpg binary not found), but I'll look into it.

dl1ely:
I would be happy to support with testing on windows, unfortunately i have no deeper knowledge of go. I took a quick look into the codebase, but could not identifiy a problem spot-on.
If i can help somehow, let me know.

dl1ely:
What does the part in front of the first `file://` mean anyway? Is gitcli right? I wanted to setup a local store on my harddrive.

dominikschulz:
[...]

Issue #443, opened 2017-11-03
Fix integration tests on windows


dominikschulz:
Right now the integration tests (`make test-integration`) fail on windows (at least on AppVeyor).

lewisdiamond:
Do you need help debugging and testing on Windows? I have created a VM and I'm trying to set it up internally for people who use Windows.

dominikschulz:
Yes, please. We made some (poorly documented) progress on windows but I'm not completely sure about the current state.

lewisdiamond:
Where do you hang out usually? IRC? I'd be willing to do some PRs to fix the issues I'm facing but since I haven't done much in go I may need some pointers on gopass' architecture.

Project0:
still some issues, as far i can see most is related to wrong paths or failed mock inits: https://ci.appveyor.com/project/Project0/gopass?fullLog=true

Project0:
one issue we run into (in pkg/fsutil/fsutil_test.go), will be fixed with go 1.11 : https://github.com/golang/go/issues/24441#issuecomment-377984978

martinhoefling:
Integration tests will be introduced via #1171

PR #912, opened 2018-08-15
WIP: Fix some windows related path problems

Project0:
* Add wrapper for filepath.Abs, it results to different behaviour between unix and windows
* Fix breaking windows paths while parsing paths to URLs


Project0:
fixes #753, reduces errors in integration tests #443

PR #929, opened 2018-09-15
Offer secret selection on edit command

### LINKED ISSUE #927: Enhance editing existing secrets
### Summary

Editing an existing entry is a bit inconvenient at the moment, I think. If I don't know the exact path of the secret to edit, first I have to search for the path via `find` and then I could edit it copying the printed path and open that path via `edit` command.

### Expected behavior

It would be nice if `edit` command would open same wizard than `show` if the given secret name isn't unique instead of opening an empty editor to create a new secret (which isn't what most users want to do if they type `gopass edit <secret-xy>`, I assume). From within the wizard it should be possible to switch to vim (or whatever your console editor of choice is) for one of the listed secrets via a keyboard shortcut.

I would like to have that possibility to edit one of the listed secrets from within the wizard even if it was opened via `show` command. So we don't have to implement different variants of the wizard, imho.

dominikschulz:
This commit adds the secret selection wizard to
the edit command as well. This is triggered in
case no exact match was found. If the user aborts
the wizard then this is handled as if a new secret
was to be created.

Fixes #927

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -320,7 +320,7 @@ func getCommands(ctx context.Context, action *ap.Action, app *cli.App) []cli.Com
 		},
 		{
 			Name:  "edit",
-			Usage: "Edit new or existing secret",
+			Usage: "Edit new or existing secrets",
 			Description: "" +
 				"Use this command to insert a new secret or edit an existing one using " +
 				"your $EDITOR. It will attempt to create a secure temporary directory " +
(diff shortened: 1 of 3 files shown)


martinhoefling:
I think the behavior of selecting a secret as implemented here is much better. Not sure if creating a new one in the abort case is understood. Maybe one can initialize the secret with a comment or so? Alternatively abort means abort and the command is terminated. Edit implies to me that the secret exists already.

dominikschulz:
Ok, good idea. We originally implemented this as pass did, but I agree. It's confusing. Maybe we should add a switch to edit if you want to (possibly) create a new secret and otherwise fail if it doesn't exist already.

PR #930, opened 2018-09-15
Limit bash completion to current scope

### LINKED ISSUE #926: Enhance auto completion
### Summary

Auto-completion of `show` command is not very clear and helpful at the moment if one uses a large password store or even multiple stores with hundreds of entries. At the moment completion just lists all secrets at once.

### Steps To Reproduce

Just type `gopass show` and hit TAB key.

image

### Expected behavior

For me it would be much more helpful if it would work like bash auto completion for filesystem (e.g. using `ls`). Auto-completion should just take the next path hierarchy into account for auto completion.

If auto completion would take arbitrary parts of the next path hierarchy into account, like `zsh` auto completion for filesystem that would be even better:

gopass_evenbetterautocompletion

dominikschulz:
This commit limits the bash completion output
to the current scope.

Fixes #926

### DIFF
--- a/pkg/action/completion.go
+++ b/pkg/action/completion.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"regexp"
+	"sort"
 	"strings"
 
 	fishcomp "github.com/gopasspw/gopass/pkg/completion/fish"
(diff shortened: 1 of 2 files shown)


dominikschulz:
Thanks for testing. I'll fix the tests, than merge.

PR #935, opened 2018-09-23
simple/tree: use filepath.Separator for Windows compatibility

droseger:
Apparently, PR #421 reverted a Windows compatibility fix by again using "/" to detect the directory separator. Because of this, gopass currently doesn't create a proper tree on Windows (folder\file is recognized as one item).
Using `filepath.Separator` didn't change anything about the output of `go test` but it fixes the tree creation, therefore this reversion must have been unintentional.


dominikschulz:
Thanks for sending this PR.
Unfortunately I can not test this on windows and I'm not sure why #421 changed this in the first place.

Maybe @destructive-dragon or @zwopir can advise.

colemickens:
This also fixes https://github.com/gopasspw/gopass/issues/982 and is hard to imagine it being a bad thing to add. It doesn't hurt on Windows AFAICT -- both `gopass foo/bar` and `gopass foo\bar` still work -- meanwhile this fixes other lookups like via the browser extension.

colemickens:
I really, really would encourage this get merged. I reinstalled Windows and going back to stable releases is disappointing after having been running with this patch for months.

PR #961, opened 2018-10-19
Let TravisCI run against latest Go 1.11 release

muesli:
Automatically tests against the latest available Go 1.11 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
   - git config --global user.email foo.bar@example.org
 
 go:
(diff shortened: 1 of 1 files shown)


muesli:
Hm, I believe the Travis error is unrelated to this change?

dominikschulz:
Not sure, this might actually be a change in `archive/tar` for Go 1.11. Need to investigate.

dominikschulz:
The 1.11 release actually broke some testcases due to the way archive/tar was used.
Fixed in #979

PR #971, opened 2018-11-06
Rewrite tty() by using C ttyname() instead of "/proc/self/fd/0"

### LINKED ISSUE #895: tty() function does not work on NetBSD
### Summary

[tty()](https://github.com/gopasspw/gopass/blob/v1.8.2/pkg/backend/crypto/gpg/cli/utils.go#L53) tries to read symbolic link of `/proc/self/fd/0`.
It is not a symbolic link on NetBSD, but a character special file.  So `GPG_TTY` is not set.

Also, I think OpenBSD has problem too as it does not provide procfs.

### Steps To Reproduce

Similar to #350.
`GPG_TTY` is not set and pinentry-curses does not show prompt.
Then `Failed to decrypt` error happens.

### Expected behavior

`GPG_TTY` is set before gpg is executed.
Workaround is to set `GPG_TTY=$(tty)` manually.

### Environment

- OS: NetBSD
- OS version: NetBSD 8.0_RC2
- gopass Version: 1.8.2
- Installation method: gopass repo

iquiw:
Fixes #895.

I tried to write unit test for this, but failed as stdin is "/dev/null" in `go test`.

Manually tested about `Failed to decrypt` error case on NetBSD 8.0, Linux (Ubuntu 18.04), macOS (Mojave) and worked.

### DIFF
--- a/pkg/backend/crypto/gpg/cli/utils.go
+++ b/pkg/backend/crypto/gpg/cli/utils.go
@@ -7,10 +7,6 @@ import (
 	"time"
 )
 
-var (
-	fd0 = "/proc/self/fd/0"
-)
-
 // parseTS parses the passed string as an Epoch int and returns
 // the time struct or the zero time struct
 func parseTS(str string) time.Time {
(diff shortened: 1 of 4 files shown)


iquiw:
hmm, Travis build failed...

```
>> CROSSCOMPILE darwin/amd64# github.com/gopasspw/gopass/pkg/backend/crypto/gpg/cli
pkg/backend/crypto/gpg/cli/gpg.go:50:11: undefined: tty
```

iquiw:
@dominikschulz OK, closing this. I will consider other options and discuss in #895.

dominikschulz:
@iquiw Thanks for your understanding. Let's try to find another way to fix this.

PR #975, opened 2018-11-10
jsonapi: return totp in the api response to enable 2FA use in gopassbridge

colemickens:
See: https://github.com/gopasspw/gopassbridge/issues/87

This returns the OTP code, if available over the jsonapi.

This is done in a way so as to be minimally invasive: we either return a good totp, or we return the original payload. This has the side effect of swallowing errors, but errors are effectively swallowed in this path anyway due to how `gopassbridge` presents errors.


colemickens:
All green, test added, thanks again for the fast review!

PR #976, opened 2018-11-10
jsonapi: more flexible host matches

colemickens:
For `jsonapi` when querying on host, (what the browser extension does):

1. Search for exact matches (or, as the existing logic was doing, progressively more "base" subdomains - aka - `foo.alpha.beta.com`, then `alpha.beta.com`, then `beta.com`
2. Also search for the same matches, but looser.

This will help users, like myself, that store passwords in directories in broad categories.

This should preserve the existing logic in cases where "better" matches were found, but will still help the user when there are looser matches available that may be desired without having to manually search for the same hostname that *was already queried*. 

For example, right now, the extension will not find `colemickens/github.com` when I am on https://github.com. With this PR, that is resolved.

---

Please let me know if it's more appropriate to change the browser extension side of things. cc: @Pharb

### DIFF
--- a/pkg/jsonapi/responses.go
+++ b/pkg/jsonapi/responses.go
@@ -55,16 +55,22 @@ func (api *API) respondHostQuery(ctx context.Context, msgBytes []byte) error {
 	}
 	choices := make([]string, 0, 10)
 
-	for !isPublicSuffix(message.Host) {
-		// only query for paths and files in the store fully matching the hostname.
-		reQuery := fmt.Sprintf("(^|.*/)%s($|/.*)", regexSafeLower(message.Host))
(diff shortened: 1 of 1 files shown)


martinhoefling:
So, I personally see the risk that you can easily trick the user to enter credentials for a third party site. It is  also not clear to me, why `colemickens/github.com` is not found? Can you supply a unit test for this case?

martinhoefling:
can you rebase onto pr #977? Lets see if these cases are still considered.

martinhoefling:
#980 adds a test that should mimick the situation that you describe in the first post. Looks good for me. Can you check out what's the difference between the test and the problem statement @colemickens?

PR #977, opened 2018-11-10
Api tests

martinhoefling:
Add two new api testcases to cover two edge-cases.


colemickens:
If you change your first example to better match the example I gave in #976.

```diff
- `\["awesomePrefix/other.host/other"\]`,
+ `\["awesomePrefix/other.host"\]`,
```

then you see it fails:

```
Expect "["awesomePrefix/other.host/other"]" to match "\["awesomePrefix/other.host"\]"
```

But then again, it also still fails after my attempted change, so maybe I've got something else wrong?

PR #978, opened 2018-11-11
Add an extra command to setup gopass as git's `credential.helper`

### LINKED ISSUE #891: Document how command `git-credential` works

veecue:
Fixes #891

Signed-off-by: veecue <veecue@ventos.tk>


dominikschulz:
You need to remove Line 31 in this file to fix the tests:  https://github.com/gopasspw/gopass/blob/master/tests/uninitialized_test.go#L31

worldofpeace:
@veecue What organization do the credentials need to be in for this to actually work? For like ssh auth and https.

veecue:
Oops, never meant to close this... fix and reply to @worldofpeace will be pushed tomorrow

veecue:
Well, someone (tm) was too stupid to use git...
@worldofpeace You can see the structure in https://github.com/gopasspw/gopass/blob/master/pkg/action/git-credential.go#L126
The credentials are stored in `git/<host>/<username>`, so if you want to add them manually, this is the way to organize them. (tip: you can mount another store `git` or `git/example.com`). However since the credential cache is intended to be filled by git upon password entry, it is not documented at the moment.
Git's credential-cache feature only works with https (and technically others with basic-auth) so for using SSH, you will have to setup `ssh-agent` or something similar (and yes, I have thought of implementing a version of ssh-agent using gopass)

worldofpeace:
@veecue 
Thanks for the explaination.

PR #988, opened 2018-11-19
Adding autocompletion of the folders for insert and generate in Zsh

AnomalRoil:
This PR is adding a `list --folders` flag that allows to print a **flat** list of the all folders in our password store, allowing to easily `insert` or `generate` new secrets in the right folders.

Then, I've changed the Zsh completion to have a `_gopass_complete_folders` function to call when using `insert` or `generate`, that will allow to autocomplete the path to the secret. 
I'm not using bash, so I didn't change it.

The whole comes with tests for the new features, excepted the actual autocompletion using zsh, since it wasn't tested in the first place.

I'm afraid I've also changed a couple details I stumbled upon: using `text/template` in the zsh autocomplete template instead of html (since it's not html we're working on) and correcting the order of the args in a couple tests too.

Notice the flat folder list is sorted using sort, in order to ease the testing. Zsh is already sorting the whole on its own, so it is not required per se.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -780,6 +780,10 @@ func getCommands(ctx context.Context, action *ap.Action, app *cli.App) []cli.Com
 					Usage: "Print flat list",
 				},
 				cli.BoolFlag{
+					Name:  "folders, fo",
+					Usage: "Print flat list of folders",
+				},
+				cli.BoolFlag{
 					Name:  "strip-prefix, s",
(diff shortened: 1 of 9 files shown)


AnomalRoil:
Sorry for the codequality issues, I changed my dev environment and didn't realize I wasn't auto-running `go fmt` when saving files anymore.

PR #1006, opened 2018-12-05
This reverts regression of OTP in 1.8.3

nokernel:
This is my first attempt at contributing. Sorry if it is not made following standard.

It is also my 1st github project contribution.

This proposal is only tring to make 1.8.3 works again with TOTP tokens in password format : `otpauth://totp/...`

Those lines broke that in version 1.8.3.

Feel free to correct, as I am also a total newb in golang.

### DIFF
--- a/pkg/jsonapi/responses.go
+++ b/pkg/jsonapi/responses.go
@@ -139,8 +139,6 @@ func (api *API) respondGetData(ctx context.Context, msgBytes []byte) error {
 	currentTotp, _, err := otp.Calculate(ctx, "_", sec)
 	if err == nil {
 		responseData["current_totp"] = currentTotp.OTP()
-	} else if err != otp.ErrNoTotpEntry {
-		responseData["current_totp"] = "failed to retrieve: " + err.Error()
 	}
 
 	converted := convertMixedMapInterfaces(interface{}(responseData))
--- a/pkg/otp/otp.go
+++ b/pkg/otp/otp.go
@@ -2,7 +2,6 @@ package otp
 
 import (
 	"context"
-	"fmt"
 	"io/ioutil"
 	"strings"
 
@@ -10,11 +9,6 @@ import (
 
 	"github.com/gokyle/twofactor"
 	"github.com/pkg/errors"
-)
-
-var (
-	// ErrNoTotpEntry signals a failed OTP for a secret with OTP information
-	ErrNoTotpEntry = fmt.Errorf("no totp entry in secret")
 )
 
 // Calculate will compute a OTP code from a given secret
@@ -34,11 +28,9 @@ func Calculate(ctx context.Context, name string, sec store.Secret) (twofactor.OT
 	// check yaml entry and fall back to password if we don't have one
 	label := name
 	secKey, err := sec.Value("totp")
+
 	if err != nil {
 		secKey = sec.Password()
-	}
-	if secKey == "" {
-		return nil, label, ErrNoTotpEntry
 	}
 
 	if strings.HasPrefix(secKey, "otpauth://") {
--- a/pkg/otp/otp.go
+++ b/pkg/otp/otp.go
@@ -34,9 +34,9 @@ func Calculate(ctx context.Context, name string, sec store.Secret) (twofactor.OT
 	// check yaml entry and fall back to password if we don't have one
 	label := name
 	secKey, err := sec.Value("totp")
-	if secKey == "" {
+	/*if secKey == "" {
 		return nil, label, ErrNoTotpEntry
-	}
+	}*/
 	if err != nil {
 		secKey = sec.Password()
 	}
--- a/pkg/otp/otp.go
+++ b/pkg/otp/otp.go
@@ -34,11 +34,11 @@ func Calculate(ctx context.Context, name string, sec store.Secret) (twofactor.OT
 	// check yaml entry and fall back to password if we don't have one
 	label := name
 	secKey, err := sec.Value("totp")
-	/*if secKey == "" {
-		return nil, label, ErrNoTotpEntry
-	}*/
 	if err != nil {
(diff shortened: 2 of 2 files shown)


nokernel:
cc @dominikschulz @colemickens 

This is just a disable what broke 1.8.2 for me. It might be brutal. It did the trick for me.

nokernel:
Much better now updated without commenting out things.

dominikschulz:
It looks like https://github.com/gopasspw/gopass/commit/dec28c396069e6ef16e2b52467a6b4a0e70958a2 broke the OTP handling, indeed.

I'm inclined to just revert that commit, as this commit seems to be breaking the JSONAPI changes that came with it.

nokernel:
> It looks like [dec28c3](https://github.com/gopasspw/gopass/commit/dec28c396069e6ef16e2b52467a6b4a0e70958a2) broke the OTP handling, indeed.
> 
> I'm inclined to just revert that commit, as this commit seems to be breaking the JSONAPI changes that came with it.

That would work :)

Sad because I was hoping to have that OTP through gopass browser plugin ;)

colemickens:
Yes, it looks like I broke the "fallback to password field" if it fails to find the `otpauth://` or `totp:` entries.

This change seems correct (and is hopefully what I'd meant to do originally!).

@dominikschulz, I didn't totally understand your comment. I'm not aware of anything wrong with the json handling, and the commit that touches it relies on the commit where I introduced this bug. Merging this fix seems like the best/fatest/simplest route to me.

Sorry for breaking this case, and thanks for the fix @nokernel .

colemickens:
@nokernel congrats on your first GitHub PR too! That's cool!

nokernel:
> @nokernel congrats on your first GitHub PR too! That's cool!

@colemickens thanks!

colemickens:
@dominikschulz So now I see the problem via the failing Travis job here (https://travis-ci.org/gopasspw/gopass/builds/463839907?utm_source=github_status&utm_medium=notification).

I don't know what I should do about this because apparently this is "valid", in that `gopass` will just to use any old password value as an OTP, which indeed will be invalid in numerous, likely most, cases.

Let me know how you want this to behave, or how you want me to patch up the tests and I can.

colemickens:
I guess I can just blanket ignore any errors when generating the OTP code and just not be able to give the user a better message if the OTP usage specifically fails -- which I guess I can't do anyway given the current expected behavior. :/

Let me know and I can send a PR or send @nokernel another commit it can be merged as their first PR.

nokernel:
> I guess I can just blanket ignore any errors when generating the OTP code and just not be able to give the user a better message if the OTP usage specifically fails -- which I guess I can't do anyway given the current expected behavior. :/
> 
> Let me know and I can send a PR or send @nokernel another commit it can be merged as their first PR.

There was this issue that looked like that, for some reason the YAML format for totp looks to be broken (for me) or I dont know how it used to work.

For me this does not work :
```yaml
totp: tokenhere
```

while this works :
```yaml
---
totp: tokenhere
```

Being bad at golang I was not able to find why do I need that `---` for TOTP to work in the yaml format.

Could it be that the test is missing those `---` also?

colemickens:
@nokernel I don't know. I also find the YAML behavior confusing and have tried to just avoid it. All of my OTP codes are in the URL form that you have, which are found properly via the function we're touching. It may be that `gopass` needs the YAML separator (`---`) to know that the body is actually YAML...

Further to your point, and to the oddities at play, the `totp:` and `otp://` handling is different here... so, yeah, I don't know. I've tried to only touch what I need to because I find a number of things like this confusing in the codebase. Yaml doesn't seem well handled in the browser extension, either.

However, from what I can tell, those are unrelated to the regression I introduced, and would recommend a new Issue if you want to get to the bottom of them. 

---

Anyway, I have a branch here: https://github.com/colemickens/gopass/commits/master
* it passes the `jsonapi` tests (and all tests that don't require env setup, anyway)
* it reverts the commit where I touched the Calculate function
* it adds another commit that fixes the usage of Calculate due to the revert
* manually tested, OTP still shows in browser for me

I'd write a test for this, but again, given the "desired" behavior and the way the tests are written, that isn't as easy as it might seem.

@nokernel Feel free to take my commits, or changes, and apply them here if you want to finish out this PR. (Or whatever @dominikschulz says :smile:)

colemickens:
@nokernel to be more specific, part of your fix causes the OTP Calculate function to fallback to trying the password as an OTP code. In any of the tests where an OTP code wasn't set, my Err change was causing an error to appear in the JSON response.

I assumed that "invalid OTP code" is something we'd want ot surface to the user, but it turns out, the desired behavior, I guess, is to swallow errors, since they will occur any time a password field is not a valid OTP secret.

Personally, I'd never have fallen back like that and treated other random fields as OTP secrets, but I guess it's too late for that.

dominikschulz:
@colemickens Thanks for your input on this issue. I wasn't sure if you're around to comment on this. My concern with the mentioned commit is that you should (almost) never handle return values before checking the error. Also the commit didn't come with tests for the changes which makes it harder to fix without risking collateral damage. Unfortunately we failed to notice this during review. Glad you're around to help with a proper fix.

@nokernel This PR looks good by now, but I don't want to merge this with failing tests. Can you proceed as @colemickens suggested (i.e. take his changes to fix the tests) or do you need help with that?

colemickens:
@dominikschulz Agreed, I wound up actually reverting the commit as you suggested. There are tests for the actual feature change, at least. I've added some basic OTP tests, to prevent this, in another PR: https://github.com/gopasspw/gopass/pull/1008.

nokernel:
> Feel free to take my commits, or changes, and apply them here if you want to finish out this PR.

@colemickens ok I took them, thanks for that hehe, at this point this becomes only a revert ;)

At least this would have helped add more testing.

PR #1031, opened 2018-12-29
Show jsonapi option in help

martinhoefling:
see last comment in https://github.com/gopasspw/gopassbridge/issues/96


dragon788:
Somehow jsonapi fell back out of the `gopass help` options?

PR #1057, opened 2019-02-06
completion: fix zsh completion for folders

vincentbernat:
For `insert`, `generate` and `list`, the folder completion doesn't
work as expected. First, we should continue completion even if we
already have a `/` in the current name. Otherwise, we can only
complete the first level. Then, don't specify `/` as a separator for
`_values`, this is for specifying multiple values.

Moreover, allow completing for passwords for `generate` (ability to
change a password) and for `insert` (ability to append data).

cc @AnomalRoil

### DIFF
--- a/pkg/completion/zsh/template.go
+++ b/pkg/completion/zsh/template.go
@@ -21,8 +21,8 @@ _{{ $prog }} () {
               {{- end }}
               {{ if .Flags }}_arguments :{{ range .Flags }} "{{ . | formatFlag }}"{{ end }}{{ end }}
               _describe -t commands "{{ $prog }} {{ .Name }}" subcommands
-              {{ if or (eq .Name "copy") (eq .Name "move") (eq .Name "delete") (eq .Name "show") (eq .Name "edit") }}_{{ $prog }}_complete_passwords{{ else if or (eq .Name "insert") (eq .Name "generate")  (eq .Name "list") }}if ! compset -P '*/'; then _{{ $prog }}_complete_folders
-              fi{{ end }}
+              {{ if or (eq .Name "insert") (eq .Name "generate")  (eq .Name "list") }}_{{ $prog }}_complete_folders{{ end }}
+              {{ if or (eq .Name "copy") (eq .Name "move") (eq .Name "delete") (eq .Name "show") (eq .Name "edit") (eq .Name "insert") (eq .Name "generate") }}_{{ $prog }}_complete_passwords{{ end }}
               ;;
 {{- end }}
           *)
@@ -52,7 +52,9 @@ _{{ $prog }}_complete_passwords () {
 }
 
 _{{ $prog }}_complete_folders () {
-    _values -s '/' 'folders' $({{ $prog }} ls --folders --flat)
+    local -a folders
+    folders=("${(@f)$({{ $prog }} ls --folders --flat)}")
+    _describe -t folders "folders" folders -qS /
 }
 
 _{{ $prog }}`


vincentbernat:
Could you describe what exactly you type? I don't get that. If I type `gopass generate deb<TAB>`, I get `gopass generate debian` and I can continue with `gopass generate debian/something`. Maybe you want to have a `/` automatically added when we match a folder?

AnomalRoil:
Assume I have the following in my password store:
```
lol
├── b1
│   └── foo
└── zen
```
And let's say I want to add a new entry bar in the folder `lol/b1`, I type `gopass generate lo<Tab x 4>`, it select the entry I want as on this picture:
image

and then I type the name of my entry "bar" and then I get `gopass generate lol/b1 bar` instead of `gopass generate lol/b1/bar`. 

But since `b1` is a folder, we don't want a space inserted after it.   
If it were an existing entry, it would be fine, since we might be wanting to insert a new tag.

AnomalRoil:
See also my commit here: https://github.com/gopasspw/gopass/pull/988/commits/c8824a5f37e8999735220e48b09906ad2d72fd2a

I first tried using a trailing slash in the folder list --flat function directly, but since it would keep inserting a space, I tried moving to the `-s` option, which is also not ideal, I agree.

vincentbernat:
`-qS /` would do that, but I don't know how to use it with values. However, I have a simpler solution, I'll push an additional commit.

vincentbernat:
I have amended the commit. You should now get the `/` at the end of the folder. This was easier than I thought because it is not optional (you cannot call `insert` or `generate` on a folder).

AnomalRoil:
Any reason you're using the `gopass` name explicitly instead of through the
$prog variable?

Anyway, I'll try it and let you know how it works.

Le mer. 6 févr. 2019 à 20:21, Vincent Bernat <notifications@github.com> a
écrit :

vincentbernat:
No, it was a copy/paste error on my side. I've amended the commit to fix that.

AnomalRoil:
Still, Zsh inserts a space when completing with tab. I really don't understand how we could tell it not to do it...
Sadly, the `add-space` zsh style which would have made handling this easy in `$fpath[-1]/_expand` seems to not be used when relying on `_values` (you can see it using zsh's `_complete_debug`).

I've already tried using `_values -S /` and modifying the list to say each entry takes an optional argument, but it wouldn't work neither.
  
That being said, we _could_ work this around by telling the `insert` and `generate` function that if a tag is added to a folder, then it should insert it as a new entry in that folder, but that would imply changing the functionalities of `gopass` to work around a completion bug, without solving the problem.

vincentbernat:
OK, I'll come with another fix tomorrow. The `-S` in for `_values` is not the same than for `compadd` or friends. So, it should be a matter to convert `_values` to `_describe`.

vincentbernat:
I have amended again the commit with something that should now work for you. If you select a folder from the menu, you will now get a slash appended to it and no additional space. Also, folders are completed first, since it makes more sense this way.

vincentbernat:
❦  9 février 2019 03:44 -08, Yolan Romailler <notifications@github.com>:

> On a side-note, I'm curious about the `(@f)` syntax, what does it do?
> I've never encountered it and my Google-fu is failing me there.

The "f" is equivalent to ps:\n: which splits a variable following new
lines. The "@" is to turn the result into an array. More here:
 http://zsh.sourceforge.net/Doc/Release/Expansion.html#Parameter-Expansion

It's useful to impress people but may become very hard to read. ;-)

dominikschulz:
Thanks a lot @vincentbernat. Shell completing is something I keep having trouble with.

PR #1059, opened 2019-02-06
Handle multiple private keys matching an identifier

vincentbernat:
The default behavior of gnupg when multiple keys are matching (like
when using an email address as an identifier) is to use the first one.
We make gopass use the same logic.

In my case, I have two keys matching my email address. The second one
is revoked. gopass is trying, for some reason, to use the second one.
I don't quite understand where it decides that, but making it choose
the first one earlier makes it work.


vincentbernat:
An alternative for me would be to not use a revoked key.

PR #1062, opened 2019-02-08
fs-store: remove empty directories

### LINKED ISSUE #1009: Handle `rmdir` of empty directories when moving or deleting last password from a directory
### Summary

Handle `rmdir` of empty directories when moving or deleting last password from a directory.

### Steps To Reproduce
With this passwords structure
```bash
myname1
    |
    ----myname1
myname2
   |
   ----myname2
```

Do
```bash
gopass mv myname1/myname1 myname2/myname1
gopass mv myname2/myname1 myname1
Error: failed to save secret 'myname1': a folder named myname1 already exists
```

### Expected behavior
gopass shall remove empty directories when moving/deleting last password in a directory.

### Environment

- OS: Linux
- OS version: Ubundu 18.04
- gopass Version: gopass 1.8.3-git+HEAD go1.11.2 linux amd64
- Installation method: go build

### Additional context
This is happening since 1st public version of gopass released.

corneliusweig:
The fs-store backend did not clean up empty directories, so that the following sequence of commands would fail:
$ gopass mv foo/last_file_in_directory bar
$ gopass mv bar foo # fails because foo exists and is a directory

Now, the-fs store backend removes all empty parent directories when a secret is deleted.

Implementation note: because os.Remove will only remove empty directories, this is the fastest and easiest solution.

Fixes #1009

### DIFF
--- a/pkg/backend/storage/fs/store.go
+++ b/pkg/backend/storage/fs/store.go
@@ -8,6 +8,7 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"syscall"
 
 	"github.com/gopasspw/gopass/pkg/fsutil"
 	"github.com/gopasspw/gopass/pkg/out"
@@ -54,7 +55,25 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
 func (s *Store) Delete(ctx context.Context, name string) error {
 	path := filepath.Join(s.path, filepath.Clean(name))
 	out.Debug(ctx, "fs.Delete(%s) - %s", name, path)
-	return os.Remove(path)
+
+	err := os.Remove(path)
+	if err != nil {
+		return err
+	}
+
+	// remove all empty parent directories
+	for dir := filepath.Dir(path); ; dir = filepath.Dir(dir) {
+		err = os.Remove(dir)
+		switch {
+		case err == nil:
+			continue
+		case err.(*os.PathError).Err == syscall.ENOTEMPTY:
+			// ignore when directory is non-empty
+			return nil
+		default:
+			return err
+		}
+	}
 }
 
 // Exists checks if the named entity exists
--- a/pkg/backend/storage/fs/store.go
+++ b/pkg/backend/storage/fs/store.go
@@ -63,6 +63,7 @@ func (s *Store) Delete(ctx context.Context, name string) error {
 	return s.removeEmptyParentDirectories(path)
 }
 
+// Deletes all empty parent directories up to the store root
 func (s *Store) removeEmptyParentDirectories(path string) error {
 	parent := filepath.Dir(path)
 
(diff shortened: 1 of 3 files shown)


pkg/backend/storage/fs/store.go
```diff
@@ -54,7 +55,25 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
 func (s *Store) Delete(ctx context.Context, name string) error {
 	path := filepath.Join(s.path, filepath.Clean(name))
 	out.Debug(ctx, "fs.Delete(%s) - %s", name, path)
-	return os.Remove(path)
+
+	err := os.Remove(path)
+	if err != nil {
```
dominikschulz: I would normally put the err assignment one the same line as the if.

pkg/backend/storage/fs/store.go
```diff
@@ -54,7 +55,25 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
...
+		return err
+	}
+
+	// remove all empty parent directories
+	for dir := filepath.Dir(path); ; dir = filepath.Dir(dir) {
```
dominikschulz: The three form for is uncommon in Go. Please add a comment describing what it does (e.g. "start with parent, then work upwards, ...").

pkg/backend/storage/fs/store.go
```diff
@@ -54,7 +55,25 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
...
+		err = os.Remove(dir)
+		switch {
+		case err == nil:
+			continue
+		case err.(*os.PathError).Err == syscall.ENOTEMPTY:
+			// ignore when directory is non-empty
+			return nil
+		default:
+			return err
+		}
```
dominikschulz: This loop makes me feel uneasy. This reminds me of that one time when I identified a function annotated with "recursive delete, can go up" for deleting a whole production system. While this shouldn't happen I would prefer to limit it strictly to never exceed the password store root dir.

pkg/backend/storage/fs/store.go
corneliusweig: yes, that looks much better :)

pkg/backend/storage/fs/store.go
corneliusweig: You are absolutely right. Even though it _should_ not happen, the effects of this action must remain limited to the gopass managed directory.

pkg/backend/storage/fs/store.go
corneliusweig: :+1:

corneliusweig: With the other requested changes it makes much more sense to turn this loop into a recursive function. It should be much more readable now

PR #1064, opened 2019-02-08
Disable CGO to work around alpine/musl restrictions (#1063)

### LINKED ISSUE #1063: Gopass in docker image cannot be executed
### Summary
The environment for the `gopass` in the provided Dockerfile is broken. When running `docker run -ti gopass-test`, I get
```
standard_init_linux.go:190: exec user process caused "no such file or directory"
```

Before, I could run a sandbox'ed gopass environment with
```
docker build . -t gopass-test
docker run --entrypoint /bin/ash --rm -ti -v $PWD/my-home:/root gopass-test
```
But today I noticed this does not work anymore.

### Steps To Reproduce
```bash
docker build . -t gopass-test
docker run --rm -ti gopass-test  # or
docker run --entrypoint /bin/ash --rm -ti -v $PWD/my-home:/root gopass-test
```
Inside the container, `ldd /usr/bin/gopass` yields a very suspicious output:
```
        /lib64/ld-linux-x86-64.so.2 (0x7f77faddb000)
        libpthread.so.0 => /lib64/ld-linux-x86-64.so.2 (0x7f77faddb000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f77faddb000)
Error relocating usr/bin/gopass: __vfprintf_chk: symbol not found
Error relocating usr/bin/gopass: __fprintf_chk: symbol not found
```

### Environment

- OS: Gentoo Linux
- OS version: Linux 4.18.13
- gopass Version: `gopass 1.8.4 (6c3ae0ec 2019-02-08 22:13:51) go1.11.5 linux amd64`
- Installation method: master tip

corneliusweig:
https://github.com/mattes/migrate/issues/241 describes a very similar issue. One solution is to disable the C language binding (see also https://golang.org/cmd/cgo/).

Fixes #1063

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,7 +5,7 @@ RUN apk add -U make gcc musl-dev ncurses git
 ADD .   /go/src/github.com/gopasspw/gopass
 WORKDIR /go/src/github.com/gopasspw/gopass
 
-RUN TERM=vt100 make install
+RUN CGO_ENABLED=0 TERM=vt100 make install
 
 FROM alpine:3.7
 RUN apk add -U git gnupg


ncoop:
Release 1.8.5 runs on alpine :+1:

PR #1065, opened 2019-02-10
Add AutoPrint option

dominikschulz:
This PR adds an AutoPrint config option.


dominikschulz:
The show action already prints the password, without any special options. So I currently fail to see how that would apply there.

martinhoefling:
You're right, I was mislead by the case for multiple matches, where you can select what to do.

PR #1069, opened 2019-02-15
Add Gopass UI to the list of available GUIs for gopass

jverhoelen:
As it is a pleasure for us to work with gopass in real projects, a fellow employee and me started to work on [Gopass UI](https://github.com/codecentric/gopass-ui). In our team we are already using it heavily because we have some not-so-techie people who are not too command line affine.
We think it's in a good condition to make it public and surely want to put more effort into it.

We would be very happy to get feedback on the UI and functionality across all platforms (especially Windows and Linux as we are MacOS users ourselves). It would be great to have it added to your docs. Feel free to propose a different structure or change the wording.

Cheers,
Jonas


jverhoelen:
I’m glad you like it 😊 Thanks for the fast merge.

PR #1099, opened 2019-04-07
Don't expand tilde (fixes #872)

### LINKED ISSUE #872: Use '~' in config.yml
It would be great to use `~`in the `config.yml` file so that it uses relative resolution instead of absolute resolution. The idea behind this is to be able to commit the config file in a git repository and deploy it on different machines where homedir can be different.
I tested by replacing `/home/<foo>` with `~` in the config file and it works fine. The only thing is that everytime I change configuration, the `~` is replaced back by `/home/<foo>`.

mcanevet:
The idea behind this is to be able to commit the config file in a git repository and deploy it on different machines where homedir can be different.


mcanevet:
@dominikschulz can you have a look at this PR please?

martinhoefling:
Since we changed that behaviour already quite some times, we first have to think about which way it makes most sense.

sudoforge:
For what it's worth, I would love to see this officially supported as well. The current behavior\* is a little odd, since _reads_ of the configuration file will handle tilde expansion just fine, however, new writes to the configuration file expand the tilde in place.

_\*: At least, as far back as the ~two years that I've been using `gopass`_

dominikschulz:
Don't worry, we're not going to ignore this issue.
It's just that this path handling has been causing different issues in the past and we don't want to break another use case by fixing this.

We'd like to come up with proper solution, document it and have proper test cases.
Unfortunately so far nobody has done that.

sudoforge:
Yeah, there's also a story around handling existing YAML anchors when writing the configuration with similar behavior, but this is obviously more complex -- I'll give a think to this over the next week or so and see if I can't come up with a good story for it.

dominikschulz:
Thank you for your contribution.
Unfortunately we can not merge this PR at this time.
However we'll keep track of this issue in #872 and try to implement a proper fix.

Issue #1005, opened 2018-12-04
failed to retrieve secret on substore


Dian4TI:
### Summary
Unable to access secrets on substore (git), altough it is possible create new ones.

### Steps To Reproduce
In [1]: gopass clone https://github.com/xyz/example substore
Out [1]: Mounted password store C:\Users\user\.password-store-substore at mount point `substore`...
Configuring git repository ...
Use User (user@mail.com) for password store git config? [Y/n/q]:
Warning: Failed to commit .gitattributes to git
Your password store is ready to use! Have a look around: `gopass list substore`

In [2]: gopass ls substore
Out [2]: Error: failed to list store: failed to add file: Lstat : Cannot Find the specified path

Extra steps to overcome the above: manually change config.yml from "gpgcli-gitcli-fs+file:\Users\user\.password-store-substore" to "gpgcli-gitcli-fs+file:///C:%5CUsers%5Cuser%5C.password-store-substore"

In [3]: gopass ls substore
Out [3]: substore (C:\Users\user\.password-store-substore)
(no errors, but no secrets also)

In [4]: gopass mysecret
Out [4]: Entry 'mysecret' not found. Starting search...
Found exact match in 'substore\websites\www.website.net\mysecret'

Error: failed to retrieve secret 'substore\websites\www.website.net\mysecret': Entry is not in the password store

### Expected behavior
primarily:  "gopass myscret" should show the secret
on second place: "gopass ls substore" should show the list of secrets

### Environment
-OS: Windows 10 Pro (x64)
-OS version: 1803
- gopass: 1.8.3

### Additional context

The file of the secret exists local and remotely.
I can create new secrets, and they are synced, and all goes well but cannot read them back.

mmulqueen:
I'm having the same problem with gopass 1.8.4 and was having a similar problem with gopass 1.8.3. Platform is Ubuntu 18.04.

The way I solved it in the end was cloning as before, I had mixed success unmounting it with gopass's `gopass mounts umount` command because of the errors and found it most reliable to remove it from config.yml (on Ubuntu this is .config/gopass/config.yml).

I then re-mounted it and everything was OK. I learnt that gopass expects the following order and an absolute path:

    gopass mounts add substore /home/username/.password-store-substore-/

I also got similar results by redoing the clone with all arguments specified:

    gopass clone user@githost.example.com:/path/to/secrets.git substore --crypto gpgcli --sync gitcli --path=/home/username/.password-store-

My problems seemed to stem from:

[...]

PR #1106, opened 2019-04-30
Update list.go

jeromebagattin:
fixe issue #1005
With Windows sep is \ and break search when is use between alias and path


dominikschulz:
Hi,

I'm afraid but I still don't have any way to test this on Windows. And we've been changing these separators back and forth a few times I won't merge this unless I'm confident that it will fix an issue and not open another one.

The next step towards better Windows support would probably mean setting up some kind of test environment or someone taking (strong) ownership over the windows port.

Issue #1111, opened 2019-05-10
Windows : Failed to retrieve secret in mounted store


juliengb:
### Summary

When using windows, secrets within a folder within a mounted store cannot be retrieved correctly.

### Steps To Reproduce

* Set up a mounted store in Windows (named "pro" in this example)
* Insert passwords
```
> echo OK | gopass insert pro/leaf
> echo OK | gopass insert pro/foo/leaf
```
* Try to retrieve passwords
```
> gopass show leaf
+---------------------------------------
|gopass
|Found secrets - Please select an entry
+---------------------------------------
 pro/leaf
 pro\foo/leaf
```
 * Selecting `pro\foo/leaf` results in
```
Error: failed to retrieve secret 'pro\foo/leaf': Entry is not in the password store
```

### Expected behavior
* Entry listed as `pro/foo/leaf` instead of `pro\foo/leaf`
* Retrievable entry

### Environment
- OS: Windows 7
- gopass Version: 1.8.5
- Installation method: Download and unzip windows binaries from github
- Mount config :
```yaml
mounts:
  pro:
    askformore: false
    autoclip: true
    autoprint: false
    autoimport: true
    autosync: false
    check_recipient_hash: false
    cliptimeout: 45
    concurrency: 1
    editrecipients: false
    nocolor: false
    noconfirm: true
    nopager: false
    notifications: true
    path: gpgcli-gitcli-fs+file:///D:%5CPath%5CTo%5CPassword%5CStore
    safecontent: false
    usesymbols: false
```

### Additional context
* Secret can be retrieved when using forward slashes
```
> gopass show pro/foo/leaf
OK
```
* Secrets that are not in a mounted store can be retrieved properly
```
> echo OK | gopass insert foo/leaf
> echo OK | gopass insert foo/bar/leaf
> gopass show leaf
+---------------------------------------
|gopass
|Found secrets - Please select an entry
+---------------------------------------
 foo/leaf : OK
 foo\bar/leaf : OK
 pro/leaf : OK
 pro\foo/leaf : Error: failed to retrieve secret 'pro\foo/leaf': Entry is not in the password store
```
* The problem also affects jsonapi, rendering the browser add-on useless for me
* The listing of secrets also show slash handling problems :
```
> echo OK | gopass insert foo/bar/baz/leaf
> echo OK | gopass insert foo/bar/baz/qux/leaf
> gopass ls -f foo
foo/bar\baz\qux/leaf
foo/bar\baz/leaf
[...]

PR #1138, opened 2019-07-08
resolve issue #1111

cymruu:


siebertm:
If it somehow helps: I tried this on my win10 and it works both on commandline as well as in chrome using gopass bridge!

Thanks a lot!

Issue #1134, opened 2019-07-03
Notification count incorrect with multiple stores


pieterlexis:
### Summary
When running `gopass sync` with multiple stores, the notification shows the wrong summary of entries removed/added. The summary seems to show the net result of the sync, but does not reflect what actually happened.

The console showed me the following:
```
Sync starting ...
[<root>] 
   git pull and push ... OK (Removed 11 entries)
   importing missing keys ... OK
   exporting missing keys ... nothing to do
   done
[company-passwd] 
   git pull and push ... OK (Added 1 entries)
   importing missing keys ... OK
   exporting missing keys ... nothing to do
   done
All done
```

Whereas the notification showed me "Removed 10 entries" (couldn't get a screenshot in time, sorry), which is indeed the net result of the operation, but not an accurate representation of the result.

NB: I did not test the behaviour with one configured store.

### Steps To Reproduce
* Have 2 configured stores
* Remove N entries from store A (from another machine, or outside of the gopass install)
* Add M (where M<N) entries to store B (from another machine, or outside of the gopass install)
* run `gopass sync`
* Observe the notification shows "`N-M` entries removed"

### Expected behavior
I would expect the notification to show "Removed N entries, added M entries".

### Environment
- OS: Arch Linux
- OS version: `Linux perzik 5.1.15-arch1-1-ARCH #1 SMP PREEMPT Tue Jun 25 04:49:39 UTC 2019 x86_64 GNU/Linux`
- gopass Version:
```
gopass 1.8.5 (2019-03-13 17:51:51) go1.12 linux amd64
<root>     - gpg 2.2.16 - git 2.22.0 -   fs 0.1.0
Available Crypto Backends: gpgcli, openpgp, plain, xc
Available RCS Backends: gitcli, noop
Available Storage Backends: fs, inmem
```
- Installation method: Arch Repo

pieterlexis:
not stale, see #1140

PR #1140, opened 2019-07-10
Count changes for each store

cymruu:
A proposition to resolve issue #1134

### DIFF
--- a/pkg/action/sync.go
+++ b/pkg/action/sync.go
@@ -47,9 +47,13 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
 	// calculate number of changes entries
 	totalAdded := 0
 	totalRemoved := 0
+<<<<<<< HEAD
 	if len(changes) > 0 {
 		out.Print(ctx, "Sync summary:\n")
 	}
+=======
+	out.Print(ctx, "Sync summary:\n")
+>>>>>>> issue#1134
 	for k := range changes {
 		if changes[k] > 0 {
 			out.Print(ctx, fmt.Sprintf("Mount %s: Added %d entries", k, changes[k]))
@@ -58,7 +62,11 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
 			out.Print(ctx, fmt.Sprintf("Mount %s: Removed %d entries", k, -1*changes[k]))
 			totalRemoved += -1 * changes[k]
 		} else {
+<<<<<<< HEAD
 			out.Print(ctx, fmt.Sprintf("Mount %s: no changes", k))
+=======
+			out.Print(ctx, fmt.Sprintf("Mount %s: didn't change", k))
+>>>>>>> issue#1134
 		}
 	}
 	diff := ""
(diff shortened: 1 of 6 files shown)


pkg/action/sync.go
```diff
@@ -42,29 +37,44 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
...
+	if len(changes) > 0 {
+		out.Print(ctx, "Sync summary:\n")
+	}
+	for k := range changes {
+		if changes[k] > 0 {
+			out.Print(ctx, fmt.Sprintf("Mount %s: Added %d entries", k, changes[k]))
+			totalAdded += changes[k]
+		} else if changes[k] < 0 {
+			out.Print(ctx, fmt.Sprintf("Mount %s: Removed %d entries", k, -1*changes[k]))
+			totalRemoved += -1 * changes[k]
```
dominikschulz: Remove the ´-1 *´, just add up what's removed.

pkg/action/sync.go
```diff
@@ -42,29 +37,44 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
...
+		} else {
+			out.Print(ctx, fmt.Sprintf("Mount %s: no changes", k))
+		}
 	}
 	diff := ""
```
dominikschulz: Maybe use a strings.Builder here?

pkg/action/sync.go
```diff
@@ -42,29 +37,44 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
...
-	if numEntries > 0 {
-		diff = fmt.Sprintf(" Added %d entries", numEntries)
-	} else if numEntries < 0 {
-		diff = fmt.Sprintf(" Removed %d entries", -1*numEntries)
+	if totalAdded > 0 {
+		diff += fmt.Sprintf(" Added %d entries", totalAdded)
+	}
+	if totalRemoved > 0 {
+		diff += fmt.Sprintf("\nRemoved %d entries", totalRemoved)
```
dominikschulz: Remove the "\n", please.

pkg/action/sync.go
```diff
@@ -42,29 +37,44 @@ func (s *Action) sync(ctx context.Context, c *cli.Context, store string) error {
...
 	}
 	_ = notify.Notify(ctx, "gopass - sync", fmt.Sprintf("Finished. Synced %d remotes.%s", numMPs, diff))
 
 	return nil
 }
 
 // syncMount syncs a single mount
-func (s *Action) syncMount(ctx context.Context, mp string) error {
+func (s *Action) syncMount(ctx context.Context, mp string) (*syncResult, error) {
```
dominikschulz: Don't use a pointer to syncResult. We don't need to modify the object later.

pkg/action/sync.go
```diff
@@ -90,24 +100,25 @@ func (s *Action) syncMount(ctx context.Context, mp string) error {
...
+			return nil, err
 		}
 
 		out.Error(ctx, "Failed to push '%s' to its remote: %s", name, err)
-		return err
+		return nil, err
 	}
 	out.Print(ctxno, color.GreenString("OK"))
-
+	syncResult := &syncResult{name, 0}
```
dominikschulz: Don't explicitly init default values, i.e. drop the 0.

pkg/action/sync.go
cymruu: Isn't that convenient so I can return nil?

pkg/action/sync.go
cymruu: Since if entries were removed it returns negative number it's necessary to not print "removed -5 entries". Or I am wrong?

PR #1166, opened 2019-08-16
[WIP] Make pathsep platform depenent

martinhoefling:
Windows uses \ and os.PathSeparator should yield the correct one.

### DIFF
--- a/pkg/jsonapi/responses.go
+++ b/pkg/jsonapi/responses.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path"
 	"regexp"
 	"strings"
@@ -15,10 +16,6 @@ import (
 	"github.com/gopasspw/gopass/pkg/store/secret"
 
 	"github.com/pkg/errors"
-)
-
-var (
-	sep = "/"
 )
 
 func (api *API) respondMessage(ctx context.Context, msgBytes []byte) error {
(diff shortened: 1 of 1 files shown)


Pharb:
@martinhoefling I do not know much about handling Windows file paths in Go, but I am wondering if gopass store paths are supposed to be actual file paths or a more abstract path, independent of the file system (and also independent of the operating system)?

According [to this StackOverflow post](https://stackoverflow.com/questions/9371031/how-do-i-create-crossplatform-file-paths-in-go):

> path is intended for forward slash-separated paths only (such as those used in URLs), while path/filepath manipulates paths across different operating systems.

Maybe the use of `os.PathSeparator` together with functions like `path.Base` could cause different or similar problems?

Do we know how this is handled by gopass outside the jsonapi package?

Also the state of Windows support is still unclear, maybe because there is no maintainer using Windows regularly: https://github.com/gopasspw/gopass/blob/v1.8.6/docs/setup.md#windows-1

Lastly I think this could be related to https://github.com/gopasspw/gopass/issues/1111 and https://github.com/gopasspw/gopass/pull/1138, maybe it is even the same issue?

martinhoefling:
@Pharb good point. I think what we really would need is working tests / CI on more than linux. I'll mark this PR as WIP as I cannot asses atm if it really solves any issue or even creates another one.

AuspeXeu:
With respect to the issue in the JSON api. It's still weird that the paths seem to have mixed separators under windows. But this happens only for sup-stores.

AuspeXeu:
So, I have done some more testing and using the test-client from the gopassbridge repository, it turns out that the response it gets from gopass' JSON api already carries the wrong separators. However, this is only the case for mounted stores. Example: `storeA\\domain.com/me@example.com` is returned for an entry in `storeA` while an entry in the root store correctly returns `domain.com/me@example.com`.

PR #1171, opened 2019-08-29
Windows tests

martinhoefling:
The intention of this WIP PR is to make tests working on OSX and Windows. Maybe first with an ignored result in order to fix them and bring them to a consistent handling of path separators. The semantics as mentioned by @Pharb and discussed in PR #1166 is an out of scope topic here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,26 @@
 language: go
 dist: trusty
-os:
-  - linux
+
+matrix:
+  include:
+    - os: windows
+      env: SLOW_TEST_FACTOR=100 BUILDFLAGS='' GOPASS_OUTPUT=gopass.exe
+    - os: osx
+      env: SLOW_TEST_FACTOR=100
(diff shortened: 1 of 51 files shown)


martinhoefling:
It makes tests at least run on windows but the result is ignored. Also OSX is reenabled and some tests have a larger delay on OSX to make them pass. Maybe as a starter, we can merge it.

Pharb:
If the CI runs green, I think it would be great to have multi-platform testing.

Maybe instead of the `allow_failures` we could start with [skipping tests](https://golang.org/pkg/testing/#hdr-Skipping) that fail on Windows. At least that would avoid regressions.

One risk might be more flaky CI runs, but worst case we could always disable them again.

martinhoefling:
@dominikschulz , please re-review. I have disabled several tests for windows but at least it's a start to improve the coverage and functionality.

PR #1203, opened 2019-10-21
Add documentation for all config.yml options

### LINKED ISSUE #1052: Document all config options
gopass has several config options, but I'm not sure what all of them do.
```
> gopass config
root store config:
  askformore: false
  autoclip: false
  autoimport: true
  autosync: true
  check_recipient_hash: false
  cliptimeout: 30
  concurrency: 1
  editrecipients: false
  nocolor: false
  noconfirm: true
  nopager: false
  notifications: false
  path: gpgcli-gitcli-fs+file:///Users/tomkel/.password-store
  safecontent: true
  usesymbols: false
```

[There is some documentation on this](https://github.com/gopasspw/gopass/blob/master/docs/config.md), but it doesn't document all of the options. Please update the documentation?

rianadon:
Fixes #1052. Hopefully this brings the gopass documentation up to perfect level :sunglasses: .

While looking into these options, I noticed `recipient_hash` was always nil. I'm not sure whether this is intentional, but if it's supposed to be modified and stored then it isn't.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -42,14 +42,22 @@ All configuration options are also available for reading and writing through the
 
 This is a list of available options:
 
-| **Option**    | **Type** | Description |
-| ------------- | -------- | ----------- |
-| `askformore`  | `bool`   | If enabled - it will ask to add more data after use of `generate` command. |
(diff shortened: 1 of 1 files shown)


tomkel:
thank you!

PR #1222, opened 2019-12-01
Use present imperative tense for commits on move events

hakusaro:
I know what you're thinking: "you sent a PR for _this_? come now!"

This changes the commit messages for moving files to be present imperative tense instead of past tense. Related things that bother me are "Exported public keys" and "Added recipient" which disagree with "Save secret" and "Remove blah from store" in terms of tense. If we're good with using present imperative tense everywhere, I will gladly send patches that change the convention everywhere.


vngzs:
Good, this is in alignment with [best practices on commit messages](https://chris.beams.io/posts/git-commit/#imperative).

PR #1231, opened 2019-12-19
Add env command

### LINKED ISSUE #1227: [feat] gopass env command (export keys as environment variables)
Hello,

I would like to propose a `gopass env` command, in the same spirit of the [envchain](https://github.com/sorah/envchain) or [aws-vault](https://github.com/99designs/aws-vault) utilities.

For example, assuming a password store:

```
scaleway
    ├── SCW_ACCESS_KEY
    └── SCW_SECRET_KEY
```

Running
```
gopass env scaleway -- terraform apply
```
would execute the command `terraform apply` making available as environment variables they keys found below the `scaleway` directory.

If the argument to `env` is not a directory but a key, then gopass would export only that key.
If the argument is a directory that contains subdirectories, then the invocation would fail.

I think that this feature would greatly simplify the usage of gopass for certain cases, such as the terraform example or any example where the utility to be invoked supports environment variables.

For example, the example above today requires something like

```
env SCW_ACCESS_KEY=$(gopass scaleway/access_key) \
    SCW_SECRET_KEY=$(gopass scaleway/secret_key) \
    terraform apply
```

or something like

```
terraform apply \
  -var="access_key=$(gopass scaleway/access_key)" \
  -var="secret_key=$(gopass scaleway/secret_key)"
```

If this feature is considered useful, I would like to propose a PR.

Thanks!

dominikschulz:
Fixes #1227 

### DIFF
--- a/commands_test.go
+++ b/commands_test.go
@@ -34,6 +34,7 @@ var commandsWithError = map[string]struct{}{
 	".create":            {},
 	".delete":            {},
 	".edit":              {},
+	".env":               {},
 	".find":              {},
 	".generate":          {},
 	".git.push":          {},
@@ -90,7 +91,7 @@ func TestGetCommands(t *testing.T) {
 	c.Context = ctx
 
 	commands := getCommands(ctx, act, app)
-	assert.Equal(t, 32, len(commands))
+	assert.Equal(t, 33, len(commands))
 
 	prefix := ""
 	testCommands(t, c, commands, prefix)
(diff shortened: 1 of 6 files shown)


marco-m:
Hello @dominikschulz, I am the proposer of #1227, thanks for picking this up :-)
I looked at the code, first thing that I see missing are the tests :-), is this because of the `WIP` tag? (I am not familiar with the `gopass` project conventions).
Thanks!

Issue #1214, opened 2019-11-24
Cannot install gopass with go get


acheron2302:
### Summary
I don't know why but when i run `go get github.com/gopasspw/gopass`
It return this error and then stop install:
```
# github.com/gopasspw/gopass/pkg/completion/zsh
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:25:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.BoolFlag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:27:8: undefined: cli.BoolTFlag
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:29:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.Float64Flag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:31:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.GenericFlag (IsSet method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:33:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.Int64Flag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:35:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.Int64SliceFlag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:37:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.IntFlag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:39:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.IntSliceFlag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:41:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.StringFlag (Apply method has pointer receiver)
go/src/github.com/gopasspw/gopass/pkg/completion/zsh/completion.go:43:3: impossible type switch case: f (type cli.Flag) cannot have dynamic type cli.StringSliceFlag (Apply method has pointer receiver)
[...]

Issue #1199, opened 2019-10-16
undefined: ghrel.FetchLatestRelease


DamianSkrzypczak:
### Summary
Error `undefined: ghrel.FetchLatestRelease` appears during `go get` installation.

### Steps To Reproduce
Attempt to install `gopass` using `go get` with go version 1.13.1
in project outside `$GOPATH` which have `go.mod` 

### Expected behavior
Installations successes

### Environment
go version 1.13.1
GO111MODULE=on
Installation happens for project that uses `go.mod`

- OS: Manjaro
- OS version: 4.19.78-1-MANJARO
- gopass Version: 1.8.6
- Installation method: `go get` (`go get -u` checked as well)

### Additional context
After brief check I see that `master` branch from github.com/dominikschulz/github-releases/ghrel contains `FetchLatestRelease` but tag `v0.0.1` doesn't. Release `github-releases` `v0.0.2` should fix it. @dominikschulz

DamianSkrzypczak:
Console output:
```bash
> go get github.com/gopasspw/gopass
go: finding github.com/xrash/smetrics latest
go: finding golang.org/x/crypto latest
go: finding github.com/martinhoefling/goxkcdpwgen latest
go: finding github.com/muesli/crunchy latest
go: finding github.com/alecthomas/binary latest
go: finding golang.org/x/sys latest
go: finding github.com/kevinburke/ssh_config latest
go: finding github.com/jbenet/go-context latest
go: finding golang.org/x/net latest
go: finding github.com/jsimonetti/pwscheme latest
go: finding github.com/kballard/go-shellquote latest
go: finding github.com/armon/go-metrics latest
go: finding github.com/nsf/termbox-go latest
go: finding golang.org/x/time latest
go: finding github.com/mitchellh/go-ps latest
go: finding github.com/muesli/goprogressbar latest
go: finding github.com/skip2/go-qrcode latest
# github.com/gopasspw/gopass/pkg/updater
go/pkg/mod/github.com/gopasspw/gopass@v1.8.6/pkg/updater/update.go:94:10: undefined: ghrel.FetchLatestRelease
```

DamianSkrzypczak:
Is there is any progress on this issue done so far?

acheron2302:
It seem like changing the the go.mod `github.com/dominikschulz/github-releases v0.0.1` to `github.com/dominikschulz/github-releases v0.0.0-20171016132906-7635b43b2447` fix the problem

[...]

PR #1242, opened 2020-01-23
Fix go get and update to v1.8.7

acheron2302:
- First i change the github.com/urfave/cli to gopkg.in/urfave/cli.v1, because if you download with go get outside $GOPATH, then it will use the latest version which is the version 2. But in v2 they change all of the type so it break when you try to install with go get outside $GOPATH

- Second we need to update to version 1.8.7 with git tag v1.8.7, so that when they use go get inside the $GOPATH it will clone from git tag and our v1.8.6 tag is commit 6 month ago without the new go 1.11 module. So to fix that problem we need to update the to new tag v1.8.7

Fix issue:
#1214 #1199


Pharb:
Thanks a lot for the effort @acheron2302, it is greatly appreciated!

If I look at the documentation here, it mentions that the default should still be v1: https://github.com/urfave/cli#using-v1-releases

>  if you download with go get outside $GOPATH, then it will use the latest version which is the version 2

What is the reason to "go get" without the $GOPATH set?

According to the examples there, it should not be necessary to use gopkg.in anymore: https://blog.golang.org/v2-go-modules

Maybe I don't fully understand the issue, but what is the exact reason to switch the import to gopkg.in?

acheron2302:
@Pharb In https://blog.golang.org/v2-go-modules, it said: 

> Before Go modules, gopkg.in allowed package maintainers to follow what we now refer to as the import compatibility rule

In ơgo module doc](https://github.com/golang/go/wiki/Modules#why-does-installing-a-tool-via-go-get-fail-with-error-cannot-find-main-module):
>Leave GO111MODULE unset (the default, or explicitly set GO111MODULE=auto), which results in friendlier behavior. This will give you Go 1.10 behavior when you are outside of a module and hence will avoid go get reporting cannot find main module.

So that mean if you are outside a module and then it will return to 1.10 behavior which mean downloading the latest version into src. So that mean it will download the v2 from urfave and everything go kaboom. So we need to change it to gopkg.in and it will fix the problem

The another solution i haven't mention is that: In the README.md you guys should change the download to: `GO111MODULE=on go get -u`. But the problem it's that with it it will download v1.8.6 from this github and 1.8.6 release is 7 month ago and it still use gopkg.lock which is very outdate. So you guys need to update the release. And the reason i change it to gopkg.in is because in case somebody run `go get -u ` outside then there won't be any error. But i think by update the release to v1.8.7 and change the way to download to `GO111MODULE=on go get -u` will work just fine without the need of gopkg.in

muesli:
I think this may potentially break usage of Go modules, though, as the canonical path of urfave/cli changed back to github.com/urfave/cli. I think we still build within GOPATH (and without using Go modules) on TravisCI. Maybe it's a good time to change that now, so we also verify that the package versions specified in `go.mod` actually work with our builds?

VERSION
```diff
@@ -1 +1 @@
-1.8.6
+1.8.7
```
dominikschulz, line 1: This is incremented on release only.
Our latest release is 1.8.6, so this should stay as is.

dominikschulz:
Obsoleted by #1273.

PR #1267, opened 2020-04-23
We must not parallelize the audit operation

muesli:
Sadly this breaks gnupg in various ways by launching multiple processes of it.
See the entire discussion in the original pull request by @metalmatze:

https://github.com/gopasspw/gopass/pull/245

### DIFF
--- a/pkg/audit/audit.go
+++ b/pkg/audit/audit.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"runtime"
 
 	"github.com/gopasspw/gopass/pkg/notify"
 	"github.com/gopasspw/gopass/pkg/out"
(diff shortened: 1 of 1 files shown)


pkg/audit/audit.go
```diff
@@ -47,7 +46,14 @@ func Batch(ctx context.Context, secrets []string, secStore secretGetter) error {
 
 	// Spawn workers that run the auditing of all secrets concurrently.
 	validator := crunchy.NewValidator()
-	maxJobs := runtime.NumCPU()
+
+	// It would be nice to parallelize this operation and limit the maxJobs to
+	// runtime.NumCPU(), but sadly this causes various problems with multiple
+	// gnupg jobs running parallely. See the entire discussion here:
```
martinhoefling, line 52: ./pkg/audit/audit.go:52:23: "parallely" is a misspelling of "parallelly" :-)

metalmatze:
Ok. :blush: :+1:

pkg/audit/audit.go
muesli: Thank you! I can not count how many times in my life I have typoed this word :smile:

muesli:
@martinhoefling @metalmatze Done and done.

muesli:
Feel free to rebase/merge this PR whenever it suits you. I don't have the permissions to do so myself :wink:

dominikschulz:
This depends on the GnuPG version you are using. It used to work fine on my ancient Ubuntu 16.04 (gnupg 1.4).

On 19.10 (gnupg 2.2) the useability is not that nice. So I think this is OK as a mitigation until we properly fix it.

muesli:
@dominikschulz Interesting, I guess that's possible even though I would expect it to fail in the same way when accessing smartcards. I'm a bit shocked Ubuntu 16.04 still shipped with GnuPG v1. I don't think I've seen anything below v2 in more than a decade :smile:

dominikschulz:
See https://packages.ubuntu.com/xenial/gnupg: gnupg 1.4.20

But that's not my point. I'm currently contemplating if we need encryption at all (don't worry, the answer is most likley yes) and if so, if we want to support bulk operations at all (since they could be misused to extract credentials from the store).

So anything that involves bulk decrypt operations is likely slated for removal anyway.

Issue #525, opened 2017-12-15
Support for reencrypt


jerger:
as longterm password-store users we've build around some make extensions over plain password store for handling keys, bash-autocompletion and multi-stores (https://github.com/DomainDrivenArchitecture/password-store-for-teams). Two things I miss after reading your documentation:

* do you support reencryption (e.g. if new members join a team)?
* I found the key-import but do you offer a key-publish also?

BR - gopass is really cool :-)

dominikschulz:
- Running `gopass recipients add` or `gopass recipients remove` will automatically reencrypt the whole store
- Adding a recipient will automatically export his key to `.gpg-keys/<ID>`

jerger:
cool - thanx :-)

nylocx:
I hope it is ok to bring such an old issue up again.
I have another use case that requires a command to reencrypt all or a specific secret in a store.
I use the wonderful iOS app [pass for ios](https://github.com/mssun/passforios) to use pass on my iPhone. It currently does not support encryption with multiple public keys (but it is being worked on in https://github.com/mssun/passforios/issues/338). Anyways there might be other clients facing the same problem. So I would like to have a way to fix the secrets encrypted with just a single key. Currently I have to remove a recipient and then readd the same recipient and this way doing the reencryption twice (once for removal and ones for adding).
It would be nice to just have a `gopass reencrypt --store some_store` command. I guess I could come up with some script doing this directly via gpg encrypt/decrypt, but providing the command should not be very hard, as reencryption is already implemented.

PS: Thanks for the great work, I really love gopass.

martinhoefling:
I reopened it for now but it needs sb who has time to implement it.

martinhoefling:
Stupid question regarding your usecase: does `gopass fsck` add your missing recipients? I've tested it here and it seems to work. Maybe that should be documented somewhere.

nylocx:
I just checked and you are right, I totally missed that functionality. The only thing that is strange is if I do: `gopass fsck --store private` I get an error saying:
`Incorrect Usage: flag provided but not defined: -store`

martinhoefling:
Ok, I am keeping this open for improving the documentation.

Paktosan:
I hope it's okay to dump this in this issue and not open a new one.
I need to reencrypt some files because I missed some updated keys.
While running `gopass fsck --decrypt` fixes that, it also changes every entry for file format and creates a commit for each.
In my case, this results in over 350 commits.
[...]

PR #1269, opened 2020-04-23
add sidenote how secrets with missing recipients can be fixed / reencrypted

martinhoefling:
Clarifies issue raised in #525


muesli:
LGTM.

Issue #1298, opened 2020-04-28
Simplify releases


dominikschulz:
We should be able to simplify releases to a short, one step process.

* Require the use a RELEASE_NOTES tag in every merge commit
* Create a shell script the requires GITHUB_TOKEN and possibly VERSION input.
* Parse every commit since the last release extracting RELEASE_NOTES, adding that to our CHANGELOG
* Run all tests and a goreleaser dry run
* Publish the release, reading the GitHub release notes from our changelog

This is blocked by #1296

dominikschulz:
This was closed in error.

I'm still looking for a way to enforce changelog entries to be part of PR descriptions / merge commits.
If we find a way to to that, I could probably script something to do the remaining steps (although I'm happy if someone wants to pick this up).

dominikschulz:
Added the CommitCheck integration to check for RELEASE_NOTES entry in the PR description. Will test if this work the way I expect.

With this me might actually be able to do fully scripted releases.

PR #1302, opened 2020-04-29
[WIP] update config to build with latest goreleaser

martinhoefling:
This PR is WIP in order to simplify releases, see #1298.

Current state: `goreleaser --snapshot --skip-publish --rm-dist` runs fine.

Unfortunately arch linux packages (packman) cannot be built anymore. Any ideas how to proceed or if they are used at all? @metalmatze created the AUR package as far as I read.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -2,107 +2,102 @@
 # Release automation
 #
 # Build customization
-build:
-  binary: gopass
-  flags: |
-    -tags netgo -gcflags="-trimpath=$GOPATH" -asmflags="-trimpath=$GOPATH"
-  env:
-    - CGO_ENABLED=0
-  ldflags: |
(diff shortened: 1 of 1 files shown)


dominikschulz:
Thanks a lot!

I would not worry about Arch, there is an official package, so we can point Arch Linux users there: https://www.archlinux.org/packages/community/x86_64/gopass/

martinhoefling:
OK, let's merge this for now. I will go through it and see if we can make some additions (e.g. bintray). We can do the next release with the new version and adapt the documentation.

dominikschulz:
I'd rather not invest too much time into bintray.
My experience with that has been pretty bad so far and we haven't had up to date packages there for a long time.

I'd rather not provide a package repository then a completely outdated one.

PR #1305, opened 2020-04-29
Create new non-root stores in XDG compliant locations

### LINKED ISSUE #1225: Do not store user files in root of home
### Summary

When starting gopass, it tries to create a .password-store file directly at the root of $HOME
### Steps To Reproduce

Start gopass

### Expected behavior

FIle .password-store should be stored at $XDG_CONFIG_HOME, if it does not exist, it should default to $HOME/.config/gopass on unix and %appdata%/gopass on Windows.
### Environment

Windows 10 1809
Linux Mint 19
gopass 1.8.6
Installation from github packages/binaries

### Additional context

While far away from critical, let's try to not clutter users' root directory, you cannot predict what will be displayed or not to users when they browse their files.

Please take a look at https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

dominikschulz:
This partly addresses #1225 in that it create new substores in XDG compliant
locations.

RELEASE_NOTES=[ENHANCEMENT] Create new sub stores in XDG compliant locations.

### DIFF
--- a/pkg/config/location.go
+++ b/pkg/config/location.go
@@ -37,11 +37,7 @@ func configLocation() string {
 	// Second, check for the "XDG_CONFIG_HOME" environment variable
 	// (which is part of the XDG Base Directory Specification for Linux and
 	// other Unix-like operating sytstems)
-	if xch := os.Getenv("XDG_CONFIG_HOME"); xch != "" {
-		return filepath.Join(xch, "gopass", "config.yml")
-	}
-
-	return filepath.Join(Homedir(), ".config", "gopass", "config.yml")
+	return filepath.Join(UserConfig(), "config.yml")
 }
 
 // configLocations returns the possible locations of gopass config files,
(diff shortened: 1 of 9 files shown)


dominikschulz:
WIP'd because I don't know if we should change this pre-2.0.

pkg/config/location.go
```diff
@@ -64,11 +64,18 @@ func configLocations() []string {
...
 func PwStoreDir(mount string) string {
 	if mount != "" {
-		return fsutil.CleanPath(filepath.Join(Homedir(), ".password-store-"+strings.Replace(mount, string(filepath.Separator), "-", -1)))
+		cleanName := strings.Replace(mount, string(filepath.Separator), "-", -1)
+		// comply with the freedesktop XDG spec
+		// https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+		if xdh := os.Getenv("XDG_DATA_HOME"); xdh != "" {
+			return fsutil.CleanPath(filepath.Join(xdh, "gopass", "stores", cleanName))
+		}
+		return fsutil.CleanPath(filepath.Join(Homedir(), ".local", "share", "gopass", "stores", cleanName))
```
Pharb, line 73: For better cross-platform support, maybe we can use a library to determine the correct location: https://github.com/ProtonMail/go-appdir

This should also make most uses of [mitchellh/go-homedir](https://github.com/mitchellh/go-homedir) obsolete here?

dominikschulz: Yes, that is a good idea. The code gets cleaner that way.
Unfortunately we can't get rid completey of `mitchellh/go-homedir`, yet.

PR #1360, opened 2020-05-13
Clarify fish completion setup

sergiuszkierat:
This is a clarification how to use auto completion in fish

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -281,7 +281,11 @@ Exit and re-run zsh if the last command failed.
 
 ### Enable fish completion
 
-If you use the [fish](https://fishshell.com/) shell, you can enable experimental shell completion. Copy the file `fish.completion` to `~/.config/fish/completions/gopass.fish` and start a new shell.
+If you use the [fish](https://fishshell.com/) shell, you can enable experimental shell completion by the following command:
+```fish
+$ mkdir -p ~/.config/fish/completions and; gopass completion fish > ~/.config/fish/completions/gopass.fish
+```
+and start a new shell afterwards.
 
 Since writing fish completion scripts is not yet supported by the CLI library we use, this completion script is missing a few features. Feel free to contribute if you want to improve it.
 


sergiuszkierat:
Due to the exception
```
Something went wrong ...
You may have been logged out.
```
at commitcheck, I am not able to see [commitcheck](https://www.commitcheck.com/status/gopasspw/gopass/f469909e-3197-4aba-b012-5ae2fa88810d) message. What is wrong? :thinking:

tomkadwill:
@sergiuszkierat thanks for raising this! looks like there is an authorization error in the API call I'm making to GitHub to check whether you have read permission for this repo. I'll contact GitHub about that.

In the meantime, I should be able to ignore the auth check for public repos like this one, making the statuses viewable to anyone. I'll let you know when that's done, shouldn't take longer than a day or two.

sergiuszkierat:
> @sergiuszkierat thanks for raising this!
> ...

@tomkadwill thanks for fast reaction, I am waiting for it :hourglass_flowing_sand:

dominikschulz:
Thanks!

The commit check complains about a missing RELEASE_NOTES tag in your commit: https://github.com/gopasspw/gopass/blob/master/CONTRIBUTING.md#pull-request-checklist

But for this one this isn't necessary.

tomkadwill:
@sergiuszkierat FYI I've fixed the status page for [CommitCheck](https://commitcheck.com) so you should be able to view status pages from now on

sergiuszkierat:
@tomkadwill, confirmed. I am able to view the CommitCheck's link which I mentioned above.

Btw, I don't see any details why it fails as @dominikschulz wrote [here](https://github.com/gopasspw/gopass/pull/1360#issuecomment-628860821) :thinking: Is it an expected outcome?

tomkadwill:
Thanks for confirming 👍

Yeah that's a good point. The status page feature is relatively new and the initial idea was to show users which commit(s) failed. This is useful if you have multiple commits in a PR and only some fail the regex check. Eg: https://twitter.com/TomKadwill/status/1218280859465342976/photo/1

I'll add a ticket to the backlog to include details about why it fails, as @dominikschulz wrote. I think it's a great suggestion.

Issue #1310, opened 2020-05-01
Avoid heuristics for secret content


dominikschulz:
Currently we apply multiple heuristics to the content of a secret, leading to confusing or hard to fix behaviour, e.g. #1235, #1121 .

We should come up with a clear, consistent story for the content of secrets.

One option would be to introduce a content type header (e.g. [MIME](https://tools.ietf.org/html/rfc2045)) and threat every secret without a content type as plain text.

This would allow to simplify both binary as well as YAML and K/V type secrets.

dominikschulz:
If possible this should land before a 2.x release.

PR #1385, opened 2020-05-23
Add initial draft of the gopass API

### LINKED ISSUE #1379: [API] Define and implement gopass API
This is a follow up to #849.

After we have unexported most packages gopass can no longer be embedded / used as an API.

Going forward we'd like to reduce the number of features or integrations included but at the same time we'd like to encourage embedding gopass through a clean API.

We'll need to define such and API and implement it, possibly with an example API user (e.g. HIBP auditing).

dominikschulz:
This commit contains the initial draft of the gopass API.
Design and implementation of the API are work in progress - even after
this PR has been merged. Feedback welcome but please use with caution.
Expect breaking changes to the API. Especially once we redesign the
secrets implementation.

As a proof of concept it also moves the following subcommands out of the main binary:

- jsonapi
- hibp
- git-credentials

This is done to validate the API design and show how the API should be used.
These additional binaries might want to move to their own repositories once the API
is finalized.

At least the secrets part of the API will need to change. However this is part of a
larger effort that does not fit this PR.

Fixes #1379

RELEASE_NOTES=[ENHANCEMENT] Add gopass API (unstable)


cmd/gopass-hibp/hibp_test.go
```diff
@@ -157,11 +154,52 @@ func TestHIBPAPI(t *testing.T) {
...
 
 	// test with one entry
-	assert.NoError(t, act.HIBP(c))
+	assert.NoError(t, act.api(c.Context, false))
 	buf.Reset()
 
 	// add another one
-	assert.NoError(t, act.insertStdin(ctx, "baz", []byte("foobar"), false))
-	assert.Error(t, act.HIBP(c))
+	// TODO - assert.NoError(t, act.gp.Set(ctx, "baz", []byte("foobar"), false))
```
martinhoefling, line 161: Reminder: TODO

cmd/gopass-hibp/hibp_test.go
```diff
@@ -114,6 +110,8 @@ func testWriteGZ(fn string, buf []byte) error {
 }
 
 func TestHIBPAPI(t *testing.T) {
+	t.Skip("TODO")
```
martinhoefling, line 113: Reminder: TODO

martinhoefling: ... nevermind, we probably just wanna check at the end for all TODOs.

cmd/gopass-jsonapi/main.go
```diff
@@ -0,0 +1,111 @@
...
+	app.EnableBashCompletion = true
+	app.Action = func(c *cli.Context) error {
+		if strings.HasSuffix(os.Args[0], "native_host") || strings.HasSuffix(os.Args[0], "native_host.exe") {
+			return ja.listen(c)
+		}
+		return cli.ShowAppHelp(c)
+	}
+	app.Commands = []*cli.Command{
+		{
+			Name:        "listen",
```
martinhoefling, line 59: If we have the ability to split up this into two binaries, I would propose to have a separate binary for the api listener. That would also make the windows naming hack obsolete (which has been removed in this PR already) and prevent accidental calling of jsonapi listen.

commands_test.go
```diff
@@ -91,7 +91,7 @@ func TestGetCommands(t *testing.T) {
 	c.Context = ctx
 
 	commands := getCommands(ctx, act, app)
-	assert.Equal(t, 34, len(commands))
+	assert.Equal(t, 32, len(commands))
```
martinhoefling, line 94: Hooray! :-)

cmd/gopass-jsonapi/main.go
dominikschulz: I don't mind splitting up the listener further.
No need to do this as part of this PR, but once we merge this you can go ahead and do that.

commands_test.go
dominikschulz: > Where to place YAML / KV support of the secrets. Is that provided by the API?

Yes, this will be part of the API.
However that part is not fully fleshed out, yet.
So far this is still exposing the old secrets interface.

After this PR has been merged I'd like to work on #1310 which will most likely overhaul the secrets implementation and change the corresponding API.

> As long as we refactor out CLI commands in the same repo, I things we're in a good shape of experimenting with the API without breaking it. Ideally, we could also try to do the same with gopass main command.

No, I don't plan to have the main gopass binary use the API ever. At least not before a 2.0 release. Maybe after that, but probably not. The reason is simple: the main binary needs to access a lot of internals which are going to change a lot. This would make the API ugly, brittle and overly complicated. Even after we introduce an API gopass will still primarily be an CLI programm. The API is just a very simple and intentionally limited extension point for features we don't want to merge in the main binary.

We can certainly revisit this discussion after we have a stable 2.x release with a trimmed down feature set, but right now this would be a lot of extra work for little or no benefit.

cmd/gopass-jsonapi/main.go
martinhoefling: Sure, let's keep it as it is - whatever is simpler.

PR #1415, opened 2020-06-07
Introduce typed MIME Secrets

### LINKED ISSUE #1310: Avoid heuristics for secret content

dominikschulz:
This commit introduces a new MIME-based secrets format that will
eventually replace any existing secret format.

Fixes #1310

RELEASE_NOTES=[BREAKING] New secrets format.


vincentbernat:
I am a bit worried about this change. Does this mean gopass won't be compatible with pass anymore? Notably, I am using browserpass and from my understanding, this change makes it compatible with it.

dominikschulz:
Reading and writing those is still supported.
New credentials will be created using the new format.

For a browser integration we recommend using gopasspw/gopassbridge.

tmccombs:
will gopassbridge work with the new format?

> New credentials will be created using the new format.

So, that means newly generated credentials won't be compatible with pass?

tmccombs:
Also, is there any documentation of how the new system will work?

dominikschulz:
Yes, new credentials will not be compatible with pass.
The documentation is indeed still a bit lacking. We'll work on that for the next release. For the time being consider not building from source unless you want to help the development (the README has been updated to that extent).

To be clear: These files can still be decrypted and edited just fine with any other password store implementation, but the assumption that the password in on the first line will be violated.

walteweiss:
So after this feature upgrade we will be able to decrypt the passwords with pass, but since the password may be not just as the very first line, then it may copy something different, right? Or is there something completely different? Will existing mobile clients (for iOS and Android) be able to work with that?

dominikschulz:
The first line of the new format is a fixed identifier string. Should be pretty obvious.
I don't really know much about existing mobile clients since I don't have use case for those myself, but I they might not work as expected.

PR #1420, opened 2020-06-15
.goreleaser.yml: Build for arm and arm64 as well

mrueg:
This should add release binaries for arm v6/v7 and arm64 as well to github so users can install gopass on  arm (e.g. raspberry pi) as well from published releases.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -26,12 +26,14 @@ builds:
       - windows
     goarch:
       - amd64
+      - arm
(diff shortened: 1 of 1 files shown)


dominikschulz:
Please add a release note to your commit, like: RELEASE_NOTES=[ENHANCEMENT] Add arm and arm64 binaries

mrueg:
Done, thanks I forgot about that!

Issue #1436, opened 2020-06-28
summon integration does not work for secrets with additional information atttached


ckolumbus:
### Summary
using gopass (1.9.2) as as summon provider (via linking the gopass executable into the summon provider directory) gopass outputs the whole secret content when queried instead of only the secret, as required by the protocol

### Steps To Reproduce

> gopass init
> gopass new
```
  Creating generic secret ...
  [1] Name                          []: test
  [2] Generate password?            [Y/n/q]: 
    [a] Human-pronounceable passphrase? (see https://xkcd.com/936/) [y/N/q]: 
    [b] How long?                   [24]: 
    [c] Include symbols?            [y/N/q]: 
    [d] Strict rules?               [y/N/q]: 
  [3] Enter zero or more key value pairs for this secret:
    [a] Name (enter to quit)        []: foo
    [b] Value for Key 'foo'         []: bar
    [a] Name (enter to quit)        []: 
```
> summon -p gopass --yaml 'X : !var misc/test' bash -c 'echo $X'

```
GxAicJQrGAA534Lo5wi5eByZ foo: bar
```

### Expected behavior
expected output: one line with the secret

```
GxAicJQrGAA534Lo5wi5eByZ
```

### Environment

- OS: ManjaroLinux 
- OS version: Linux xxx 5.4.44-1-MANJARO # 1 SMP PREEMPT Wed Jun 3 14:48:07 UTC 2020 x86_64 GNU/Linux
- gopass Version: 
```
gopass 1.9.2 (2020-05-13 21:14:02) go1.14.2 linux amd64
<root>     - gpg 2.2.20 - git 2.27.0 -   fs 0.1.0
Available Crypto Backends: age, gpgcli, plain, xc
Available RCS Backends: gitcli, noop
Available Storage Backends: fs, inmem
```
- Installation method: manjore package manager from community repo (`pacman -S gopass`)
- used summon version: cyberark-summon  0.6.7-1  (via AUR package)

ckolumbus:
As addition: i can provide a PR but wanted to discuss the proper/expected behavior before implementing anything.

Options from the top of my head:
* output secret only when called with no options and just a secret path as parameter: i don't no what other functionilty would break here but this would be my preferred solution
* change behavior when called with a different name (e.g link as "summon-gopass" and trigger different behavior is exe name is detected)
* implement a new cmd line program similar to git-credential-helper

other ideas? Feedback?

dominikschulz:
[...]

PR #1438, opened 2020-06-28
implement dedicated summon provider (#1436)

ckolumbus:
initial implementation of a dedicated summon-provider implementation for gopass as the default output behavior of `gopass` violates the `summon` contract by adding all meta data.
This implementation outputs only the secret, no other special handling is implemented.

### DIFF
--- a/cmd/gopass-summon-provider/summon-provider.go
+++ b/cmd/gopass-summon-provider/summon-provider.go
@@ -1,15 +1,11 @@
 package main
 
 import (
-	"io"
-	"fmt"
-
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/gopass"
 	"github.com/urfave/cli/v2"
 )
-
 
 type gc struct {
 	gp gopass.Store
@@ -24,11 +20,6 @@ func (s *gc) Get(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	password := secret.Get("password")
-	i, err := io.WriteString(out.Stdout, password+"\n")
-	_ = i
-	if err != nil {
-		return fmt.Errorf("could not write to stdout: %s", err)
-	}
+	out.Print(ctx, secret.Get("password"))
 	return nil
 }
--- a/cmd/gopass-summon-provider/summon-provider.go
+++ b/cmd/gopass-summon-provider/summon-provider.go
@@ -1,0 +1,34 @@
+package main
+
+import (
+	"io"
+	"fmt"
+
+	"github.com/gopasspw/gopass/internal/out"
+	"github.com/gopasspw/gopass/pkg/ctxutil"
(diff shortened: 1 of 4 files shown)


cmd/gopass-summon-provider/summon-provider.go
```diff
@@ -0,0 +1,34 @@
...
+	ctx := ctxutil.WithGlobalFlags(c)
+	ctx = ctxutil.WithNoNetwork(ctx, true)
+	path := c.Args().Get(0)
+	secret, err := s.gp.Get(ctx, path, "latest")
+	if err != nil {
+		return err
+	}
+	password := secret.Get("password")
+	i, err := io.WriteString(out.Stdout, password+"\n")
+	_ = i
```
dominikschulz, line 29: Why not just `out.Print(ctx, secret.Get("password"))`?

cmd/gopass-summon-provider/summon-provider_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	stdout := &bytes.Buffer{}
+	out.Stdout = stdout
+	color.NoColor = true
+	defer func() {
+		out.Stdout = os.Stdout
+		termio.Stdin = os.Stdin
+	}()
+
+	c := gptest.CliCtx(ctx, t)
```
dominikschulz, line 37: Something like this should work a be sufficient for this test:

```
assert.NoError(t, act.Get(gptest.CliCtx(ctx,t, "foo")))
assert.Equal(t, "bar", buf.String())
```

cmd/gopass-summon-provider/summon-provider.go
dominikschulz: `out.Print` is like `fmt.Println` but uses `out.Stdout` which is easy to capture in tests.

ckolumbus:
Test implemented, refactoring of outputting done. How can i fix the CommitCheck errors?

cmd/gopass-summon-provider/summon-provider_test.go
ckolumbus: refactored

dominikschulz:
CommitCheck wants a RELEASE_NOTES=... tag in all commits. This will be used to assemble the release notes.

ckolumbus:
> CommitCheck wants a RELEASE_NOTES=... tag in all commits. This will be used to assemble the release notes.

@dominikschulz could you maybe provide a link to some examples? What are the best practices for this in case of several (unimportant) check-ins? Squashing?

dominikschulz:
Yes, we usually squash commits in a PR.

ckolumbus:
@dominikschulz sorry for bothering you again:

> CommitCheck wants a RELEASE_NOTES=... tag

so I squash everything and add one RELEASE_NOTES tag?

dominikschulz:
Yes, that's how I'd it.
I'd prefer to only require that tag for the merge commit but couldn't figure out how to do that reliably.

Issue #1460, opened 2020-07-20
The list cmd handles shadowing incorrectly


AnomalRoil:
### Summary
When mounting a substore on a path that already exists in the password store, the philosophy was to behave like unix's `mount`, which means all gopass commands should behave like if the previously existing path did not exist anymore until the mount is unmounted.

Currently the `edit`, `insert`, `generate`, `remove` and `show` command seem to work as they should.
But the `list` command is still displaying the initial entry instead of the mounts' content. 

### Steps To Reproduce
```
gopass init
gopass generate foo/bar 20
gopass generate foo/baz 20
gopass generate bar/baz 20
gopass mounts add foo
gopass generate foo/bar 20
gopass show foo/baz
gopass ls
```

And then see the problems:
 - `gopass ls` is not displaying the mount as it should (in blue with the path in parentheses)
 - `gopass show foo/baz` is triggering search and is _finding_ `foo/baz`, whereas it shouldn't since it should be shadowed by the mount. But then it is correctly failing to decrypt it, saying `Error: failed to retrieve secret 'foo/baz': Entry is not in the password store`
 - `gopass insert/edit/rm/generate` seem to work as they should on the mount only.  

### Expected behaviour
```
> gopass show foo/baz
Entry 'foo/baz' not found. Starting search...
Error: no results found
```
or maybe a message about having found it but it being shadowed by a mount, instead of the current :
```
> gopass show foo/baz
Entry 'foo/baz' not found. Starting search...
Found exact match in 'foo/baz'

Error: failed to retrieve secret 'foo/baz': Entry is not in the password store
```

And also having:
```
> gopass ls
gopass
├── bar
│   └── baz
└── foo (/path/to/the/mount)
    └── bar
``` 
instead of 
```
gopass
├── bar
│   └── baz
└── foo
    ├── bar
    └── baz
```

### Environment
- gopass Version: gopass 1.9.3-rc.0
- Installation method: from source

### Additional context
I am going to add a new test in a PR very soon. 
Notice that this test is currently failing and as such skipped in the integration test.
But it allows to detect this problem for `gopass ls`.

AnomalRoil:
I used the new test to git bisect and find the commit that introduced the error:

```
668dd0a486bf2df92231b0e03734240b8d0d0e57 is the first bad commit
commit 668dd0a486bf2df92231b0e03734240b8d0d0e57
Author: Dominik Schulz
Date:   Fri May 29 21:59:35 2020 +0200

[...]

PR #1461, opened 2020-07-20
Handling 2 regressions from #1391

### LINKED ISSUE #1458: Panic upon config call with substores.
### Summary

When running `gopass config` on a password-store with a config file coming from a previous version and a substore, here is what happened to me:
```
gopass config
root store config:
  autoclip: true
  autoimport: true
  cliptimeout: 45
  confirm: false
  editrecipients: false
  exportkeys: false
  nocolor: false
  nopager: false
  notifications: true
  path: /home/anomalroil/password-store
  safecontent: true
mount 'test' => '/home/anomalroil/test'
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x48 pc=0x561459926f24]

goroutine 1 [running]:
github.com/gopasspw/gopass/internal/store/leaf.(*Store).Storage(...)
	/home/anomalroil/go/src/github.com/gopasspw/gopass/internal/store/leaf/store.go:163
github.com/gopasspw/gopass/internal/action.(*Action).getOnDiskStorage(0xc000205340, 0x561459ec4300, 0xc000343290, 0x0, 0x0, 0x2, 0x2)
	/home/anomalroil/go/src/github.com/gopasspw/gopass/internal/action/config.go:116 +0x134
github.com/gopasspw/gopass/internal/action.(*Action).printConfigValues(0xc000205340, 0x561459ec4300, 0xc000343290, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/anomalroil/go/src/github.com/gopasspw/gopass/internal/action/config.go:60 +0x18a
github.com/gopasspw/gopass/internal/action.(*Action).Config(0xc000205340, 0xc000366b80, 0x0, 0xc0002f14f0)
	/home/anomalroil/go/src/github.com/gopasspw/gopass/internal/action/config.go:21 +0x598
github.com/urfave/cli/v2.(*Command).Run(0xc0001f70e0, 0xc000366a80, 0x0, 0x0)
	/home/anomalroil/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/command.go:164 +0x4e2
github.com/urfave/cli/v2.(*App).RunContext(0xc000001c80, 0x561459ec4300, 0xc000343290, 0xc0001a6020, 0x2, 0x2, 0x0, 0x0)
	/home/anomalroil/go/pkg/mod/github.com/urfave/cli/v2@v2.2.0/app.go:306 +0x816
main.main()
	src/github.com/gopasspw/gopass/main.go:83 +0x420
```
### Steps To Reproduce
Have a mounted substore and then do:
``` 
gopass config
``` 

### Solution
This was introduced by the refactoring done #1391 

(Fix is incoming as a PR soon.)

### LINKED ISSUE #1459: Custom configuration not possible for non-remote substores
It seems that #1391 introduces a "regression" in the sense that it is not possible any more to set different options than for our main store for sub-store that are mounted and are non-remote.

There was the introduction of the notion of "remote config", this seems to still support custom options for a remote store, but for local stores, it appears that now the `gopass config --store` flag is ineffective.

Is this a decision that was made to drop custom configuration per-mounted substores, or is this an unwanted side effect?

This allows to "refactor" a bit the code, I'll drop a PR about it soon.

AnomalRoil:
Fixes #1458
This fixes a panic caused by a `nil` dereference since the store variable is usually "" because it is the _ root_ store.
Notice that `getOnDiskStorage` is really meant to work on substores, not on the root one, hence the panic.

Also notice that #1391 removed capacity to set per-mount config options for non-remote stores, see #1459. 
This is doing the small refactoring of the code mentioned in that issue in commit `85abbcc`.
Fixes #1459 

This PR is also adding some more integration tests to catch an issue like #1458 and a regression like #1459.

This is also adding a test to detect the wrong behaviour mentioned in #1460.

### DIFF
--- a/tests/config_test.go
+++ b/tests/config_test.go
@@ -1,22 +1,35 @@
 package tests
 
 import (
+	"strings"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"github.com/stretchr/testify/assert"
 )
 
-func TestConfig(t *testing.T) {
+func TestBaseConfig(t *testing.T) {
 	ts := newTester(t)
 	defer ts.teardown()
 
 	out, err := ts.run("config")
 	assert.NoError(t, err)
-	assert.Contains(t, out, "autoimport: true")
-	assert.Contains(t, out, "cliptimeout: 45")
-	assert.Contains(t, out, "confirm: false")
-	assert.Contains(t, out, "path: ")
-	assert.Contains(t, out, "safecontent: false")
+	wanted := `root store config:
+  autoclip: false
+  autoimport: true
+  cliptimeout: 45
+  confirm: false
+  editrecipients: false
+  exportkeys: false
+  nocolor: false
+  nopager: false
+  notifications: false
+`
+	wanted += "  path: " + ts.storeDir("root") + "\n"
+	wanted += "  safecontent: false"
+
+	assert.Equal(t, wanted, out)
 
 	invertables := []string{
 		"autoimport",
@@ -45,4 +58,36 @@ func TestConfig(t *testing.T) {
 		assert.NoError(t, err)
 		assert.Equal(t, "cliptimeout: 120", out)
 	})
+}
+
+func TestMountConfig(t *testing.T) {
+	ts := newTester(t)
+	defer ts.teardown()
+
+	// we add a mount:
+	_, err := ts.run("init --store mnt/m1 --path " + ts.storeDir("m1") + " --storage=fs " + keyID)
+	require.NoError(t, err)
+
+	_, err = ts.run("config")
+	assert.NoError(t, err)
+
+	wanted := `root store config:
+  autoclip: false
+  autoimport: true
+  cliptimeout: 45
+  confirm: false
+  editrecipients: false
+  exportkeys: false
+  nocolor: false
+  nopager: false
+  notifications: false
+  path: `
+	wanted += ts.storeDir("root") + "\n"
+	wanted += `  safecontent: false
+mount 'mnt/m1' => '`
+	wanted += ts.storeDir("m1") + "'\n"
+
+	out, err := ts.run("config")
+	assert.NoError(t, err)
+	assert.Equal(t, strings.TrimSpace(wanted), out)
 }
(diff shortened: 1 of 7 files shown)


tests/config_test.go
```diff
@@ -46,3 +59,35 @@ func TestConfig(t *testing.T) {
 		assert.Equal(t, "cliptimeout: 120", out)
 	})
 }
+
+func TestMountConfig(t *testing.T) {
+	ts := newTester(t)
+	defer ts.teardown()
+
+	// we add a mount:
+	out, err := ts.run("init --store mnt/m1 --path " + ts.storeDir("m1") + " --rcs=noop " + keyID)
```
dominikschulz, line 68: `--rcs=noop` isn't supported anymore as of #1455. Please use `--storage=fs` instead.

tests/mount_test.go
```diff
@@ -15,6 +15,52 @@ func TestSingleMount(t *testing.T) {
 	ts.initStore()
 	ts.initSecrets("")
 
+	out, err := ts.run("init --store mnt/m1 --path " + ts.storeDir("m1") + " --rcs=noop " + keyID)
```
dominikschulz, line 18: Same here. Please use `--storage=fs` instead of `--rcs=noop`.

tests/mount_test.go
```diff
@@ -23,36 +69,54 @@ func TestSingleMount(t *testing.T) {
...
+        │   ├── secret
+        │   └── twoliner
+        └── foo
+            └── bar`
 
 	out, err = ts.run("list")
 	assert.NoError(t, err)
-	t.Skip("integration test seems broken, but not manual tests")
+
+	t.Skip("this test is currently failing because the list is shadowing the mount")
```
dominikschulz, line 119: This is referencing #1460 I guess?
In that case the skipping is good until we fix it, but the reason seems to be odd (better than the previous one, but still). Shouldn't it say `this test is currently failing because the list command is not properly shadowing the mount`?

dominikschulz:
I'm afraid you'll have to rebase. #1463 fixed the tree regression.

AnomalRoil:
Alright, done @dominikschulz :)

dominikschulz:
Thank you, this is a great PR. Very much appreciated!

Issue #1474, opened 2020-08-01
gopass env crashes when missing command to execute


marco-m:
### Summary

The `env` subcommand (described in #1227, implemented in #1231) crashes when the command-line is missing the command to execute.

Assuming that:

* `foo` is an existing secret
* `packer` is an existing executable

This works:

    gopass env foo packer

This crashes (this is a mistake in the invocation, but it should not crash):

    gopass env foo

### Steps To Reproduce

See section above.

### Expected behavior

gopass should return an usage error, since this is a user mistake in composing the command-line.

### Environment

- OS: N/A
- OS version: N/A
- gopass Version: latest release (1.9.2) and build from today's master branch (commit 45a34f44d7).
- Installation method: N/A

### Additional context

PR #1475 adds tests and fixes the crash.

dominikschulz:
Yes, indeed. This is a bug. Thanks for providing a fix!

PR #1475, opened 2020-08-01
Fix env command crash

marco-m:
`env` command: do not crash if called without a command to execute

Fixes: #1474.

Best reviewed commit-per-commit.

As explained in the comments with the fix, I am not sure that is the correct place.

* env command: tests: make the test more stringent
* env command: tests: add test for secret not found
* env command: tests: add test for program not found
* FIX env command: do not crash if called without a command to execute

### DIFF
--- a/internal/action/env.go
+++ b/internal/action/env.go
@@ -20,6 +20,10 @@ func (s *Action) Env(c *cli.Context) error {
 	name := c.Args().First()
 	args := c.Args().Tail()
 
+	if len(args) == 0 {
+		return ExitError(ExitUsage, nil, "Missing subcommand to execute")
+	}
+
 	if !s.Store.Exists(ctx, name) && !s.Store.IsDir(ctx, name) {
 		return ExitError(ExitNotFound, nil, "Secret %s not found", name)
 	}
--- a/internal/action/env.go
+++ b/internal/action/env.go
@@ -12,7 +12,9 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
-// Env implements an env subcommand the populates the env of an subprocess with a set of secrets
+// Env implements the env subcommand. It populates the environment of a subprocess with
+// a set of environment variables corresponding to the secret subtree specified on the
+// command line.
 func (s *Action) Env(c *cli.Context) error {
 	ctx := ctxutil.WithGlobalFlags(c)
 	name := c.Args().First()
--- a/internal/action/env_test.go
+++ b/internal/action/env_test.go
@@ -13,7 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func TestEnv(t *testing.T) {
+func TestEnvLeafHappyPath(t *testing.T) {
 	u := gptest.NewUnitTester(t)
 	defer u.Remove()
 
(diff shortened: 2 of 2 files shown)


internal/action/env.go
```diff
@@ -20,6 +20,12 @@ func (s *Action) Env(c *cli.Context) error {
 	name := c.Args().First()
 	args := c.Args().Tail()
 
+	// FIXME? I have the feeling that this test should be done when parsing the command
```
marco-m, line 23: Unsure if this is the right place for the fix, but I don't know urfave...

marco-m:
I will be happy to make the CommitCheck bot happy, but I don't know how :-)

internal/action/env_test.go
```diff
@@ -34,6 +34,9 @@ func TestEnv(t *testing.T) {
 		stdout = os.Stdout
 	}()
 
+	// Command-line would be: "gopass env foo env", where "foo" is an existing
+	// secret with value "secret". We expect to see the key/value in the output
+	// of the /usr/bin/env utility in the form "FOO=secret".
 	assert.NoError(t, act.Env(gptest.CliCtx(ctx, t, "foo", "env")))
-	assert.Contains(t, buf.String(), "secret")
+	assert.Contains(t, buf.String(), "FOO=secret\n")
```
dominikschulz, line 41: This is a very good refinement for the existing test.
One could even consider populating foo with a long, random password to absolutely ensure that the correct secret is displayed. But this is already much better than before.

internal/action/env.go
dominikschulz: This looks correct to me.
What are you unsure about?

marco-m: I wasn't clear :-)
I meant: one could argue that this error should be detected and handled directly by urfave when parsing the command-line, in the same spirit as omitting a required argument or flag. But I am not familiar with urfave, so I don't know if this is possible at all.

If it is fine for you to keep the check here, then it is fine for me also :-)

internal/action/env_test.go
marco-m: Thanks.
Yes, I wanted to use something more unique than "secret", but it looks like that "secret" is the default value used by the mocking system (the test itself doesn't mention it), so I left it as-is. If you point me to how to inject a more unique value, I am happy to adapt the code here.

internal/action/env.go
dominikschulz: Maybe there is a way to do that, but I'm not aware of it.
So let's keep as is.

internal/action/env_test.go
dominikschulz: Maybe you want to leave a TODO for me to add that later?
Definitely ok as-is for this PR.

dominikschulz:
CommitCheck would like a RELEASE_NOTES=short change desc tag per commit.

internal/action/env.go
marco-m: removed comment

internal/action/env_test.go
marco-m: added TODO

marco-m:
> CommitCheck would like a RELEASE_NOTES=short change desc tag per commit.

done

marco-m:
@dominikschulz addressed all review comments.

PR #1476, opened 2020-08-02
env command: log only if gopass debug

marco-m:
- env command: log only if gopass debug (before, it was logging always)
- give some more information in the debug messages

Signed-off-by: Marco Molteni <marco.molteni@mailbox.org>


marco-m:
As with the previous PR, If I follow CommitCheck link, it doesn't tell me what is wrong...

Issue #1362, opened 2020-05-15
show command misbehavior


aleprovencio:
### Summary
The `show` argument used to print the full contents of a secret, now it just copies the password to clipboard. This is a bigger problem with multiline secrets.

### Steps To Reproduce
``gopass show secret.gpg``

### Expected behavior
Secret should be printed on stdout

### Environment
- Archlinux
- gopass 1.9.2-1

SalocinHB:
I'm seeing the same bug on macos 10.15. 

@aleprovencio as a workaround, you can use `gopass show -f secret.gpg`, that will print the secret instead of copying it, but I agree that this is pretty annoying.

aleprovencio:
Good catch @SalocinHB, that prevents downgrading :)

marco-m:
There is also another related bug: if you don't have `xsel` or `xclip`, you get a warning that mentions the non-existing flag `-p`:

```
$ gopass show foo/bar
WARNING: No clipboard available. Install xsel or xclip or use -p to print to console
```

and

```
gopass show -p foo/bar
Incorrect Usage: flag provided but not defined: -p

NAME:
   gopass show - Display a secret

USAGE:
   gopass show [command options] [arguments...]

DESCRIPTION:
   Show an existing secret and optionally put its first line on the clipboard. If put on the clipboard, it will be cleared after 45 seconds.

OPTIONS:
   --clip, -c        Copy the first line of the secret into the clipboard (default: false)
   --alsoclip, -C    Copy the first line of the secret and show everything (default: false)
   --qr              Print the first line of the secret as QR Code (default: false)
   --force, -f       Display the password even if safecontent is enabled (default: false)
   --password, -o    Display only the password (default: false)
   --sync, -s        Sync before attempting to display the secret (default: false)
   --revision value  Show a past revision
   --help, -h        show help (default: false)
```

@dominikschulz let me know if you want me to file a separate issue instead.

klartext:
Thanks for reporting this. It's really annoying not to have the answer on stdout.
Also bad: I don't want the output be in the clipboard.
If by accident copying the clipboard to where it does not belong, that's quite bad.

d1ee:
I have the same problem, but I just had to change to config and it works again.

```
# Somehow, autoclip config was true
$ gopass config autoclip
autoclip: true
```

```
$ gopass config autoclip false
```
OR
```
gopass config --store <storename> autoclip false
```

dominikschulz:
[...]

PR #1513, opened 2020-08-12
Remove autoclip integration for gopass show

### LINKED ISSUE #1441: gopass defaults to copy rather than show even if --clip, -c is NOT specified
### Summary 
`gopass somesite` copies the password rather than `show`-ing the entry which should be the default EVEN if `show` is specified explicitly.

### Steps To Reproduce
Execute:
`gopass somesite`

OR

`gopass show somesite`

### Expected behavior
The entry contents are displayed, ie:
```
secure-password
username: some-user
notes .etc
...
```

### Actual behavior
The password is copied to the clipboard and the contents are not shown EVEN IF `gopass show xxxx` is used.
```
somesite
✔ Copied sitesite to clipboard. Will clear in 45 seconds.
```

### Environment

> OS version:
```
❯ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G13033

❯ uname -a
Darwin mb2016.local 17.7.0 Darwin Kernel Version 17.7.0: Mon Apr 27 20:25:53 PDT 2020; root:xnu-4570.71.80~1/RELEASE_X86_64 x86_64
```
> - gopass Version:
```bash
gopass 1.9.2 (2020-06-04 20:35:38) go1.14.3 darwin amd64
<root>     - gpg 2.2.20 - git 2.25.0 -   fs 0.1.0
Available Crypto Backends: age, gpgcli, plain, xc
Available RCS Backends: gitcli, noop
Available Storage Backends: fs, inmem
Version check timed out

### for the source version
 /Users/vio/projects/go/bin/gopass version
gopass 1.9.2-git+HEAD go1.14 darwin amd64
<root>     - gpg 2.2.20 - git 2.25.0 -   fs 0.1.0
Available RCS Backends: gitcli, noop, ondisk
Available Storage Backends: fs, inmem, ondisk
```
- Installation method: `brew` (same behaviour with the version installed from source with `go get -u -v github.com/gopasspw/gopass`)

## Comments
This is extremely inconvenient because in order to see the full contents one needs to use `gopass edit somesite` which opens the contents in edit mode which is NOT the best user experience since it takes much longer than just printing the data to the terminal. If somebody needs to just copy the password they can either pass `-c` or create an alias. Please return the previous `show` behaviour!

### LINKED ISSUE #1502: Remove autoclip for gopass show
The `autoclip` config option was introduced for `gopass generated` and extended to cover `gopass show` as well at some point. However this leads to confusing and complex effects on the different flags and switches with no clear benefit.

We should remove the `autoclip` implementation in `gopass show` (only).
For `gopass generate` we'll likely want to keep it.

This is a child issue of #1488

dominikschulz:
Fixes #1502
Fixes #1441
Fies #1362 

RELEASE_NOTES=[CLEANUP] Remove autoclip for gopass show

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -48,7 +48,7 @@ This is a list of available options:
 | **Option**       | **Type** | Description |
 | ---------------- | -------- | ----------- |
 | `askformore`     | `bool`   | If enabled - it will ask to add more data after use of `generate` command.  DEPRECATED in v1.9.3 |
-| `autoclip`       | `bool`   | Always copy the password created by `pass generate`. |
+| `autoclip`       | `bool`   | Always copy the password created by `gopass generate`. Only applies to generate. |
 | `autoimport`     | `bool`   | Import missing keys stored in the pass repository without asking. |
 | `autosync`       | `bool`   | Always do a `git push` after a commit to the store. Makes sure your local changes are always available on your git remote. DEPRECATED in v1.9.3 |
 | `concurrency`    | `int`    | Number of threads to use for batch operations (such as reencrypting).  DEPRECATED in v1.9.3 |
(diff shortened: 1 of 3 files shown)


AnomalRoil:
Fixes #1362 too

dominikschulz:
Indeed, thank you!

PR #1559, opened 2020-08-26
Refactoring List

### LINKED ISSUE #1558: The `list` command is not behaving as expected
### Summary
We defined the behavior of the `list` command in #1488 
It is currently not behaving as discussed.

## Issues
For the `list` command,  the `--limit` flag should have a default value of `-1`, IMO, since the value 0 should only list the direct parents of the prefix (with / being the default prefix).

Currently the 0 value displays everything, but it should only display the entries in the given prefix. 

The `--limit` flag should starts counting its depth from the root store, which means that a depth of 0 should only lists the items directly in the root gopass store.

Furthermore the `-d` flag does not respect the  `--limit` as it should.

Finally the display of the `--flat` and `--folders` flags are not appending a separator at the end of the line when it's a folder.

NB: Fixes incoming in a PR.

AnomalRoil:
Fixes #1558

This is also including a refactoring of the action/list.go in order to avoid code duplication.

It is important to notice that now one must set MaxDepth to -1 in order to list all the items in a subtree instead of 0. The 0 is now only listing the direct parents of the subtree, 1 is listing the subitems, and so on as explained in docs/commands/list.md

I've also added some tests in order to make sure that the limit flag behaves as we expect it to.


internal/action/list.go
```diff
@@ -50,28 +51,33 @@ func (s *Action) List(c *cli.Context) error {
...
-		return ExitError(ExitNotFound, nil, "Entry '%s' not found", filter)
+
+	sep := string(filepath.Separator)
+
+	if filter == "" {
+		// We list all entries then
+		stripPrefix = true
+	} else {
+		// We restrict ourselves to the filter
+		var err error // to avoid shadowing l
```
dominikschulz, line 66: nit: Could you move this below the if statement, please?

dominikschulz: Also maybe extend the comment, e.g. "avoid shadowing l, we need it's value after the if clause" (or similar)

internal/action/recipients.go
```diff
@@ -53,7 +53,7 @@ func (s *Action) recipientsList(ctx context.Context) []string {
 		return nil
 	}
 
-	return tree.List(0)
+	return tree.List(-1)
```
dominikschulz, line 56: Should we have a constant for this special value?
`tree.INF` or something?

internal/tree/node.go
```diff
@@ -123,17 +125,29 @@ func (n *Node) list(prefix string, maxDepth, curDepth int, files bool) []string
 	}
 	prefix += n.Name
 
+	// if it's a file and we are looking for files
 	if n.Type == "file" && files {
+		//we return the file
```
dominikschulz, line 130: nit: space after `//`

dominikschulz (approved):
LGTM. Thanks for the additional tests and comments on the tree implementation!

internal/action/recipients.go
AnomalRoil: Sure, it involves importing tree in many places, but I guess it should be fine.

dominikschulz (approved):
LGTM

PR #1572, opened 2020-09-06
Re-introduce gopass -c

dominikschulz:
This PR re-introduces the `gopass -c` flag along the `-u` flag to make using the default action more convenient.

RELEASE_NOTES=[ENHANCEMENT] Re-introduce gopass -c

### DIFF
--- a/docs/commands/gopass.md
+++ b/docs/commands/gopass.md
@@ -8,6 +8,7 @@ has two different modes.
 ```
 $ gopass
 $ gopass entry
+$ gopass -c entry
 ```
 
 ## Modes of operation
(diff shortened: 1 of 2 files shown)


AnomalRoil:
This should probably be a draft, it's still missing the logic that would make it work.

dominikschulz:
Are you sure? The logic should be already there: https://github.com/gopasspw/gopass/blob/master/internal/action/show.go#L28

It just wasn't exposed for the default command. The code actually handling this was always there since it's being use by the `goass show` subcommand.

dominikschulz:
In my tests it works as expected.

AnomalRoil:
Ah perfect then.

It might have been better to add an integration test as well, but that's fine I guess.

dominikschulz:
Yes, that's true.

PR #1578, opened 2020-09-11
Fix notifications text on darwin

ilyaglow:
Hi,

I noticed that copy to clipboard notifications on mac (Catalina 10.15.6) looks almost blank:

Screenshot 2020-09-11 at 21 16 16

Proposed change fixes this behaviour.

Also, there is a problem with an icon in notifications, because macs don't have `~/.cache` folder by default and the recent change relies on it #1557, but this PR doesn't cover it.

### DIFF
--- a/internal/notify/notify_darwin.go
+++ b/internal/notify/notify_darwin.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"os"
 	"os/exec"
-	"strings"
 
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 )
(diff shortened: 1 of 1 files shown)


dominikschulz:
Thanks! I'll test it on MacOS before merging, but looks good.

Do you want to open a bug for the icon issue?

ilyaglow:
I made a separated follow up PR to fix it, but I'm not sure how it fits in the current code base #1580. I'll open an issue if you think it's worth it.

dominikschulz:
So I can't reproduce the problem without this fix, but with the fix it also works for me and I think your code is (more) correct.

dominikschulz (approved):
LGTM

PR #1579, opened 2020-09-15
Correcting newlines handling

### LINKED ISSUE #788: Inconsistency for newlines
Hi.
I have an issue with `msmtp` which requires me to add `; echo` at the end of `passwordeval "gopass show -f mail"`, otherwise, I get the following error:
```
msmtp: output of 'gopass show -f mail' is longer than 500 characters, or is not terminated by newline
```
I think it's because [the newline is not output when it's not in a tty](https://github.com/justwatchcom/gopass/blob/master/pkg/action/show.go#L127).
I think it would be better to have a consistent behavior to avoid this issue.
Thanks.

### LINKED ISSUE #1570: Edit Secret: ERROR: Can not parse Gopass native Secret.
### Summary
Since updating gopass to the latest release my workflow with this tool is a total pain. My latest problem is a broken edit function. Trying to edit and extend a secret causes the following error:

```
ERROR: Can not parse Gopass native Secret. Retry? [Y/n/q]: y
```

### Steps To Reproduce
Edit a secret and add an entry as `key: value`

### Environment
OS: Arch Linux
gopass Version:  1.10.1

### LINKED ISSUE #1573: gopass insert broken in 1.10 and appends a new line when reading from stdin
### Summary

When creating passwords from outside gopass and passing them via STDIN, in gopass 1.10.1 a newline is appended before the string, fundamentally changing the credentials.

### Steps To Reproduce
```
$ echo hello | gopass insert mystore/myvalue
$ gopass show -f mystore/myvalue

hello
$
```

### Expected behavior
```
$ echo hello | gopass insert mystore/myvalue
$ gopass show -f mystore/myvalue
hello
$
```

### Environment

```
$ gopass --version
gopass 1.10.1 go1.15 darwin amd64
$ uname -v
Darwin Kernel Version 19.6.0: Thu Jun 18 20:49:00 PDT 2020; root:xnu-6153.141.1~1/RELEASE_X86_64
```

- Installation method: brew

### LINKED ISSUE #1574: Safecontent is not warning anymore and displays a newline instead
### Summary

When we only had a secret in an entry that was `show`ed with `safecontent` enabled, it is supposed to error out and tell you to use `--unsafe, -u, -f` to display it.

It is currently not doing so and displays an empty line instead.

### Steps To Reproduce
```
$ ./gopass generate fo2 5  
$ ./gopass show fo2

$ ./gopass show -f fo2
LJrIS
```

### Expected behavior
```
$ ./gopass generate fo2 5  
$ ./gopass show fo2
Error: no safe content to display, you can force display with -f
$ ./gopass show -f fo2
LJrIS
```

### Environment

Gopass git tip

### LINKED ISSUE #1575: Gopass generate is adding a newline below the secret
### Summary
When using `generate`, the secret is added on the first line and a second empty line is added below it.

This is not handled properly by safecontent and such it displays an empty line when the secret is shown.

``` 
$ gopass generate fo6 5 
Password for fo6 generated
$ gopass show fo6


$ gopass show -f fo6
AQ5D7


``` 

Plus the show command is displaying an extra new line.

### LINKED ISSUE #1592: Certain flag combinations are undocumented or no-ops
It is possible sometimes to use a combination of flags that does nothing:

```
gopass generate test/noops 20
gopass show -c -u test/noops
gopass show -c -u -o test/noops
```

Also using:
```
gopass show -C -u test/noops
gopass show -C -o -u test/noops
```
will show all the content but not copy to the clipboard.

But `gopass show -C -o test/noops` works as I would expect it to.

There are probably other cases, but most issues seems to arise when using the `-u` flag

AnomalRoil:
Fixes #788
Fixes #1575
Fixes #1574
Fixes #1573
Fixes #1570
Fixes #1592

This is trying to correct the way we display and write newlines.
Newlines are appended by the MIME type after the header but won't be displayed by show now unless the MIME body is non-empty.

Notice the `mime` setting is significantly changing the way secrets are handled, written and displayed, also how the newlines are treated.
I've tried to add tests to take this into account.

I've also added unit tests and integration tests accordingly to try and detect such regressions in the future.

This is changing the behaviour of insert to make it compatible with the new MIME format by parsing its stdin input with the secret parser before writing it.

This is also refactoring the code a bit, removing dead code such as the YAML special handling.

### DIFF
--- a/internal/action/completion.go
+++ b/internal/action/completion.go
@@ -2,10 +2,11 @@ package action
 
 import (
 	"fmt"
-	"github.com/gopasspw/gopass/internal/tree"
 	"regexp"
 	"runtime"
 	"strings"
+
+	"github.com/gopasspw/gopass/internal/tree"
 
 	fishcomp "github.com/gopasspw/gopass/internal/completion/fish"
 	zshcomp "github.com/gopasspw/gopass/internal/completion/zsh"
--- a/internal/action/copy_test.go
+++ b/internal/action/copy_test.go
@@ -68,6 +68,6 @@ func TestCopy(t *testing.T) {
 	buf.Reset()
 
 	assert.NoError(t, act.show(ctx, c, "zab/bam/zab", false))
-	assert.Equal(t, "\nbarfoo\n", buf.String())
+	assert.Equal(t, "Password: barfoo\n", buf.String())
 	buf.Reset()
 }
--- a/internal/action/env.go
+++ b/internal/action/env.go
@@ -2,11 +2,12 @@ package action
 
 import (
 	"fmt"
-	"github.com/gopasspw/gopass/internal/tree"
 	"os"
 	"os/exec"
 	"path"
 	"strings"
+
+	"github.com/gopasspw/gopass/internal/tree"
 
 	"github.com/gopasspw/gopass/internal/debug"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
--- a/internal/action/find.go
+++ b/internal/action/find.go
@@ -3,9 +3,10 @@ package action
 import (
 	"context"
 	"fmt"
-	"github.com/gopasspw/gopass/internal/tree"
 	"sort"
 	"strings"
+
+	"github.com/gopasspw/gopass/internal/tree"
 
 	"github.com/gopasspw/gopass/internal/cui"
 	"github.com/gopasspw/gopass/internal/debug"
--- a/internal/action/fsck.go
+++ b/internal/action/fsck.go
@@ -1,9 +1,10 @@
 package action
 
 import (
-	"github.com/gopasspw/gopass/internal/tree"
 	"os"
 	"path/filepath"
+
+	"github.com/gopasspw/gopass/internal/tree"
 
 	"github.com/gopasspw/gopass/internal/config"
 	"github.com/gopasspw/gopass/internal/out"
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -3,12 +3,13 @@ package action
 import (
 	"context"
 	"fmt"
-	"github.com/gopasspw/gopass/internal/tree"
 	"path"
 	"regexp"
 	"sort"
 	"strconv"
 	"strings"
+
+	"github.com/gopasspw/gopass/internal/tree"
 
 	"github.com/gopasspw/gopass/internal/clipboard"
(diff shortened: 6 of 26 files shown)


AnomalRoil:
> Furthermore the order of the key-value entries is not sorted anymore
so that they get displayed as they've been written in the entry.

@dominikschulz  This is actually causing me an issue since the KV-values are a map and maps are randomly ranged over in Go.

But I believe we should guarantee the user that their password entries are going to be displayed as they've wrote them, what do you think?

dominikschulz:
I think it would be nice to keep the order, but I don't think the MIME parser can do that (and I certainly don't want to build our own).

AnomalRoil:
Well, I can see how I could had a field in the MIME header to handle the key ordering transparently, but not sure if we want the added complexity.

AnomalRoil:
Todos:

- [ ] Check that #788 is solved and/or if we want `-o` to output a newline no matter what, even if not in a tty.
- [ ] Correct documentation accordingly including the non-tty behaviour
- [ ] Add tests to avoid regressions and add tests for non-tty behaviour
- [ ] Make sure the tests would catch #1575 now 
- [ ] Add a new one if not (generate adding a newline at the end of the secret)
- [ ] Make sure the tests would catch #1574 now 
- [ ] Add a new one if not (safecontent warning not displayed with mime set to false)
- [ ] Make sure the tests would catch #1573 now  
- [ ] Add a new one if not (gopass `insert` was inserting newlines all over the place)
- [ ] Make sure the tests would catch #1570 now  
- [ ] Add a new one if not (no new line at the end of an edited secret would fail to be parsed)
- [ ] Investigate the strange behaviour I found where the secret and the body are both displayed on the same line depending on the `mime` and `safecontent` setting and the secret format  
- [ ] Open an issue about this ^

AnomalRoil:
Alright, I'm going to let this on the side for now:

- [ ] Correct documentation accordingly including the non-tty behaviour
- [ ] Add tests to avoid regressions and add tests for non-tty behaviour
- [ ]  Make sure the tests would catch #1570 now
- [ ] Add a new one if not (no new line at the end of an edited secret would fail to be parsed)

 because it proved harder to test/do than expected. 

But I could confirm it works with `msmtp` and its `passwordeval` setting, so that means non-tty behaviour is probably correct and #788 is definitively corrected.

dominikschulz (approved):
LGTM. Thanks a lot!

PR #1580, opened 2020-09-16
Create cache folder if doesn't exist

ilyaglow:
RELEASE_NOTES=[BUGFIX] Create cache folder if doesn't exist. Relevant
for darwin that doesn't have ~/.cache folder by default.

Signed-off-by: Ilya Glotov <contact@ilyaglotov.com>


dominikschulz (approved):
LGTM. Thank you!

PR #1587, opened 2020-09-18
Remove xc backend

dominikschulz:
RELEASE_NOTES=[CLEANUP] Remove the unfinished xc backend


AnomalRoil:
Nice cleanup 😍

dominikschulz:
Yes, I don't want to waste time dealing with this anymore as it likely won't have any future anyway. Maybe we'll need to resurrect some parts later, but then we can just grab it from the git history.

Issue #1561, opened 2020-08-27
Cannot find password store, instead creates a gpgcli-gitcli-fs+file:/ directory in the working directory


ondrik:
### Summary
Hi! Many thanks for providing this great password manager! From a few days ago, I've started having the following issue.

Since gopass 1.10.0, gopass seems it cannot find the password store in `~/.password-store`. Instead, it creates the directory `gpgcli-gitcli-fs+file:/` in the current directory.

### Steps To Reproduce
In an empty directory:
```
$ ls
$ gopass
It seems you are new to gopass. Do you want to run the onboarding wizard? [Y/n/q]: q
$ ls
gpgcli-gitcli-fs+file:/
```

### Environment

- OS: Mac OS X Catalina
- OS version: 19.6.0 Darwin Kernel
- gopass Version: 1.10.0, 1.10.1
- Installation method: brew

### Additional context
Content of `~/.config/gopass/config.yml`:
```{yaml}
root:
  askformore: false
  autoclip: false
  autoimport: true
  autosync: true
  check_recipient_hash: false
  cliptimeout: 45
  concurrency: 1
  editrecipients: false
  nocolor: false
  noconfirm: true
  nopager: false
  notifications: false
  path: gpgcli-gitcli-fs+file:///Users/ondra/.password-store
  recipient_hash: {}
  safecontent: true
  usesymbols: false
mounts: {}
version: 1.8.1
```

dominikschulz:
Uh oh, that sounds like a config migration bug. The old scheme prefix should be automatically stripped.
Thanks for reporting and providing a copy of your config. Will use that in an attempt to reproduce.

jonathanrohland:
@dominikschulz I ran into this problem as well and would like to help writing a fix (or maybe add an appropriate warning). However, I'm not sure what your plan going forward with this is. I took a look into the code but since I'm not familiar with the project yet it's quite likely that I haven't found every relevant piece of code. If I understood it correctly the problem occurs because the config-value for the path previously used to contain the prefix `gpgcli-gitcli-fs+file://`, but now it doesn't anymore.
Introducing a legacy-config-type (as in `config/legacy.go`) doesn't seem like a viable solution, since the format of the config hasn't actually changed (only the format of the expected value).

I can see a few ways out of this:
* Checking the config on load and logging a warning (+exiting) when the obsolete prefix is found.
* Checking the config on load and dropping the prefix (either silently or with a warning message)
[...]

PR #1588, opened 2020-09-20
Update fixed path in Pre182 legacy-config

jonathanrohland:
Hello everyone,
this PR should fix https://github.com/gopasspw/gopass/issues/1561. The issue was just a typo that caused the change to be written to the input-object instead of the output-object.
Let me know if this works for you and thanks for maintaining gopass!


jonathanrohland:
I'll fix the commits and will open a new PR.

PR #1596, opened 2020-09-25
Adding back the global yes flag and bugfix insert

### LINKED ISSUE #1595: Non-interactive mode not working
### Summary
I try the following with GoPass 1.10.1 on MacOS Catalina:
`gopass --yes insert --force secrets/test username: test password: test`

and i receive the following error message:
Incorrect Usage. flag provided but not defined: -yes

I remember in older versions the flag **--yes** worked fine. Was there any change?

### Expected behavior
Inserting secrets non interactively!

### Environment
- OS: MacOS X Catalina
- OS version: 10.15.6
- gopass Version: gopass 1.10.1 go1.15 darwin amd64
- Installation method: brew

AnomalRoil:
Fixes #1595

RELEASE_NOTES=[BUGFIX] Re-adding the global --yes flag
RELEASE_NOTES=[BUGFIX] Insert is not resetting the pw now if a key:value pair is specified inline

This is just adding --yes back since it was mistakenly removed.

It is also correcting the way insert behaves if we use inline key:values:

`gopass insert test key:lol` will no longer overwrite the password with the empty value is none is set. 
Thus `gopass --yes insert -f test newkey:lol oldkey:yay` will not also overwrite the key password whereas previously it would.

### DIFF
--- a/docs/commands/gopass.md
+++ b/docs/commands/gopass.md
@@ -25,4 +25,5 @@ Flag | Aliases | Description
 ---- | ------- | -----------
 `--clip` | `-c` | Copy the password value into the clipboard and don't show the content.
 `--unsafe` | `-u` | Display unsafe content (e.g. the password) even when the `safecontent` option is set. No-op when `safecontent` is `false`.
+`--yes` |  | Assume yes on all yes/no questions or use the default on all others.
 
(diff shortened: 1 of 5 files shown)


dominikschulz (approved):
LGTM. Thanks!

Issue #1473, opened 2020-07-28
No OTP entry found: illegal base32 data at input byte 48


Rubonnek:
### Summary

Amazon secret keys can be 52 bytes long. Attempting to generate OTP for these result in the following:

```
Error: No OTP entry found for Test: illegal base32 data at input byte 48
```

### Steps To Reproduce
*Test.gpg*
```
Test
---
totp: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

Then execute:
```shell
gopass otp Test
```

### Expected behavior
OTP password is generated.

### Environment

- OS: Arch Linux
- OS version: Linux harmonia 5.4.53-1-lts
- gopass Version: gopass 1.9.2 (2020-07-13 14:34:42) go1.14.4 linux amd64
- Installation method: Arch Linux `community` repository

duckpuppy:
The key they provide just needs padding - for a 52 character TOTP key, append `====` to pad it out to an 8-character boundary. This could be done automatically.

See: https://github.com/rsc/2fa/pull/4

Rubonnek:
Oh wow, I wasn't aware of that. I was able to get it to work with that. Thanks!

nokernel:
I am doing some attempt at fixing this. I might need some guidance after I have something that I think works.

Some testers to see if otp still works.

PR #1598, opened 2020-09-25
Correct missing padding to TOTP entry.

nokernel:
Fixes https://github.com/gopasspw/gopass/issues/1473

RELEASE_NOTES=[BUGFIX] Correct missing padding to TOTP entry

This is some attempt to correct the missing padding for TOTP entry.
One example of those TOTP are the ones from Amazon that are sometime
52bytes in length.


dominikschulz (approved):
LGTM. Thank you!

PR #1603, opened 2020-09-29
Add support for libstoken and RSA token generation

buchanan:


AnomalRoil:
Nice PR,

But, this one is going to be a tricky one because it requires adding an `import "C"` statement to the codebase while we are trying to refactor as much as possible to avoid external calls to stuff such as git, GPG, etc.

Isn't there a way to support stokens using a native Go library?

buchanan:
Thanks, I think I could rewrite the library in go since its open source or I could refactor the code to use the stoken binary instead of the library and use the go native os.exec to generate the tokens.

https://github.com/cernekee/stoken

dominikschulz:
Thanks a lot for your effort, but I'm afraid we won't be able to merge any kind of CGO dependency in the foreseeable future.
This would significantly increase our maintenance burden and we're currently not staffed to handle that.

So in the interest of transparency I'd close this PR with some regrets and kindly ask you to consider a pure-Go implementation if you have the time to do so.

Issue #1602, opened 2020-09-27
add a config option to disable entry parsing


deadc0de6:
### Summary

I'd like to be able to freely edit entries using gopass and not get parts of the entered text discarded or re-ordered (#1600, #1601). Please provide a config option to disable any kind of parsing on the content of an entry (for `insert -m` or `edit`).

lsowen:
I think this would be a useful feature.  Maybe a CLI flag or environment variable to disable all parsing, and treat any `insert` / `show` as just an "opaque" string.

AnomalRoil:
This is pushed to 1.11 when we will re-introduce the (optional) parsing for Mime secrets.

For 1.10.2 we will disable input parsing altogether.

PR #1608, opened 2020-10-03
Preserve content across retries during edit

### LINKED ISSUE #1593: The retry dialog drops unwritten changes
### Summary
When we `edit` a secret and it fails, the "Retry" dialog proposes to retry and re-open the editor, but it does so while dropping all the changes, which is a terrible UX.

### To reproduce 
```
gopass generate test 5
echo "testing" | gopass insert test user
gopass edit
```
This should display:
```
Password: Gv1rX
User: testing
```

Now edit it to look like:

```
Password: Gv1rX
User: testing
This is a nice body text without a new line between the key-value part and the body.
```
And save&exit your editor.
--> gopass complains about the secret not being in Native format (i.e. good Mime format) and proposes to "Retry: Y/n/q".
if you retry, it will re-open to:
```
Password: Gv1rX
User: testing
```
Having discarded all the changes.

This is really a bad UX, leading to frustration on the user side.

This is maybe related or maybe a duplicate of #1565 but triggered by another code path.

dominikschulz:
Fixes #1593

RELEASE_NOTES=[UX] Preserve content across retries

### DIFF
--- a/internal/action/edit.go
+++ b/internal/action/edit.go
@@ -46,16 +46,22 @@ func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
 		content = bytes.TrimPrefix(content, []byte(secret.Ident+"\n"))
 	}
 
+	// preserve intermediate state across retries
+	tmpContent := content
 	for {
 		// invoke the editor to let the user edit the content
-		nContent, err := editor.Invoke(ctx, ed, content)
+		newContent, err := editor.Invoke(ctx, ed, tmpContent)
 		if err != nil {
 			return ExitError(ExitUnknown, err, "failed to invoke editor: %s", err)
 		}
+		// retain edited content for next try
+		tmpContent = newContent
 		if fromMIME {
-			nContent = append([]byte(secret.Ident+"\n"), nContent...)
-			if _, err := secparse.Parse(nContent); err != nil {
-				cont, err := termio.AskForBool(ctx, "ERROR: Can not parse Gopass native Secret. Retry?", true)
+			newContent = append([]byte(secret.Ident+"\n"), newContent...)
+			if _, err := secparse.Parse(newContent); err != nil {
+				out.Red(ctx, "ERROR: Cannot parse Gopass native Secret: %s", err)
+				out.Yellow(ctx, "Hint: Does your Key-Value section end with a new line?")
+				cont, err := termio.AskForBool(ctx, "Retry?", true)
 				if err != nil {
 					return err
 				}
@@ -65,7 +71,7 @@ func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
 				continue
 			}
 		}
-		return s.editUpdate(ctx, name, content, nContent, changed, ed)
+		return s.editUpdate(ctx, name, content, newContent, changed, ed)
 	}
 }
 


internal/action/edit.go
```diff
@@ -46,26 +46,32 @@ func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
...
 			return ExitError(ExitUnknown, err, "failed to invoke editor: %s", err)
 		}
+		tmpContent = newContent
 		if fromMIME {
-			nContent = append([]byte(secret.Ident+"\n"), nContent...)
-			if _, err := secparse.Parse(nContent); err != nil {
-				cont, err := termio.AskForBool(ctx, "ERROR: Can not parse Gopass native Secret. Retry?", true)
+			newContent = append([]byte(secret.Ident+"\n"), newContent...)
+			if _, err := secparse.Parse(newContent); err != nil {
+				out.Red(ctx, "ERROR: Can not parse Gopass native Secret: %s", err)
```
AnomalRoil, line 61: "cannot"

internal/action/edit.go
```diff
@@ -46,26 +46,32 @@ func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
...
+	tmpContent := content
 	for {
 		// invoke the editor to let the user edit the content
-		nContent, err := editor.Invoke(ctx, ed, content)
+		newContent, err := editor.Invoke(ctx, ed, tmpContent)
 		if err != nil {
 			return ExitError(ExitUnknown, err, "failed to invoke editor: %s", err)
 		}
+		tmpContent = newContent
 		if fromMIME {
```
AnomalRoil, line 58: What happens if it is not `fromMime` ?? Is it then not parsed at all?
We might want to do the same thing for insert wrt. #1602

internal/action/edit.go
```diff
@@ -46,26 +46,32 @@ func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
...
+				out.Red(ctx, "ERROR: Can not parse Gopass native Secret: %s", err)
+				out.Yellow(ctx, "Hint: Does your Key-Value section end with a new line?")
+				cont, err := termio.AskForBool(ctx, "Retry?", true)
 				if err != nil {
 					return err
 				}
 				if !cont {
 					return fmt.Errorf("malformed secret")
 				}
+				// retain edited content for next try
```
AnomalRoil, line 70: this could rather be a comment for line 57, no?

AnomalRoil (dismissed):

internal/action/edit.go
dominikschulz: Yes, if it's not fromMime it's not parsed at all.
And yes, we might want to do that for insert as well. But certainly in a different PR.

internal/action/edit.go
dominikschulz: Done.

internal/action/edit.go
dominikschulz: Uhm, yes. Indeed. I moved some things around.

internal/action/edit.go
AnomalRoil: Agreed

AnomalRoil (approved):

PR #1610, opened 2020-10-03
Add support for inserting secrets with colons in the name

### LINKED ISSUE #1609: adding secret with colon in name fails
### Summary
A secret name with a colon is interpreted as a K/V pair, instead of a secret name.

### Steps To Reproduce
```bash
 ~$ gopass insert test:test

Error: Usage: gopass insert name
```

### Expected behavior
If only  a single element is specified, then it should be taken as the secret name, even if it successfully parses as a K/V pair.

### Environment

- OS: Fedora 32
- gopass Version: gopass 1.9.2+e2d1549f452a0df1fc52e42e7d0f654334d7144e (e2d1549f452a0df1fc52e42e7d0f654334d7144e) go1.14.2 linux amd64

### Additional context
Causes problems with using gopass as the backend for the lastest version of `aws-vault`, which stores the OIDC token as `gopass insert -f -m oidc:<hostname>`

lsowen:
Fixes #1609 

Signed-off-by: lsowen <logan@s1network.com>

### DIFF
--- a/internal/action/clihelper.go
+++ b/internal/action/clihelper.go
@@ -18,6 +18,13 @@ func (a argList) Get(n int) string {
 func parseArgs(c *cli.Context) (argList, map[string]string) {
 	args := make(argList, 0, c.Args().Len())
 	kvps := make(map[string]string, c.Args().Len())
+	if c.Args().Len() == 1 {
+		// If there is only one arg, assume it is
+		// the secret name, so don't attempt to
+		// parse into args and kvps
+		args = append(args, c.Args().Get(0))
+		return args, kvps
+	}
 OUTER:
 	for _, arg := range c.Args().Slice() {
 		for _, sep := range []string{":", "="} {
--- a/internal/action/clihelper_test.go
+++ b/internal/action/clihelper_test.go
@@ -24,6 +24,11 @@ func TestParseArgs(t *testing.T) {
 			argOut: argList{"foo/bar"},
 		},
 		{
+			name:   "secret with colon",
+			argIn:  []string{"foo/bar:test"},
+			argOut: argList{"foo/bar:test"},
+		},
+		{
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 [BUGFIX] Remove misleading config error message
 [BUGFIX] Re-use existing root store
 [BUGFIX] Use standard Unix directories on MacOS
+[BUGFIX] Allow secret names to have a colon in the name
```
dominikschulz, line 7: Thank you for remembering to add this to the changelog.

In the past we didn't enforce this so we have a CommitCheck that looks for a RELEASE_NOTES= entry.

Could you please add `RELEASE_NOTES=[BUGFIX] Allow secret names to have a colon in the name` to the commit message so we can automatically parse that when we create the next release?

internal/action/clihelper.go
```diff
@@ -18,6 +18,10 @@ func (a argList) Get(n int) string {
 func parseArgs(c *cli.Context) (argList, map[string]string) {
 	args := make(argList, 0, c.Args().Len())
 	kvps := make(map[string]string, c.Args().Len())
+	if(c.Args().Len() == 1) {
```
dominikschulz, line 21: This deserves a short comment.

dominikschulz (changes requested):
Thank you very much!

Two minor nits, otherwise good.

dominikschulz (approved):
LGTM. Thank you!

dominikschulz:
You have an extra set of brackets where Go doesn't want them.
If you remove those this should be good to merge.

lsowen:
Too much java, and not enough golang lately :grin:

dominikschulz (approved):
LGTM

PR #1611, opened 2020-10-03
Do not duplicate K/V password key

### LINKED ISSUE #1607: Duplicated field in entry
### Steps To Reproduce
`gopass edit test` and add the following:

```
secret
password: oth3rpa4ss
``` 
and then `gopass show -u test` will display:

```
secret
password: oth3rpa4ss
password: oth3rpa4ss

``` 
instead, with a duplicated password entry!

### Expected behavior
Do not duplicate any field.

### Environment

- OS: Ubuntu 18.04
- gopass Version master tip
- Installation method: from source

Probably related to #1601 and/or #1600

dominikschulz:
RELEASE_NOTES=[BUGFIX] Do not duplicate key password in K/V secrets
RELEASE_NOTES=[ENHANCEMENT] Prompt for edit search result

Fixes #1607

### DIFF
--- a/internal/action/edit.go
+++ b/internal/action/edit.go
@@ -101,7 +101,13 @@ func (s *Action) editGetContent(ctx context.Context, name string, create bool) (
 			return nil
 		}
 		if err := s.find(ctxutil.WithFuzzySearch(ctx, false), nil, name, cb); err == nil {
-			name = newName
+			cont, err := termio.AskForBool(ctx, fmt.Sprintf("No secret at '%s'. Found possible match in '%s'. Edit existing entry?", name, newName), true)
+			if err != nil {
+				return "", nil, false, err
+			}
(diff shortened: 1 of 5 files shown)


AnomalRoil (approved):
Mhhh, I think this is actually related to #1607 
What happens if we have 2 passwords entries? 

Do we really want to have a special handling for the password key in the yaml and kv format??

This does fix the initial issue though.
Would you mind adding a regression test? Maybe in the integration tests?

dominikschulz:
No, I don't want to have special handling for those. In another branch I did already remove this special casing.
But since I don't know when this might be ready I thought I'd fix the original issue first.

I'll try to add a regression test.

AnomalRoil (approved):

AnomalRoil:
LGTM

PR #1613, opened 2020-10-04
Do not require empty new line terminating MIME header

### LINKED ISSUE #1593: The retry dialog drops unwritten changes

dominikschulz:
Fixes #1593

RELEASE_NOTES=[BUGFIX] Do not require final new line for MIME

### DIFF
--- a/pkg/gopass/secret/mime.go
+++ b/pkg/gopass/secret/mime.go
@@ -3,11 +3,14 @@ package secret
 import (
 	"bufio"
 	"bytes"
+	"errors"
 	"fmt"
 	"io"
 	"net/textproto"
 	"sort"
 	"strings"
+
+	"github.com/gopasspw/gopass/internal/debug"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/gopass/secret/mime.go
```diff
@@ -80,7 +83,10 @@ func ParseMIME(buf []byte) (*MIME, error) {
 	tpr := textproto.NewReader(r)
 	m.Header, err = tpr.ReadMIMEHeader()
 	if err != nil {
-		return nil, &PermanentError{Err: err}
+		if !errors.Is(err, io.EOF) {
```
AnomalRoil, line 86: Shoot I should have catched it in #1579 

I actually thought I did: https://github.com/gopasspw/gopass/blob/10e2e0ece4717a1be986ed8cb076d7da08d46486/pkg/gopass/secret/mime.go#L82-L83

I don't understand where this disappeared.

dominikschulz:
Ha, indeed. My bad. Somehow I had an ancient version of `master` checked out.

PR #1629, opened 2020-10-23
Handling key-values instead of key-value in Mime

### LINKED ISSUE #1576: Having multiple Password fields causes strange behaviour
### Summary
Using the `mime: true` setting and having multiple key-value with the same key causes unexpected behaviour, with the ordering of the different key being lexicographical instead of as per the actual order in the entry.

### Steps To Reproduce
 Create an entry containing:
```
Password: rockyou
Password: 1
Password: 2
Password: test
Password: 0

```
(Make sure to include the new line at the end so that it's using the new native secret format.)

And then show it using :
```
$ gopass edit fo9
Warning: Password is too short
$ gopass show -f fo9 
Password: 0
Password: 1
Password: 2
Password: rockyou
Password: test

$ gopass show -o fo9
0
``` 

As you can see the entries are display in another order than what we inserted in the edit mode.

And then test changing the config `mime` to false and show or edit it: all the other keys with "password" will have disappeared!

```
$ gopass config mime false
mime: false
$ gopass show -u fo9
0

$ gopass edit fo9
```

This is really not the behaviour we want with mime set to false, we'd like to have all the content displayed rather than the Password fields being squashed into a single one.

The worst is that if you start editing it or generating a new password for the Password field, it will misbehave even more !
Notice this can cause loss of data:

```
$ gopass show -f fo11
Password: 123
Password: 124
Password: 124

$ gopass generate fo11 password 4 
Password for fo11:password generated              
$ gopass show -u fo11
Password: A0Ni

```

### Expected behaviour
We should clarify the expected behaviour.

IMO:
 1. we should display entries in the order they are written in the secret file
 1. we should error out or ask the user when there are multiple entries with the same key and they do a generate or insert on that key
 1. we should make sure the mime setting is not causing strange stuff if we switch it `on` and `off` (maybe specify how the transition is supposed to happen for all the existing types of secrets?)

### LINKED ISSUE #1601: entry fields disappear
### Summary

When creating a multiline entry with more than one field with the same key, only the last entry is kept, the rest is entirely discarded.

### Steps To Reproduce

create a new multiline entry
```bash
$ gopass insert -m badtest2
```

insert the following
```
thepassword
user: a user
web: test.com
user: second user 
```

save and show the entry
```bash
$ gopass show badtest2          
thepassword
user: second user
web: test.com
```

### Expected behavior

entered text do not get altered.

### Environment

- OS: linux
- gopass Version: 1.10.1

### Additional context

see also #1600

AnomalRoil:
Fixes #1601
Fixes #1576

RELEASE_NOTES=[ENHANCEMENT] Mime type supports multiple entries with same key instead of storing last key-value entry like KV.
RELEASE_NOTES=[ENHANCEMENT] The Mime type now preserves the order of the inserted keys

This allows the new Mime type to handle multi-entries easily.

Notice this will also attempt to transform KV entries into Mime entries upon insertion when mime is enabled.

This does not change the current behaviour when mime is set to false.


AnomalRoil:
Notice this does not change the behaviour of `edit` yet, but upon insertion we should probably do the same as upon editing a new secret, I guess.

AnomalRoil:
Looks like the travis is broken:
```
     CYCLO     go: downloading github.com/fzipp/gocyclo v0.3.1
go: github.com/fzipp/gocyclo upgrade => v0.3.1
/bin/sh: 1: gocyclo: not found
make: *** [codequality] Error 1
The command "make travis" exited with 2.
```

dominikschulz:
Could you please rebase?

AnomalRoil:
@dominikschulz Works like that?

dominikschulz (approved):
LGTM

AnomalRoil:
If we are to drop Mime altogether, this can be closed since it specifically did not handle the KV case wrt to multiple keys with values.

dominikschulz:
I'm sorry for making this one obsolete, but the MIME approach was so flawed :/

AnomalRoil:
I'll come back with a PR for supporting key-valueS for the KV type ;)

Issue #1634, opened 2020-10-30
jsonapi to support edit


rounakdatta:
### Summary
Today gopass-jsonapi does not support the API to edit a password. Having that feature would definitely be useful.

### Steps To Reproduce
https://github.com/gopasspw/gopass/blob/master/docs/jsonapi.md - the documentation clearly mentions which APIs are available, and `edit` is not a part of it.

### Expected behavior
An API corresponding to `edit`
- request
```
{
   "type": "edit",
   "login": "myusername",
   "password": "mynewpassword"
}
```

- response
```
{
   "username": "myusername",
   "password": "mynewpassword"
}
```

### Environment

- OS: macOS Catalina
- OS version: Darwin
- gopass Version: 1.10.1
- Installation method: source/brew

### Additional context
Having this API as a part of the jsonapi would open up multiple other opportunities like feature of editing directly from the gopassbridge extension. Nice to have.

Pharb:
Sounds like a useful feature. 👍 
We can discuss the implementation details in the PR.

For reference, this relates to: https://github.com/gopasspw/gopassbridge/issues/178

PR #1639, opened 2020-10-31
Add support for edit as part of jsonapi

rounakdatta:
An enhancement implementation as discussed under https://github.com/gopasspw/gopass/issues/1634.

### DIFF
--- a/cmd/gopass-jsonapi/internal/jsonapi/responses.go
+++ b/cmd/gopass-jsonapi/internal/jsonapi/responses.go
@@ -38,6 +38,8 @@ func (api *API) respondMessage(ctx context.Context, msgBytes []byte) error {
 		return api.respondGetData(ctx, msgBytes)
 	case "create":
 		return api.respondCreateEntry(ctx, msgBytes)
+	case "edit":
+		return api.respondEditEntry(ctx, msgBytes)
 	case "copyToClipboard":
 		return api.respondCopyToClipboard(ctx, msgBytes)
 	case "getVersion":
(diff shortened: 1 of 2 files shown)


cmd/gopass-jsonapi/internal/jsonapi/responses.go
```diff
@@ -197,6 +199,35 @@ func (api *API) respondCreateEntry(ctx context.Context, msgBytes []byte) error {
...
 }
 
+func (api *API) respondEditEntry(ctx context.Context, msgBytes []byte) error {
+	var message editEntryMessage
+	if err := json.Unmarshal(msgBytes, &message); err != nil {
+		return errors.Wrapf(err, "failed to unmarshal JSON message")
+	}
+
+	if _, err := api.Store.Get(ctx, message.Name, "latest"); err != nil {
+		return fmt.Errorf("secret %s does not exist", message.Name)
```
Pharb, line 209: Thanks for the PR @rounakdatta.

Maybe it would be even simpler to just reuse the "create" command and allow it to update existing secrets?
As far as I see, the only difference here is the error handling.

Alternatively at least the function should be re-used, instead of duplicating the logic.

What's your preferred solution?

Pharb (changes requested):

cmd/gopass-jsonapi/internal/jsonapi/responses.go
rounakdatta: Sorry for getting back late @Pharb , agree to your point. I've simplified it now, and got rid of the unnecessarily introduced `editEntryMessage ` struct. Thanks, please review and recommend next steps.

AnomalRoil:
Thanks for the work, could you please make sure your commit message is following [our guidelines for contributing](CONTRIBUTING.md#pull-request-checklist )? 
It helps to have a nice changelog and everything.

dominikschulz:
I'm not so sure if I want to merge this PR.
So far gopassbridge has been read-only. IIUC this will introduce a new possible attack vector.
At the very least I want a clear use case definition upfront.

So if you believe this is an important addition (and not just "nice to have" - we have too much "nice to have" features already) then please update this PR to include a motivation (e.g. in the docs/usecases folder) for this feature, how it's going to be used and the possible attack vectors you can think of. Once we have that I'll review the PR.

rounakdatta:
Hi @dominikschulz, the idea of this PR was to extend the capability of the already present `respondCreateEntry` so that it supports editing (which essentially means overwriting an existing record).

> So far gopassbridge has been read-only.

But we already have the new entry capability right?

- **A possible attack vector**
Since gopassbridge would have "overwrite" access with this PR, the worst exploit would mean that all the entries are overwritten. True that this attack is not yet possible given currently it allows creation of records with non-existent keys.
However this worst exploit is only causing damage to the local pass repository, and "in case" there is a remote origin, possibility of restoration still remains :).

WDYT?

Also, would be interested to understand how such attack vectors could be carried out without direct access to the system. Thanks for pointing out.

TODO for me:
- Add docs on how this feature would be used.

dominikschulz:
@rounakdatta Sounds good. Could you add the docs? Then we could merge this.

PR #1642, opened 2020-11-02
Add chezmoi to list of integrations

twpayne:


AnomalRoil (approved):
LGTM. 
You might want to add the code fences like the other tools, if it's a cmd line tool.

dominikschulz (approved):
LGTM

twpayne:
> You might want to add the code fences like the other tools, if it's a cmd line tool.

Ah, I've just worked out what this means. In the case of chezmoi, I make a distinction between chezmoi the software, and `chezmoi` the command you run. A similar distinction exists between Go the programming language and `go` the tool you often use when developing software in Go. I'll happily add code fences if you want, but in the case of chezmoi, the non-code-fenced version is more correct.

AnomalRoil:
Perfect then :)

PR #1651, opened 2020-11-12
Fix wrong list depth in small stores

### LINKED ISSUE #1628: List output depth is dependant on the number of file entries
### Summary
When creating nested directories with a depth deeper than the number of existing file entries some branches and leaves are not displayed.

Example output:
```
➜  gopass list
gopass
└── misc/
    ├── file1
    └── folder1/
        └── folder2/
```

Expected example output:
```
➜  gopass list
gopass
└── misc/
    ├── file1
    └── folder1/
        └── folder2/
            └── file2
```

Adding another file entry under misc/ will result in the following output:
```
➜  gopass list
gopass
└── misc/
    ├── file1
    ├── file3
    └── folder1/
        └── folder2/
            └── file2
```

Culprit seems to be that the limit is set based on the length (number of files) of the store at https://github.com/gopasspw/gopass/blob/master/internal/action/list.go#L49 thus exiting here: https://github.com/gopasspw/gopass/blob/master/internal/tree/node.go#L64

I initially observed this behavior when creating a team store with one entry in the misc folder and zero entries in the local store. This resulted in only the folders being printed. Since the file entry in the team store would set the limit to 1 while having a depth of 3.

### Steps To Reproduce
1. gopass insert misc/file1
3. gopass insert misc/folder1/folder2/file2
4. gopass list

### Expected behavior
I would expect that all the available credentials over the entire hierarchy are printed (up until maybe a sensible limit).

### Environment
- OS: Mac OS X Catalina
- OS version: Darwin --- 19.6.0 Darwin Kernel Version 19.6.0: Sun Jul  5 00:43:10 PDT 2020; root:xnu-6153.141.1~9/RELEASE_X86_64 x86_64
- gopass Version: v1.10.1 (via brew), current master (from source)

AnomalRoil:
Fixes #1628

RELEASE_NOTES=[BUGFIX] List no longer uses the store size as its default depth


dominikschulz (approved):
LGTM

PR #1653, opened 2020-11-14
Link to doc on vim potential leak

oliverpool:
I just got the warning, but I had to do quite a search to find what the actual issue was, and the fix for this.

This PR simply adds a link to the doc after the warning message.


dominikschulz (approved):
LGTM

Issue #1633, opened 2020-10-28
Update documentation for gopass summon provider usage


ckolumbus:
The documentation for using `gopass` as `summon` provider is outdated. It describes the use of the `gopass` command as a provider. 

`gopass-summon-provider` has been implemented to avoid impact of the `gopass` console output on the `summon` interface.

This new command needs to be used in the description of https://github.com/gopasspw/gopass/blob/master/docs/summon-provider.md

see also #1627

AnomalRoil:
closed by #1658

PR #1658, opened 2020-11-22
doc: summon.md: we now need gopass-summon-provider

marco-m:
RELEASE_NOTES=n/a

Signed-off-by: Marco Molteni <marco.molteni@mailbox.org>

### DIFF
--- a/docs/summon-provider.md
+++ b/docs/summon-provider.md
@@ -7,13 +7,13 @@ It is used to execute a process and inject secrets from a separate store. Using 
 
 ## Summon Provider
 
-gopass can be used as [summon provider](https://cyberark.github.io/summon/#providers) out of the box, since it fulfills the summon provider contract.
+The gopass repository contains the [cmd/gopass-summon-provider](../cmd/gopass-summon-provider) tool, that can be used as [summon provider](https://cyberark.github.io/summon/#providers).
 
-To make use of gopass to retrieve the `test/db-password` secret, you can call summon with full provider path
+To make use of gopass-summon-provider to retrieve the `test/db-password` secret, you can call summon with full provider path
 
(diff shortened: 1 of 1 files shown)


marco-m:
This closes #1633

docs/summon-provider.md
```diff
@@ -7,13 +7,13 @@ It is used to execute a process and inject secrets from a separate store. Using
 
 ## Summon Provider
 
-gopass can be used as [summon provider](https://cyberark.github.io/summon/#providers) out of the box, since it fulfills the summon provider contract.
+The gopass repository contains the [cmd/gopass-summon-provider](https://github.com/gopasspw/gopass/tree/master/cmd/gopass-summon-provider) tool, that can be used as [summon provider](https://cyberark.github.io/summon/#providers).
```
AnomalRoil, line 10: Please use the [github relative links](https://github.blog/2013-01-31-relative-links-in-markup-files/) here.

AnomalRoil (changes requested):
Also, if you are using Summon yourself, maybe you explain why it's interesting to use with Gopass? 
For example: will Summon erase the env var after usage or not?

dominikschulz (dismissed):
LGTM. Thanks.

But please use relative links.

marco-m:
Ah, it looks like that a force push dismisses reviews in this repo?
In any case, I updated the commit to use relative links.

> Also, if you are using Summon yourself, maybe you could explain why it's interesting to use with Gopass? I don't know anything about Summon myself, I think it could help users to have a clearer understanding of what is Summon and why it's nice to use with Gopass.

Maybe another time.

> For example: will Summon erase the env var after usage or not?

It would be impossible for summon to leave around the environment variables actually, due to how shells and subprocesses work.

In general, summon is well though-out from a security point of view.

dominikschulz (approved):
LGTM

Issue #1466, opened 2020-07-22
gopass installation fails via go get with golang 1.14.3 on POP OS 20.04 


wakatara:
### Summary

Fresh install on POPOS and go configured for environment with viable GOROOT and GOPATH. Install fails with missing src for MINIO package.

### Steps To Reproduce

Using 1.14.3 Go lang on POP OS 20.04 install, 

```
go get -u github.com/gopasspw/gopass
```

fails with:

```
package github.com/minio/minio-go/v6: cannot find package "github.com/minio/minio-go/v6" in any of:
	/usr/lib/go-1.14/src/github.com/minio/minio-go/v6 (from $GOROOT)
	/home/daryl/Code/go-nix/src/github.com/minio/minio-go/v6 (from $GOPATH)
```

I did try to install minio-go/V6 separately, but that also did not take. It appers that dependency has moved in the last month to a V7 so unsure if that's related. 

### Expected behavior

gopass should be installed.

### Environment

- OS: POP OS
- OS version: Linux elcano 5.4.0-7634-generic #38~1592497129~20.04~9a1ea2e-Ubuntu SMP Fri Jun 19 22:43:37 UTC  x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: 1.9.3
- Installation method: go get
- Go version 1.14.3 installed via apt

### Additional context

``` bash
λ  ~ echo $GOROOT
/usr/lib/go-1.14
λ  ~ echo $GOPATH
/home/daryl/.go-nix
```

dominikschulz:
Uhm, that's weird.

Can you please try with `GO111MODULE=on`?

wakatara:
Worked like a charm. Thank you! (I am embarassed I did not think of it.).

PR #1659, opened 2020-11-23
Update go get instructions in setup.md

kifirkin:
Apply changes from #1468 to setup.md


AnomalRoil:
That actually shouldn't be needed with Go 1.13+, since we have a go.mod file in the repo, it should automagically behave just like if `GO111MODULE=on `. 

Did you encounter an issue where using `GO111MODULE=on` explicitly helped with  Go 1.13 or more?

kifirkin:
Yes, I had the same issue as here https://github.com/gopasspw/gopass/issues/1466

dominikschulz (approved):
LGTM. Thanks.

PR #1661, opened 2020-12-04
fix syntax in example for gopass config

DO1JLR:
my shell tells me that this must be called ``gopass config autosync=true``.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -272,7 +272,7 @@ try to show desktop notifications [Linux only].
 If you want gopass to always push changes in git to your default remote server (origin), enable auto sync:
 
 ```bash
-$ gopass config autosync true
+$ gopass config autosync=true
 ```
 
 ### Check Passwords for Common Flaws


dominikschulz:
This depends on your shell (which one?).
For me the documented syntax seems to work.

DO1JLR:
I tested it with ``/bin/bash`` under Archlinux and ``cmd.exe`` in windows 10.

```bash
 $  gopass config autosync true

Error: Error setting config value
 $
 $
```
*(it's not working or what am I doing wrong?)*

```bash
 $  gopass config autosync=true
 $
 $
```
*(works now perfectly fine)*

----

Look at it via [asciinema](https://asciinema.org):
```
asciinema play https://github.com/gopasspw/gopass/files/5685762/gopasserror.cast.txt
```

[gopasserror.cast.txt](https://github.com/gopasspw/gopass/files/5685762/gopasserror.cast.txt)

AnomalRoil:
I just tried and it seems that on my side the syntax with `=` is non-working.

DO1JLR:
> I just tried and it seems that on my side the syntax with `=` is non-working.

Can you maybe give me a hint what I might be doing wrong?

I really made the pull request in the belief that I had found an inconsistency in the documentation by accident.

mniak (changes requested):
The syntax `gopass config autosync=false` is wrong. It does not print an error just because gopass interprets `autosync=false` as the key, and as the value is not set for this key, it does not print anything.

This  PR should be rejected.

DO1JLR:
ok, my bad. please excuse the disturbance. i just wanted to help!

PR #1665, opened 2020-12-12
Remove MIME default

### LINKED ISSUE #1660: Set the Mime option to `false` by default
Having the `mime` option set to `true` by default, and also for users that update from 1.9.2 caused havoc.

IMO, for 1.10.2, we need to:

- [ ] Set `mime` to `false` by default
- [ ] Make sure the migration path from 1.8.x or 1.9.x are working and that `mime` will be set to `false`
- [ ] Change the setting for existing users of 1.10.1 from `true` to `false` and tell in the changelog that users who want to keep using Mime format should revert the option back to `true`.
- [ ] Make it so that if Mime is set to false, `fsck` allows users to migrate all of their secrets away from Mime back to plain. This should: if there is a Mime header, move the value of the first `Password: ` key, if any, back to the top of the secret and remove the Mime header.

dominikschulz:
Work in progress, but happy to take feedback on the interface changes.

Should be mostly complete, now working on manual testing and cleanup.

Fixes #1660


internal/action/insert.go
```diff
@@ -89,20 +89,20 @@ func (s *Action) insert(ctx context.Context, c *cli.Context, name, key string, e
...
 	if appendTo && s.Store.Exists(ctx, name) {
 		eSec, err := s.Store.Get(ctx, name)
 		if err != nil {
 			return ExitError(ExitDecrypt, err, "failed to decrypt existing secret: %s", err)
 		}
-		sec = eSec.MIME()
+		sec = eSec // .MIME()
 		sec.Write(content)
 	} else {
 		plain, err := secparse.Parse(content)
```
AnomalRoil, line 101: I think we should stop parsing the input, or at least make it conditional.

AnomalRoil: Otherwise I could handle the "stop parsing the input" in another PR to have a cleaner history?

dominikschulz: I tend to agree. But I'd first like to get everything to work before making that call.

internal/secrets/yaml_test.go
```diff
@@ -28,13 +28,17 @@ key: value
...
+	assert.Equal(t, true, ok)
+	assert.Equal(t, "bar\nbaz\n", v)
 }
 
 func TestYAMLKeyFromEmptySecret(t *testing.T) {
 	t.Logf("Get Key from empty Secret")
 	s := &YAML{}
-	assert.Equal(t, "", s.Get(yamlKey))
+	v, ok := s.Get(yamlKey)
+	assert.Equal(t, false, ok)
```
AnomalRoil, line 40: assert.False

internal/secrets/yaml_test.go
```diff
@@ -28,13 +28,17 @@ key: value
 	s, err := ParseYAML([]byte(mlValue))
 	require.NoError(t, err)
 	assert.NotNil(t, s)
-	assert.Equal(t, "bar\nbaz\n", s.Get("foo"))
+	v, ok := s.Get("foo")
+	assert.Equal(t, true, ok)
```
AnomalRoil, line 32: assert.True

internal/secrets/yaml_test.go
```diff
@@ -65,7 +69,9 @@ func TestYAMLKeyToEmptySecret(t *testing.T) {
 	s.Set(yamlKey, yamlValue)
 
 	// read back key
-	assert.Equal(t, yamlValue, s.Get(yamlKey))
+	v, ok := s.Get(yamlKey)
+	assert.Equal(t, true, ok)
```
AnomalRoil, line 73: assert.True

internal/secrets/mime.go
```diff
@@ -0,0 +1,54 @@
...
+	"fmt"
+	"io"
+	"net/textproto"
+	"strings"
+
+	"github.com/gopasspw/gopass/pkg/gopass/secret"
+)
+
+// ParseLegacyMIME is a fallback parser for the transient MIME format
+func ParseLegacyMIME(buf []byte) (*KV, error) {
```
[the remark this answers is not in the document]
AnomalRoil: Also you might want to add a test using this, no?

tests/tester.go
```diff
@@ -23,7 +23,7 @@ const (
 autoimport: true
 cliptimeout: 45
 exportkeys: false
-mime: true
+mime: false
```
AnomalRoil, line 26: should that be removed instead?

tests/show_test.go
```diff
@@ -120,77 +116,32 @@ func TestShow(t *testing.T) {
...
 
 		_, err := ts.run("generate fo2 5")
 		assert.NoError(t, err)
 
-		out, err := ts.run("show fo2")
-		assert.Error(t, err)
-		assert.Contains(t, out, "Warning: safecontent=true")
+		//out, err := ts.run("show fo2")
+		//assert.Error(t, err)
+		//assert.Contains(t, out, "Warning: safecontent=true")
```
AnomalRoil, line 128: This should still be the case for Plain secrets, no?

tests/show_test.go
```diff
@@ -120,77 +116,32 @@ func TestShow(t *testing.T) {
...
-		_, err = ts.run("config mime true")
-		assert.NoError(t, err)
-
 		_, err = ts.run("generate fo6 5")
 		assert.NoError(t, err)
 
 		out, err = ts.run("show fo6")
 		assert.NoError(t, err)
-		assert.Contains(t, out, "Password: ***")
+		assert.Contains(t, out, "password: ***")
```
AnomalRoil, line 139: and this is the case only for KV secrets now, no?
Also we're missing a `assert.NotContains` the actual password, I guess.

tests/insert_test.go
```diff
@@ -26,25 +26,24 @@ func TestInsert(t *testing.T) {
...
 		assert.Equal(t, "moar", out)
 
 		out, err = ts.run("show -f some/newsecret")
 		assert.NoError(t, err)
-		assert.Equal(t, "and\n\nmoar", out)
+		assert.Equal(t, "and\nmoar", out)
 	})
 
 	t.Run("Regression test for #1595", func(t *testing.T) {
+		t.Skip("TODO")
```
AnomalRoil, line 45: This should pass, no? It's not dependent on Mime at all.

tests/insert_test.go
```diff
@@ -26,25 +26,24 @@ func TestInsert(t *testing.T) {
 	t.Run("Regression test for #1573 without actual pipes", func(t *testing.T) {
```
AnomalRoil, line 26: This is not really a good test for #1573 now that I think about it but I'm not sure how to test stuff that needs piping bytes into Gopass

tests/binary_test.go
```diff
@@ -28,7 +28,7 @@ func TestBinaryCopy(t *testing.T) {
 	})
 
 	fn := filepath.Join(ts.tempDir, "copy")
-	dat := []byte("foobar")
+	dat := []byte("foobar\n")
```
AnomalRoil, line 31: why the extra newline?

pkg/gopass/secret/ident.go
```diff
@@ -0,0 +1,6 @@
+package secret
+
+const (
+	// Ident is the Gopass MIME secret header
+	Ident = "GOPASS-SECRET-1.0"
```
AnomalRoil, line 5: we might want to mention it's deprecated? Also why not in parsesec with the Legacy function?

AnomalRoil (changes requested):

cmd/gopass-jsonapi/internal/jsonapi/api_test.go
```diff
@@ -177,16 +177,16 @@ sub:
 
 	runRespondMessage(t,
 		`{"type":"getData","entry":"foo"}`,
-		`{"hallo":"welt","password":"20"}`,
```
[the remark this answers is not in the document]
dominikschulz: I'm having a hard time making changes to the jsonapi code because it differs much, so I've filed #1673 to move it to a different repo. For now I tend to ignore this change.

internal/action/binary.go
```diff
@@ -72,15 +72,21 @@ func secFromBytes(dst, src string, in []byte) gopass.Secret {
 
 	debug.Log("Read %d bytes of %s from %s to %s", len(in), ct, src, dst)
 
-	sec := secret.New()
-	sec.Set("Content-Type", http.DetectContentType(in))
-	sec.Set("Content-Disposition", fmt.Sprintf("attachment; filename=\"%s\"", filepath.Base(src)))
+	sec := secrets.NewKV()
+	if err := sec.Set("Content-Type", http.DetectContentType(in)); err != nil {
```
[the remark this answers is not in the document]
dominikschulz: Indeed. Thanks.

internal/action/binary.go
```diff
@@ -72,15 +72,21 @@ func secFromBytes(dst, src string, in []byte) gopass.Secret {
...
+		debug.Log("Failed to set Content-Type: %q", err)
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/action/binary_test.go
```diff
@@ -117,7 +117,7 @@ func TestBinaryCopy(t *testing.T) {
 		defer buf.Reset()
 
 		infile := filepath.Join(u.Dir, "input.txt")
-		assert.NoError(t, ioutil.WriteFile(infile, []byte("0xDEADBEEF"), 0644))
+		assert.NoError(t, ioutil.WriteFile(infile, []byte("0xDEADBEEF\n"), 0644))
```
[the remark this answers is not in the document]
dominikschulz: Yes, possibly. I've been thinking about this as well.
The final newline is due to `internal/secrets/kv.go` line 125.
But we are expecting secrets to always contain text, don't we?
And having a text file being terminated by a newline is good practice, isn't it?

internal/action/generate.go
```diff
@@ -261,7 +261,7 @@ func (s *Action) generateSetPassword(ctx context.Context, name, key, password st
 		if err != nil {
 			return ctx, ExitError(ExitEncrypt, err, "failed to set key '%s' of '%s': %s", key, name, err)
 		}
-		sec := gs.MIME()
+		sec := gs
```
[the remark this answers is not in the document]
dominikschulz: Yes, this was left over from a refactoring.

internal/action/insert.go
```diff
@@ -89,20 +90,28 @@ func (s *Action) insert(ctx context.Context, c *cli.Context, name, key string, e
...
 			return ExitError(ExitDecrypt, err, "failed to decrypt existing secret: %s", err)
 		}
-		sec = eSec.MIME()
-		sec.Write(content)
+		secW, ok := eSec.(io.Writer)
+		if !ok {
+			return fmt.Errorf("%T is not an io.Writer", eSec)
+		}
+		if _, err := secW.Write(content); err != nil {
+			return err // TODO wrap
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/action/insert_test.go
```diff
@@ -69,7 +67,7 @@ func TestInsert(t *testing.T) {
 		buf.Reset()
 
 		assert.NoError(t, act.show(ctx, gptest.CliCtx(ctx, t), "baz", false))
-		assert.Equal(t, "Content-Type: text/yaml\nPassword: foobar\n\nother: meh\nuser: name\n", buf.String())
+		assert.Equal(t, "foobar\n\n---\nother: meh\nuser: name\n", buf.String())
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/otp/otp.go
```diff
@@ -13,7 +13,7 @@ import (
 func Calculate(name string, sec gopass.Secret) (twofactor.OTP, string, error) {
 	otpURL := ""
 	// check body
-	for _, line := range strings.Split(sec.GetBody(), "\n") {
+	for _, line := range strings.Split(sec.Body(), "\n") {
 		if strings.HasPrefix(line, "otpauth://") {
```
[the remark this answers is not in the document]
dominikschulz: Some people seem to have a non-KV secret with a `otpauth://` URL.
I admit we might decide not to support that use case and ask them to use a KV secret with a `otpauth` key. But not in this PR.

internal/secrets/kv.go
```diff
@@ -141,8 +142,14 @@ func ParseKV(in []byte) (*KV, error) {
...
+		debug.Log("no KV entries")
+		//return nil, fmt.Errorf("no KV entries")
 	}
 	k.body = sb.String()
 	return k, nil
 }
+
+func (k *KV) Write(buf []byte) (int, error) {
+	k.body += string(buf)
+	return 0, nil
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/secrets/kv_test.go
```diff
@@ -19,7 +19,9 @@ password: bar
 	require.NoError(t, err)
 	assert.NotNil(t, s)
 
-	assert.Equal(t, "", s.Get("Test / test.com"))
+	v, found := s.Get("Test / test.com")
+	assert.Equal(t, false, found)
```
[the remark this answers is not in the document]
dominikschulz: Done. Thank you.

internal/secrets/kv_test.go
```diff
@@ -39,26 +41,33 @@ Test / test.com
...
 	t.Run("read some keys", func(t *testing.T) {
 		for k, v := range map[string]string{
-			"password": "somepasswd",
+			"password": "bar",
 			"url":      "http://www.test.com/",
 			"username": "myuser@test.com",
 		} {
-			assert.Equal(t, v, s.Get(k))
+			fv, found := s.Get(k)
+			assert.Equal(t, true, found)
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/secrets/kv_test.go
```diff
@@ -39,26 +41,33 @@ Test / test.com
...
+			assert.Equal(t, v, fv)
 		}
+		assert.Equal(t, "somepasswd", s.Password())
 	})
 
 	t.Run("remove a key", func(t *testing.T) {
 		s.Set("foobar", "baz")
-		assert.Equal(t, "baz", s.Get("foobar"))
+		v, ok := s.Get("foobar")
+		assert.Equal(t, true, ok)
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/secrets/kv_test.go
```diff
@@ -39,26 +41,33 @@ Test / test.com
...
+		assert.Equal(t, "baz", v)
 		s.Del("foobar")
-		assert.Equal(t, "", s.Get("foobar"))
+		v, ok = s.Get("foobar")
+		assert.Equal(t, false, ok)
```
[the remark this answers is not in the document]
dominikschulz: Done.

internal/secrets/kv_test.go
```diff
@@ -39,26 +41,33 @@ Test / test.com
...
+		assert.Equal(t, "", v)
 	})
 
 	t.Run("read the body", func(t *testing.T) {
 		body := "Test / test.com\n"
-		assert.Equal(t, body, s.GetBody())
-		assert.Equal(t, body, s.GetBody())
-		assert.Equal(t, body, s.GetBody())
+		assert.Equal(t, body, s.Body())
```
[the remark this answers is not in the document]
dominikschulz: I don't know either.

internal/secrets/mime.go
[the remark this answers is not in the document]
dominikschulz: Yes, but this needs to wait.

We still need to figure out which secret(s) to expose to our API users.
Right now there is none, i.e. `pkg/gopass` isn't fully useable.
Once we figure out how to do that I want to touch this piece again.

internal/secrets/plain.go
```diff
@@ -50,23 +48,27 @@ func (p *Plain) Keys() []string {
 }
 
 // Get returns the first line (for password) or the empty string
-func (p *Plain) Get(key string) string {
-	if strings.ToLower(key) != "password" {
-		return ""
-	}
-	br := bufio.NewReader(bytes.NewReader(p.Body))
+func (p *Plain) Get(_ string) (string, bool) {
+	return "", false
```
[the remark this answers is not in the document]
dominikschulz: I don't think we need to warn since it doesn't exist (by definition) and the get a `found` `false` returned. But logging SGTM.

internal/secrets/yaml_test.go
dominikschulz: Done.

internal/secrets/yaml_test.go
dominikschulz: Done.

internal/secrets/yaml_test.go
dominikschulz: Done.

tests/tester.go
dominikschulz: Done.

tests/show_test.go
dominikschulz: Somehow this seems to be broken and I couldn't figure out what's wrong.
Maybe the regression tests need a little more context to make them easier to understand.

tests/show_test.go
dominikschulz: What do you mean?

tests/insert_test.go
dominikschulz: Hmm, it did fail and I didn't have enough context to figure out if this was possibly tied to MIME or not.

tests/insert_test.go
dominikschulz: Good question, I don't know.

tests/binary_test.go
dominikschulz: Done.

pkg/gopass/secret/ident.go
dominikschulz: The secrets implementation needs to change once we figure out what to expose to the Gopass API users. I plan to clean up the remaining pieces by then.

tests/show_test.go
AnomalRoil: I mean that this test is testing that the `safecontent` option is correctly obstructing the output. And my point was that for plain secrets we should be doing any parsing, so it should only do so for KV values and instead for plain secrets the safecontent is just removing the first line, no?

Or do we replace the first line by `"password: ***"` in the case of plain secrets?

tests/insert_test.go
AnomalRoil: I can take a look as some of the tests in a follow up PR if you want.

AnomalRoil (approved):
LGTM.

The migration path from Mime back to plain secret is simply doing a `gopass fsck --decrypt` ?
We might want to document it and add it to the changelog.

dominikschulz:
The migration path should be fairly transparent, i.e. we should automatically parse MIME secrets and convert to KV. The decrypt flag to fack was turned into a no op. I was afraid of accidentally shredding secrets. they will be rewritten on changes or have to be manually updated. But that should be much safer.

tests/insert_test.go
dominikschulz: Yes, please. Thank you!

tests/show_test.go
dominikschulz: Ah, good point.

For plain secrets we assume that the first line contains the secret, so we could support that there as well.

AnomalRoil:
It's a bit sad to have `--decrypt` being a no-op, it's useful when a recipient operation fails in the middle or for such cases.

In any case then the specification for `fsck` is wrong now, isn't it? https://github.com/gopasspw/gopass/blob/master/docs/commands/fsck.md

dominikschulz:
I think we still might want to bring it back to comply with the spec.

PR #1668, opened 2020-12-23
Fix: Duplicate example

ssiyad:
Both commands shown in example are same so one is changed.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -203,7 +203,7 @@ $ gopass show golang.org/gopher
 Eech4ahRoy2oowi0ohl
 ```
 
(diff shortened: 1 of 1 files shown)


AnomalRoil:
Actually that's because the invocation of `gopass entryname` is somewhat deprecated IIRC, because now the `gopass` command opens an interactive prompt. 
Or is it still the default @dominikschulz ?

If you ask me, `gopass` alone can open an interactive prompt, but it'd be cleaner to have `gopass nameofentry` default to show, with all its flags to avoid having global flags and special cases to handle it.

dominikschulz:
@AnomalRoil  Yes, I was considering to deprecate `gopass entry` in favor of `gopass show entry` but I'm not so sure anymore.
Not sure I can follow your suggestion for `gopass entry` to default to `show`. Isn't that what we do right now?
If not please file a separate issue for that.

@ssiyad Thank you. LGTM.

PR #1677, opened 2021-01-01
Reduce usage of internal packages in non-core binaries

dominikschulz:
This commit reduces the usage of internal/ packages in auxiliary
binaries so we can get closer to eventually moving them to their own
repos.


AnomalRoil (dismissed):
LGTM

martinhoefling (dismissed):

martinhoefling:
How do we replace `out.WithHidden` ? E.g. https://github.com/gopasspw/gopass/blob/983076141df8d00fa8c514642dc4a60c9be1b385/cmd/gopass-jsonapi/internal/jsonapi/api.go#L22

dominikschulz:
@martinhoefling I haven't figured out all cases, yet. Let's move step by step. Need more time to refine our public API.

Issue #1676, opened 2021-01-01
Implement gopass self-update


dominikschulz:
`gopass update` doesn't verify the integrity of downloaded binaries.

If we want to keep this feature we need to implement proper update signing and verification.

PR #1679, opened 2021-01-02
Disable gopass update

### LINKED ISSUE #1675: Disable self updater
`gopass update` doesn't verify integrity of the downloaded binaries.

This is dangerous and must be disabled until we can implement a proper self updater (like e.g. `restic self-update`).

dominikschulz:
The updater is insecure since it doesn't check signature of downloaded
binaries (which aren't signed at the moment). It shouldn't be used this
way so the quickest fix is to disable it for the time being.

Fixes #1675 

### DIFF
--- a/internal/updater/update.go
+++ b/internal/updater/update.go
@@ -168,6 +168,10 @@ func updateCheckHost(u *url.URL) error {
 func updateTo(ctx context.Context, version, url string) error {
 	debug.Log("URL: %s", url)
 	out.Green(ctx, "Update available!")
+	if true {
+		out.Red(ctx, "Gopass Updater is disabled (insecure, see #1676")
+		return nil
+	}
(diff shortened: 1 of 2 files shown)


internal/updater/update.go
```diff
@@ -168,6 +168,10 @@ func updateCheckHost(u *url.URL) error {
 func updateTo(ctx context.Context, version, url string) error {
 	debug.Log("URL: %s", url)
 	out.Green(ctx, "Update available!")
+	if true {
+		out.Red(ctx, "Gopass Updater is disabled (insecure)")
```
AnomalRoil, line 172: Maybe add a reference to #1676 ?

AnomalRoil (approved):

internal/updater/update.go
dominikschulz: Done

Issue #1673, opened 2020-12-31
gopass-jsonapi: Move to separate repo


dominikschulz:
The implementation of the `gopass-jsonapi` differs a lot from the style and conventions used in other parts of `gopass`.
This frequently slows down changes to `gopass` itself.

If possible (wrt. visibility and such) we should move `gopass-jsonapi` to it's own repo, e.g. github.com/gopasspw/gopass-jsonapi

martinhoefling:
Makes sense, I think this would also allow us to define a clean API that can be used from all other tools as well.

martinhoefling:
I will create a new repo but gopass-jsonapi needs some internals atm. How do we proceed?
````
go build
package github.com/gopasspw/gopass-jsonapi
        jsonapi.go:11:2: use of internal package github.com/gopasspw/gopass/cmd/gopass-jsonapi/internal/jsonapi not allowed  
package github.com/gopasspw/gopass-jsonapi                                                                                         
        jsonapi.go:12:2: use of internal package github.com/gopasspw/gopass/cmd/gopass-jsonapi/internal/jsonapi/manifest not 
allowed                                                                                                                      allowed
package github.com/gopasspw/gopass-jsonapi
        jsonapi_windows.go:13:2: use of internal package github.com/gopasspw/gopass/internal/out not allowed
package github.com/gopasspw/gopass-jsonapi
        jsonapi.go:13:2: use of internal package github.com/gopasspw/gopass/internal/termio not allowed
package github.com/gopasspw/gopass-jsonapi
        imports github.com/gopasspw/gopass-jsonapi/internal/jsonapi
        internal\jsonapi\responses.go:11:2: use of internal package github.com/gopasspw/gopass/internal/clipboard not allowedpackage github.com/gopasspw/gopass-jsonapi
        imports github.com/gopasspw/gopass-jsonapi/internal/jsonapi
        internal\jsonapi\responses.go:12:2: use of internal package github.com/gopasspw/gopass/internal/otp not allowed      
package github.com/gopasspw/gopass-jsonapi
        jsonapi_windows.go:13:2: use of internal package github.com/gopasspw/gopass/internal/out not allowed
package github.com/gopasspw/gopass-jsonapi
        jsonapi.go:13:2: use of internal package github.com/gopasspw/gopass/internal/termio not allowed
PS C:\Users\Martin\workspace\gopasspw\gopass-jsonapi> go build
package github.com/gopasspw/gopass-jsonapi
[...]

PR #1680, opened 2021-01-02
Fixing Windows tests

AnomalRoil:
RELEASE_NOTES=[ENHANCEMENT] Full windows support

It'd be nice to finally have full windows support and I thing I'm pretty close to have achieved it.


AnomalRoil:
Windows tests are no longer allowed to fail now since they all pass.

Notice the exception for jsonapi (in the Makefile) which is considered a malware for some reason on Windows and is therefore excluded on windows.
We will be able to remove the exception once jsonapi will be moved to its own repo in #1673 .

dominikschulz (approved):
LGTM. Thanks!

PR #1684, opened 2021-01-04
Remove standards violating GPG binary location caching

### LINKED ISSUE #1662: ~/.cache/gopass/gpg-binary.loc is storing an absolute path to gpg, which overrides the environment $PATH
### Summary

`~/.cache/gopass/gpg-binary.loc` is storing an absolute path to the `gpg` executable, which overrides the environment `$PATH`.

### Steps To Reproduce

Config file:

```yaml
---
autoclip: true
autoimport: true
cliptimeout: 45
exportkeys: true
mime: true
nocolor: false
nopager: false
notifications: true
path: /Users/[my_user_name]/.password-store
safecontent: true
mounts: {}
```

```bash
export PATH=/unusual/path/to/gnupg/bin
gopass  my-password  # this works
mv /unusual/path/to/gnupg/bin /unusual/path2/to/gnupg/bin
export PATH=/unusual/path2/to/gnupg/bin
gopass  my-password  # breaks with 'Decryption failed: fork/exec /unusual/path/to/gnupg/bin/gpg'
rm ~/.cache/gopass/gpg-binary.loc
gopass  my-password  # this works again
```

### Expected behavior

Any UNIX program should *always* respect the contents of `$PATH`

### Environment

- OS: Mac OS X Catalina
- OS version: `Darwin computer.home 19.6.0 Darwin Kernel Version 19.6.0: Thu Oct 29 22:56:45 PDT 2020; root:xnu-6153.141.2.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro16,1 Darwin`
- gopass Version: `gopass 1.10.1+af9c95d20b722dee331adcc3cc1ab1ce0108dc2b (af9c95d20b722dee331adcc3cc1ab1ce0108dc2b) go1.15.6 darwin amd64`
- Installation method: [Nix](https://nixos.org/), building the latest build from source

dominikschulz:
The caching is flawed (i.e. it doesn't have any kind of expiration or fallback) and violations assumptions (i.e. should respect `$PATH`). Also the code is quite complex for little benefit.

We will expect to find gpg as `gpg` and if that doesn't work users will have to override `GOPASS_GPG_BINARY` to match their environment.

Please note that I'm keeping the Windows implementaion as-is because I'm less familiar with how this should behave on that platform.

Fixes #1662

RELEASE_NOTES=[BUGFIX] Remove GPG location caching


siriobalmelli:
This approach seems logical

internal/backend/crypto/gpg/cli/binary_others.go
```diff
@@ -0,0 +1,14 @@
...
+
+package cli
+
+import (
+	"os/exec"
+)
+
+func detectBinary(name string) (string, error) {
+	if name == "" {
+		name = "gpg"
```
AnomalRoil, line 11: how about `gpg2` when `gpg` is not found?

internal/backend/crypto/gpg/cli/loader.go
```diff
@@ -23,8 +24,9 @@ type loader struct{}
 func (l loader) New(ctx context.Context) (backend.Crypto, error) {
 	debug.Log("Using Crypto Backend: %s", name)
 	return New(ctx, Config{
-		Umask: fsutil.Umask(),
-		Args:  GPGOpts(),
+		Umask:  fsutil.Umask(),
+		Args:   GPGOpts(),
+		Binary: os.Getenv("GOPASS_GPG_BINARY"),
```
AnomalRoil, line 29: Where is that used? Does it really work??

internal/backend/crypto/gpg/cli/binary_others.go
AnomalRoil: And shouldn't this `GOPASS_GPG_BINARY` be used here in case it is set??

internal/backend/crypto/gpg/cli/binary_windows.go
```diff
@@ -3,14 +3,44 @@
...
 	"path/filepath"
+	"sort"
 
 	"golang.org/x/sys/windows/registry"
 
+	"github.com/gopasspw/gopass/internal/debug"
 	"github.com/gopasspw/gopass/pkg/fsutil"
 )
 
+func detectBinary(bin string) (string, error) {
```
AnomalRoil, line 18: is this using the `GOPASS_GPG_BINARY` when it's set?

AnomalRoil (commented):
This seems fine to me, but I'm not sure I have all the background to follow what's going on here.

internal/backend/crypto/gpg/cli/binary_others.go
dominikschulz: This is what led to this convoluted implementation in the first place.
Are there many distros still shipping gpg 1.x (as `gpg`) and gpg 2.x (as `gpg2`)?

internal/backend/crypto/gpg/cli/loader.go
dominikschulz: The loader? Yes, that's being invoked through the blank imports.

internal/backend/crypto/gpg/cli/binary_windows.go
dominikschulz: Yes, through the loader.

PR #1692, opened 2021-01-07
Update to minio/v7

### LINKED ISSUE #1691: Minio needs to be updated to v7
We are currently relying on Minio v6, it'd be best to update to the latest v7 released in July 2020 that underwent a full API refactor if we want to be future proof.

Or maybe drop it to avoid the added complexity?
Who is using S3 buckets and these remote stuff on Gopass instead of the classical git workflow?

dominikschulz:
Maybe later we want to remove this feature, but for the next release I haven't made up my mind, yet.

Fixes #1691

RELEASE_NOTES=[CLEANUP] Update to minio/v7


AnomalRoil (approved):
LGTM, but I have no idea how to test if it works "as usual" or not, since I've never worked with S3 buckets directly on my side.

dominikschulz:
Yes, this is a very experimental feature that intent to move out of the main branch.
But I don't have the time to to that right now and don't want to cause collateral damage when rushing things.

PR #1695, opened 2021-01-11
Adding Mime warning to fsck

AnomalRoil:
RELEASE_NOTES=[ENHANCEMENT] fsck now detects leftover Mime secrets
RELEASE_NOTES=[BUGFIX] fsck won't correct recipients without --decrypt

### DIFF
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -2,6 +2,7 @@ package leaf
 
 import (
 	"context"
+	"github.com/gopasspw/gopass/internal/secrets"
 	"sort"
 	"strings"
 
@@ -18,11 +19,13 @@ func (s *Store) Fsck(ctx context.Context, path string) error {
 	debug.Log("Checking %s", path)
 
 	// first let the storage backend check itself
+	out.Print(ctx, "Checking storage backend")
 	if err := s.storage.Fsck(ctx); err != nil {
 		return errors.Wrapf(err, "storage backend found errors: %s", err)
 	}
 
 	// then try to compact storage / rcs
+	out.Print(ctx, "Compacting storage if possible")
 	if err := s.storage.Compact(ctx); err != nil {
 		return errors.Wrapf(err, "storage backend compaction failed: %s", err)
 	}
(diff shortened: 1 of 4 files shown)


internal/store/leaf/fsck.go
```diff
@@ -59,10 +62,14 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) error {
 	// make sure we can actually decode this secret
 	// if this fails there is no way we could fix this
 	if IsFsckDecrypt(ctx) {
-		_, err := s.Get(ctx, name)
+		ctx = ctxutil.WithShowParsing(ctx, true)
+		secret, err := s.Get(ctx, name)
 		if err != nil {
 			return errors.Wrapf(err, "failed to decode secret %s: %s", name, err)
 		}
+		if secret.Type() == "Mime" {
```
dominikschulz, line 70: This should be: `if kv, ok := secret.(secrets.KV); ok && kv.FromMime() {`

See my other comment.

dominikschulz (changes requested):
I like the feature, but please change the implementation.
What you're doing looks like poor-mans type checking.
Go has built-in support for that. That would be cleaner and also more concise.

Please:
- keep fsck.go mostly as-is (except for my comment there)
- change kv.go for have a func FromMime() bool
- Remove all other changes

AnomalRoil:
I should have thought of using type assertions, sorry for that.
Done.

dominikschulz (approved):
LGTM

internal/store/leaf/fsck.go
```diff
@@ -59,10 +63,15 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) error {
...
 	// if this fails there is no way we could fix this
 	if IsFsckDecrypt(ctx) {
-		_, err := s.Get(ctx, name)
+		// we need to make sure Parsing is enabled in order to parse old Mime secrets
+		ctx = ctxutil.WithShowParsing(ctx, true)
+		secret, err := s.Get(ctx, name)
 		if err != nil {
 			return errors.Wrapf(err, "failed to decode secret %s: %s", name, err)
 		}
+		if kv, ok := secret.(*secrets.KV); ok && kv.FromMime() {
```
dominikschulz, line 72: Actually we could have type checked against a locally defined interface `type fromMime interface { FromMime() bool }` here, but let's leave it as is for now.

PR #1697, opened 2021-01-12
Add Pinentry CLI fallback

### LINKED ISSUE #1655: gopass init --crypto age fails if pgp tools are not installed.
### Summary

gopass init --crypto age fails if pgp tools such as pinentry-mac  are not installed.

I am using age because for end users its so much easier for them, in that pinentry-mac is not needed.
How about 

### Steps To Reproduce

```
gopass init --crypto age
[init] Initializing a new password store ...

Error: failed to initialize store: failed to read user input: pinentry Error: exec: "pinentry-mac": executable file not found in $PATH
```

### Expected behaviour

Should not require the user to have pinentry-mac, since they are not using pgp as the basis of their crypto.

### Environment

- OS: Mac OS X  Big sur

- OS version: 
uname -a
Darwin x-MacBook-Pro.local 20.1.0 Darwin Kernel Version 20.1.0: Sat Oct 31 00:07:11 PDT 2020; root:xnu-7195.50.7~2/RELEASE_X86_64 x86_64

- gopass Version: 
gopass --version
gopass 1.10.1-git+HEAD go1.15.5 darwin amd64
- Installation method: source

### Additional context

Am integrating gopass with sops and age.

dominikschulz:
Fixes #1655

RELEASE_NOTES=[ENHANCEMENT] Add Pinentry CLI fallback

### DIFF
--- a/internal/backend/crypto/age/askpass.go
+++ b/internal/backend/crypto/age/askpass.go
@@ -8,6 +8,7 @@ import (
 	"github.com/gopasspw/gopass/internal/cache"
 	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/pinentry"
+	"github.com/gopasspw/gopass/pkg/pinentry/cli"
 )
 
 type piner interface {
@@ -38,7 +39,12 @@ func newAskPass() *askPass {
 	return &askPass{
 		cache: cache.NewInMemTTL(time.Hour, 24*time.Hour),
 		pinentry: func() (piner, error) {
-			return pinentry.New()
+			p, err := pinentry.New()
+			if err == nil {
+				return p, nil
+			}
+			debug.Log("Pinentry not found: %q", err)
+			return cli.New()
 		},
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/backend/crypto/age/askpass.go
```diff
@@ -38,7 +39,12 @@ func newAskPass() *askPass {
 	return &askPass{
 		cache: cache.NewInMemTTL(time.Hour, 24*time.Hour),
 		pinentry: func() (piner, error) {
-			return pinentry.New()
+			p, err := pinentry.New()
```
AnomalRoil, line 42: Shouldn't this be provided by the cli itself through a function "TryPinEntry" or so?
It would render the cli package a bit more useful.

AnomalRoil (approved):

internal/backend/crypto/age/askpass.go
dominikschulz: That would require more changes than I'd like to do for now (introducing some interfaces and such).

AnomalRoil (approved):

PR #1706, opened 2021-01-12
Add template functions for Bcrypt and Argon2

### LINKED ISSUE #1689: Templates should support modern password hashing functions
We should consider adding support for more modern password hashing functions like Argon2 or Bcrypt, too.

https://godoc.org/golang.org/x/crypto/argon2
https://godoc.org/golang.org/x/crypto/bcrypt
https://doc.dovecot.org/configuration_manual/authentication/password_schemes/

dominikschulz:
Fixes #1689

RELEASE_NOTES=[ENHANCEMENT] Add tpl funcs for Bcrypt and Argon2

### DIFF
--- a/internal/pwschemes/argon2i/argon2i.go
+++ b/internal/pwschemes/argon2i/argon2i.go
@@ -1,0 +1,32 @@
+package argon2i
+
+import (
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+
+	"golang.org/x/crypto/argon2"
+)
+
+const (
+	time    uint32 = 4
+	memory  uint32 = 256 * 1024
+	threads uint8  = 4
+	keylen  uint32 = 32
+)
+
+// Generate generates a new Argon2I hash with recommended values for it's
+// complexity parameters.
+func Generate(password string, saltLen uint8) (string, error) {
+	salt := make([]byte, saltLen)
+	_, err := rand.Read(salt)
+	if err != nil {
+		return "", err
+	}
+	hash := argon2.Key([]byte(password), salt, time, memory, threads, keylen)
+
+	hstr := base64.RawStdEncoding.EncodeToString(hash)
+	sstr := base64.RawStdEncoding.EncodeToString(salt)
+
+	return fmt.Sprintf("{ARGON2I}$argon2i$v=%d$m=%d,t=%d,p=%d$%s$%s", argon2.Version, memory, time, threads, sstr, hstr), nil
+}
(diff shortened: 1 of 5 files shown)


dominikschulz:
This PR doesn't come with tests as I deliberately didn't want to implement validation for these schemes, but I've did end to end tests with Dovecot and that works.

internal/pwschemes/argon2i/argon2i.go
```diff
@@ -0,0 +1,32 @@
...
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+
+	"golang.org/x/crypto/argon2"
+)
+
+const (
+	time    uint32 = 4
+	memory  uint32 = 32 * 1024
```
AnomalRoil, line 13: As per section 8 and 9 of the [Argon2 paper](https://github.com/P-H-C/phc-winner-argon2/blob/master/argon2-specs.pdf), I think we might want to increase this, unless it's the settings used by Dovecot?

internal/pwschemes/argon2id/argon2id.go
```diff
@@ -0,0 +1,32 @@
...
+	"crypto/rand"
+	"encoding/base64"
+	"fmt"
+
+	"golang.org/x/crypto/argon2"
+)
+
+const (
+	time    uint32 = 3
+	memory  uint32 = 64 * 1024
```
AnomalRoil, line 13: See previous comment. 64MB is low compared to recommended values.

internal/tpl/funcs.go
```diff
@@ -27,6 +30,9 @@ const (
 	FuncGet         = "get"
 	FuncGetPassword = "getpw"
 	FuncGetValue    = "getval"
+	FuncArgon2i     = "argon2i"
```
AnomalRoil, line 33: Please document these new functions in the Template.md documentation.

AnomalRoil (changes requested):

internal/pwschemes/argon2i/argon2i.go
dominikschulz: No, these values are encoded into the hash and Dovecot will re-use whatever we use here. I haven't read the whole paper so I'm not sure if I understand those correctly, yet. But I'll start with a little more and we can change it or make it configurable later.

internal/pwschemes/argon2id/argon2id.go
dominikschulz: Done

internal/tpl/funcs.go
dominikschulz: Done.

PR #1710, opened 2021-01-14
Add optional pinentry unescaping

### LINKED ISSUE #1621: gopass/pkg/pinentry: pinentry escapes '%' as '%25' and gopass does not unescape this
### Summary
I've been a big fan of this project for quite a while and stumbled upon a subtle issue while using another project.

I've recently switched my Yubikeys and used a PIN with special characters and ran into an issue while setting up https://github.com/FiloSottile/yubikey-agent. yubikey-agent internally uses `github.com/gopasspw/gopass/pkg/pinentry` which in turn is only used in the `age` backend within gopass itself. Because gopass always uses this implementation this issue will not result in any visible bugs.

pinentry escapes `%` signs (next to `\r` and `\n`, but these two probably never occur in a PIN) and `gopass/pkg/pinentry` does not unescape the received data. This breaks PINs that contain `%` as a character (supported by e.g. yubikeys). 

There might be some hint about this in the protocol description. http://info2html.sourceforge.net/cgi-bin/info2html-demo/info2html?(pinentry)Protocol

>  To include linefeeds or other special characters, you may percent-escape them
> (i.e. a line feed is encoded as `%0A', the percent sign itself is
> encoded as `%25').

The actual escaping comes from  [libassuan](https://gnupg.org/software/libassuan/index.html) though. libassuan provides IPC communication between the various gnupg binaries.

gnupg uses this library to communicate with pinentry instead of using raw pipes: Specifically, the getpin request is done using [do_getpin](https://github.com/gpg/gnupg/blob/da3a4c54a8ce8a7dc442c70bda9b7eda22d43e57/agent/call-pinentry.c#L1059) -> [assuan_transact](https://github.com/gpg/libassuan/blob/b270f2ec21b67f5728edae4b2ddf6fe17a0fd25a/src/client.c#L219) -> [_assuan_read_from_server](https://github.com/gpg/libassuan/blob/b270f2ec21b67f5728edae4b2ddf6fe17a0fd25a/src/client.c#L198) -> [assuan_client_read_response](https://github.com/gpg/libassuan/blob/b270f2ec21b67f5728edae4b2ddf6fe17a0fd25a/src/client.c#L71) which then unescapes the received line (see https://github.com/gpg/libassuan/blob/b270f2ec21b67f5728edae4b2ddf6fe17a0fd25a/src/client.c#L97).

### Steps To Reproduce

The escape behaviour can be reproduced by running pinentry and using the `GETPIN` command. Then enter `%` as the pin:
```
$ pinentry
OK Pleased to meet you
GETPIN
D %25
OK
```

This can be reproduced on both Mac OS (using /usr/local/MacGPG2/libexec/pinentry-mac.app/Contents/MacOS/pinentry-mac and also pinentry-curses) and Linux (pinentry) and on all pinentry programs that use libassuan. 

This does however not cause any issues within gopass itself: the age backend will happily use `%25` as the passphrase for encryption and decryption.

### Expected behavior
The `%` sign is correctly unescaped.

### Environment

- OS: doesn't matter, but I've tried this on Mac OS and Linux
- OS version: doesn't matter, but Darwin [..] 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 18 20:50:10 PDT 2020; root:xnu-4903.278.43~1/RELEASE_X86_64 x86_64 and Linux [...] 5.7.19-hardened #1-NixOS SMP Thu Aug 27 07:30:50 UTC 2020 x86_64 GNU/Linux
- gopass Version: git head
- Installation method: from source

### Additional context

At first glance there is a quick way to fix this by adding the unescaping to https://github.com/gopasspw/gopass/blob/11660a16cf76ddef57134163a7c4b0d19ab0fb7d/pkg/pinentry/pinentry.go#L114

I believe this should be sufficient since the unescaping is [only done when the reply starts with "D "](https://github.com/gpg/libassuan/blob/b270f2ec21b67f5728edae4b2ddf6fe17a0fd25a/src/client.c#L99) and this is the only place in gopass that expects this reply from pinentry.

libassuan itself seems to be much more complex and I do not believe that a full go implementation is required for just using pinentry. I'll double check to see if there are any other subtle issues that may come up when directly talking to pinentry though.

There is one rather big problem with fixing this though: the age backend will currently use whatever it gets back from GetPIN. If the passphrase has a `%` sign this will just be replaced with `%25` which will work just fine. If this escaping is fixed though suddenly any passphrase that contained a `%` before the fix will no longer work.

I'm not really sure what the best way forward is. Fixing this will break compatibility and make old users with `%` in their passphrases rather unhappy. Not fixing this will not cause any issues for users who just stick to gopass but might cause strange bugs for other projects (or other places within gopass who later switch to pinentry) who use `gopass/pkg/pinentry`

dominikschulz:
Fixes #1621

RELEASE_NOTES=[ENHANCEMENT] Add optional pinentry unescaping

### DIFF
--- a/pkg/pinentry/pinentry.go
+++ b/pkg/pinentry/pinentry.go
@@ -7,11 +7,25 @@ import (
 	"bytes"
 	"fmt"
 	"io"
+	"net/url"
+	"os"
 	"os/exec"
 	"strings"
 
 	"github.com/pkg/errors"
 )
+
+var (
+	// Unescape enables unescaping of percent encoded values,
+	// disabled by default for backward compatibility. See #1621
+	Unescape bool
+)
+
+func init() {
+	if sv := os.Getenv("GOPASS_PINENTRY_UNESCAPE"); sv == "on" || sv == "true" {
+		Unescape = true
+	}
+}
 
 // Client is a pinentry client
 type Client struct {
@@ -134,5 +148,15 @@ func (c *Client) GetPin() ([]byte, error) {
 	if !bytes.HasPrefix(ok, []byte("OK")) {
 		return nil, fmt.Errorf("unexpected response: %s", ok)
 	}
-	return pin[2:], nil
+	pin = pin[2:]
+
+	// Handle unescaping, if enabled
+	if bytes.Contains(pin, []byte("%")) && Unescape {
+		sv, err := url.QueryUnescape(string(pin))
+		if err != nil {
+			return nil, fmt.Errorf("failed to unescape pin: %q", err)
+		}
+		pin = []byte(sv)
+	}
+	return pin, nil
 }


pkg/pinentry/pinentry.go
```diff
@@ -7,12 +7,26 @@ import (
...
+	"os"
 	"os/exec"
 	"strings"
 
 	"github.com/pkg/errors"
 )
 
+var (
+	// Enable unescaping of percent encoded values, disabled by default
+	// for backwards compatability. See #1621
```
AnomalRoil, line 20: "backward compatibility"

pkg/pinentry/pinentry.go
```diff
@@ -134,5 +148,15 @@ func (c *Client) GetPin() ([]byte, error) {
 	if !bytes.HasPrefix(ok, []byte("OK")) {
 		return nil, fmt.Errorf("unexpected response: %s", ok)
 	}
-	return pin[2:], nil
+	pin = pin[2:]
+
+	// Handle unescaping, if enabled
+	if bytes.Contains(pin, []byte("%")) && Unescape {
+		sv, err := url.QueryUnescape(string(pin))
```
AnomalRoil, line 155: Ahahah, unexpected package to do so. But I guess it should work.

AnomalRoil (dismissed):
Were you able to test this on a Yubikey or do you want me to do so before merging?

pkg/pinentry/pinentry.go
```diff
@@ -7,12 +7,26 @@ import (
...
+	"net/url"
+	"os"
 	"os/exec"
 	"strings"
 
 	"github.com/pkg/errors"
 )
 
+var (
+	// Enable unescaping of percent encoded values, disabled by default
```
AnomalRoil, line 19: comment on exported var Unescape should be of the form "Unescape ..."

dominikschulz: Done

pkg/pinentry/pinentry.go
dominikschulz: Done

dominikschulz:
No, I haven't been able to test this with a YubiKey setup. Could you do that, please?

AnomalRoil:
I'll take a look this afternoon with a Yubikey.

AnomalRoil (approved):
Tested with a Yubikey and yubikey-agent, simply had to enable 
`pinentry.Unescape = true` in yubikey-agent and it worked.

But I had to update my systemd and ended up doing a full distro upgrade for this test, that was unexpected xD

PR #1711, opened 2021-01-14
Let summon provider take an optional key

raphink:


cmd/gopass-summon-provider/summon-provider.go
```diff
@@ -24,10 +25,24 @@ func (s *gc) Get(c *cli.Context) error {
 	ctx := ctxutil.WithGlobalFlags(c)
 	ctx = ctxutil.WithNoNetwork(ctx, true)
 	path := c.Args().Get(0)
+	key := c.Args().Get(1)
+	if key != "" {
+		ctx = action.WithKey(ctx, key)
```
dominikschulz, line 30: You don't need this.

cmd/gopass-summon-provider/summon-provider.go
```diff
@@ -24,10 +25,24 @@ func (s *gc) Get(c *cli.Context) error {
...
+	}
 	secret, err := s.gp.Get(ctx, path, "latest")
 	if err != nil {
 		return err
 	}
-	fmt.Fprintln(Stdout, secret.Password())
+
+	if key == "" {
```
dominikschulz, line 37: This is not how I'd do this.

What about:

```
if key := c.Args().Get(1); key != "" {
  if val, found := secret.Get(key); found {
   fmt.Fprintln(Stdout, val)
   return
  }
  fmt.Fprintln(Stdout, "Error")
}
```

dominikschulz (changes requested):
Needs some changes. Also please add a RELEASE_NOTES tag and sign off your commit.

cmd/gopass-summon-provider/summon-provider.go
raphink: That looks easier indeed, thanks.

raphink:
@dominikschulz I applied your recommendation.

That said, I wonder if there is a way that this provider could be backwards-compatible with the way we've used gopass with summon for years, which is:

```
#!/bin/sh
gopass show $(echo "${@}"|tr : \ )
```

When the password is simple, it's equivalent. However, `gopass show` displays the entire password even when there's multiple lines, and we've relied on this so far.

dominikschulz:
If you want to have the same behaviour as `gopass show` you'd want the full body.
That can be done using `string(secret.Bytes())`.

I think you have a small syntax error and are still missing a RELEASE_NOTES line as well as a Signed-Off line.

AnomalRoil:
I guess you should re-submit it to the new repo now ;)
https://github.com/gopasspw/gopass-summon-provider/pulls

PR #1713, opened 2021-01-15
Allow using tilde in config

### LINKED ISSUE #872: Use '~' in config.yml

### LINKED ISSUE #1624: gopass-generated config `path` does not accept a relative path
### Summary

A config `path` entry of `~/.password-store` will not parse, whereas an absolute path eg `/Users/sirio/.password-store` will work.

### Steps To Reproduce

1. Remove existing config:

        rm ~/.config/gopass/config.yml

2. Use `gopass` to generate a config:

        $ gopass config path "~/.password-store"
        path: ~/.password-store

    This will generate the following config:

        autoclip: false
        autoimport: true
        cliptimeout: 45
        exportkeys: true
        mime: true
        nocolor: false
        nopager: false
        notifications: true
        path: ~/.password-store
        safecontent: false
        mounts: {}

3. Generated config fails to parse

        $ gopass list
        It seems you are new to gopass. Do you want to run the onboarding wizard? [Y/n/q]:

4. Change relative path to absolute and it will now parse

        $ sed -i"" "s|~|$HOME|g" ~/.config/gopass/config.yml
        $ gopass list
        # ... list output ...

### Expected behavior

Relative paths in config should be correctly parsed.

It is important to support relative paths:

- user home directories don't have the same path between Darwin/Linux or between same-OS machines where the username is different.
- it should be possible to maintain `~/.config` as a Git repo sync'ed between multiple machines

### Environment

- OS: Darwin (macOS Catalina)

- OS version: `Darwin computer.home 19.6.0 Darwin Kernel Version 19.6.0: Mon Aug 31 22:12:52 PDT 2020; root:xnu-6153.141.2~1/RELEASE_X86_64 x86_64 i386 MacBookPro16,1 Darwin`

- gopass Version:
```
gopass 1.10.1+fc5e9b75254fd975ff458cf4843f2def21f872ee (fc5e9b75254fd975ff458cf4843f2def21f872ee) go1.15.2 darwin amd64
<root>     - gpg 2.2.23 - git 2.28.0
```

- Installation method: Nixpkgs (building latest commit from source)

dominikschulz:
Fixes #872
Fixes #1624 

RELEASE_NOTES=[BUGFIX] Allow using tilde

### DIFF
--- a/internal/store/root/init.go
+++ b/internal/store/root/init.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/internal/store/leaf"
 	"github.com/gopasspw/gopass/pkg/debug"
+	"github.com/gopasspw/gopass/pkg/fsutil"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


AnomalRoil (approved):
LGTM

PR #1715, opened 2021-01-16
UX update

### LINKED ISSUE #1698: Onboarding Wizard should ask for passphrase
At the moment the onboarding wizard generates a random passphrase when it needs to create a new identify.
It should ask for a user supplied passphrase.

dominikschulz:
Streamline command output, add unicode icons and as for passphrase
during onboarding.

Fixes #1698

### DIFF
--- a/internal/action/commands.go
+++ b/internal/action/commands.go
@@ -4,6 +4,46 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+// ShowFlags returns the flags for the show command. Exported to re-use in main
+// for the default command.
+func ShowFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.BoolFlag{
+			Name:    "clip",
+			Aliases: []string{"c"},
+			Usage:   "Copy the password value into the clipboard",
+		},
+		&cli.BoolFlag{
+			Name:    "alsoclip",
+			Aliases: []string{"C"},
+			Usage:   "Copy the password and show everything",
+		},
+		&cli.BoolFlag{
+			Name:  "qr",
+			Usage: "Print the password as a QR Code",
+		},
+		&cli.BoolFlag{
+			Name:    "unsafe",
+			Aliases: []string{"u", "force", "f"},
+			Usage:   "Display unsafe content (e.g. the password) even if safecontent is enabled",
+		},
+		&cli.BoolFlag{
+			Name:    "password",
+			Aliases: []string{"o"},
+			Usage:   "Display only the password. Takes precedence over all other flags.",
+		},
+		&cli.StringFlag{
+			Name:  "revision",
+			Usage: "Show a past revision. Does NOT support RCS specific shortcuts. Use exact revision or -N to select the Nth oldest revision of this entry.",
+		},
+		&cli.BoolFlag{
+			Name:    "noparsing",
+			Aliases: []string{"n"},
+			Usage:   "Do not parse the output.",
+		},
+	}
+}
+
 // GetCommands returns the cli commands exported by this module
 func (s *Action) GetCommands() []*cli.Command {
 	return []*cli.Command{
@@ -36,11 +76,11 @@ func (s *Action) GetCommands() []*cli.Command {
 		},
 		{
 			Name:  "audit",
-			Usage: "Scan for weak passwords",
+			Usage: "Decrypt all secrets and scan for weak or leaked passwords",
 			Description: "" +
 				"This command decrypts all secrets and checks for common flaws and (optionally) " +
 				"against a list of previously leaked passwords.",
-			Before: s.Initialized,
+			Before: s.IsInitialized,
 			Action: s.Audit,
 			Flags: []cli.Flag{
 				&cli.IntFlag{
@@ -59,13 +99,13 @@ func (s *Action) GetCommands() []*cli.Command {
 				"It can either be used to retrieve the decoded content of a secret " +
 				"similar to 'cat file' or vice versa to encode the content from STDIN " +
 				"to a secret.",
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.Cat,
 			BashComplete: s.Complete,
 		},
 		{
 			Name:  "clone",
-			Usage: "Clone a store from git",
+			Usage: "Clone a store from a git repository",
 			Description: "" +
 				"This command clones an existing password store from a git remote to " +
 				"a local password store. Can be either used to initialize a new root store " +
@@ -88,7 +128,7 @@ func (s *Action) GetCommands() []*cli.Command {
 		},
 		{
 			Name:  "config",
-			Usage: "Edit configuration",
+			Usage: "Configuration editor",
 			Description: "" +
 				"This command allows for easy printing and editing of the configuration. " +
 				"Without argument, the entire config is printed. " +
@@ -105,10 +145,10 @@ func (s *Action) GetCommands() []*cli.Command {
 		},
 		{
 			Name:        "convert",
-			Usage:       "Convert a store",
+			Usage:       "Convert a store to different backends",
 			Description: "Convert a store to a different set of backends",
 			Action:      s.Convert,
-			Before:      s.Initialized,
+			Before:      s.IsInitialized,
 			Hidden:      true,
 			Flags: []cli.Flag{
 				&cli.StringFlag{
@@ -144,7 +184,7 @@ func (s *Action) GetCommands() []*cli.Command {
 				"automatically copy recursively. In that case, the source directory is re-created " +
 				"at the destination if no trailing slash is found, otherwise the contents are " +
 				"flattened (similar to rsync).",
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.Copy,
 			BashComplete: s.Complete,
 			Flags: []cli.Flag{
@@ -157,12 +197,12 @@ func (s *Action) GetCommands() []*cli.Command {
 		},
 		{
 			Name:  "delete",
-			Usage: "Remove secrets",
+			Usage: "Remove one or many secrets from the store",
 			Description: "" +
 				"This command removes secrets. It can work recursively on folders. " +
 				"Recursing across stores is purposefully not supported.",
 			Aliases:      []string{"remove", "rm"},
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.Delete,
 			BashComplete: s.Complete,
 			Flags: []cli.Flag{
@@ -187,7 +227,7 @@ func (s *Action) GetCommands() []*cli.Command {
 				"for storing your secret while the editor is accessing it. Please make " +
 				"sure your editor doesn't leak sensitive data to other locations while " +
 				"editing.",
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.Edit,
 			Aliases:      []string{"set"},
 			BashComplete: s.Complete,
@@ -208,7 +248,7 @@ func (s *Action) GetCommands() []*cli.Command {
 			Name:         "env",
 			Usage:        "Run a subprocess with a pre-populated environment",
 			Description:  "This command runs a sub process with the environment populated from the keys of a secret.",
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.Env,
 			BashComplete: s.Complete,
 			Hidden:       true,
@@ -220,7 +260,7 @@ func (s *Action) GetCommands() []*cli.Command {
 				"This command will first attempt a simple pattern match on the name of the " +
 				"secret.  If there is an exact match it will be shown directly; if there are " +
 				"multiple matches, a selection will be shown.",
-			Before:       s.Initialized,
+			Before:       s.IsInitialized,
 			Action:       s.FindNoFuzzy,
 			Aliases:      []string{"search"},
 			BashComplete: s.Complete,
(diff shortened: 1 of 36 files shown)


internal/action/commands.go
```diff
@@ -36,11 +36,11 @@ func (s *Action) GetCommands() []*cli.Command {
 		},
 		{
 			Name:  "audit",
-			Usage: "Scan for weak passwords",
+			Usage: "Decrypt all secrets and scan for weak or shard passwords",
```
AnomalRoil, line 39: "shard"? Did you mean "shared" as in "leaked"?

internal/action/delete.go
```diff
@@ -63,7 +63,7 @@ func (s *Action) deleteKeyFromYAML(ctx context.Context, name, key string) error
 		return ExitError(ExitIO, err, "Can not delete key '%s' from '%s': %s", key, name, err)
 	}
 	sec.Del(key)
-	if err := s.Store.Set(ctxutil.WithCommitMessage(ctx, "Updated Key in YAML"), name, sec); err != nil {
+	if err := s.Store.Set(ctxutil.WithCommitMessage(ctx, "Updated Key"), name, sec); err != nil {
```
AnomalRoil, line 66: Wouldn't it be best to have the name of the entry in the commit message as well?

internal/action/edit.go
```diff
@@ -30,7 +30,7 @@ func (s *Action) Edit(c *cli.Context) error {
 func (s *Action) edit(ctx context.Context, c *cli.Context, name string) error {
 	ed := editor.Path(c)
 	if err := editor.Check(ctx, ed); err != nil {
-		out.Red(ctx, "Failed to check editor config: %s", err)
+		out.Red(ctx, "❌ Failed to check editor config: %s", err)
```
AnomalRoil, line 33: This would be best done using `out.Error`, no? Since it's an error and it would then print on stderr instead of stdout.

internal/action/edit.go
```diff
@@ -76,7 +76,7 @@ func (s *Action) editGetContent(ctx context.Context, name string, create bool) (
 			return nil
 		}
 		if err := s.find(ctxutil.WithFuzzySearch(ctx, false), nil, name, cb); err == nil {
-			cont, err := termio.AskForBool(ctx, fmt.Sprintf("No secret at '%s'. Found possible match in '%s'. Edit existing entry?", name, newName), true)
+			cont, err := termio.AskForBool(ctx, fmt.Sprintf("Secret does not exist %q. Found possible match in %q. Edit existing entry?", name, newName), true)
```
AnomalRoil, line 79: This prints:
> Secret does not exist "tests". Found possible match in "test". Edit existing entry?

Wouldn't it be best like this:
`"Secret %q does not exist. Found possible match with %q. Edit this entry instead?"`

internal/action/edit.go
```diff
@@ -97,7 +97,7 @@ func (s *Action) editGetContent(ctx context.Context, name string, create bool) (
 	}
 
 	if !create {
-		out.Yellow(ctx, "Entry %s not found. Creating new secret ...", name)
+		out.Yellow(ctx, "⚠ Entry %s not found. Creating new secret ...", name)
```
AnomalRoil, line 100: This could be printed on stderr with a "Warning" prepended using the `out.Warning`, no?

internal/action/repl.go
```diff
@@ -114,6 +114,11 @@ func (s *Action) REPL(c *cli.Context) error {
 		}
 		out.Red(c.Context, "Error: %s", err)
```
AnomalRoil, line 115: This could be using `out.Error`

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+
+	debug.Log("Starting Onboarding Wizard - remote: %s - team: %s - create: %t - name: %s - email: %s", remote, team, create, ctxutil.GetUsername(ctx), ctxutil.GetEmail(ctx))
+
+	crypto := s.getCryptoFor(ctx, team)
+	if crypto == nil {
+		return fmt.Errorf("can not continue without crypto")
+	}
+	debug.Log("Crypto Backend initialized as: %s", crypto.Name())
+
+	// check for existing GPG keypairs (private/secret keys). We need at least
```
AnomalRoil, line 60: Is this only looking for GPG keypairs? What about the `age` backend?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	// one useable key pair. If none exists try to create one
+	if !s.initHasUseablePrivateKeys(ctx, crypto) {
+		out.Print(ctx, "🔐 No useable cryptographic keys. Generating new key pair")
+		out.Print(ctx, "🕰 Key generation may take up to a few minutes")
+		if err := s.initGenerateIdentity(ctx, crypto, ctxutil.GetUsername(ctx), ctxutil.GetEmail(ctx)); err != nil {
+			return errors.Wrapf(err, "failed to create new private key")
+		}
+		out.Print(ctx, "🔐 Cryptographic keys validated")
```
AnomalRoil, line 68: Are they "validated" or "generated" at this point?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	debug.Log("Has useable private keys")
```
AnomalRoil, line 71: Might be good to log the corresponding public key, no?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	if bv, err := termio.AskForBool(ctx, "⚠ Do you want to enter a passphrase? (otherwise we generate one for you)", false); err != nil && bv {
+		pwGenerated = false
+		sv, err := termio.AskForPassword(ctx, "✍ Please enter your passphrase")
+		if err != nil {
+			return errors.Wrapf(err, "Failed to read passphrase")
+		}
+		passphrase = sv
+	}
+
+	out.Yellow(ctx, "⏳ This can take a long time. If you get impatient see https://github.com/gopasspw/gopass/blob/master/docs/entropy.md")
```
AnomalRoil, line 110: This shouldn't be the case since kernel version 5.6, since `/dev/random` now only blocks if the CPRNG hasn't been yet initialized and then it becomes non-blocking just like `/dev/urandom`.

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	if want, err := termio.AskForBool(ctx, "Continue?", true); err != nil || !want {
+		return errors.Wrapf(err, "User aborted")
+	}
+
+	if err := crypto.GenerateIdentity(ctx, name, email, passphrase); err != nil {
+		return errors.Wrapf(err, "failed to create new private key in batch mode")
```
AnomalRoil, line 116: Why mention the batch mode here? That's not true for `age` or other backend than gpg, I guess, no?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	out.Print(ctx, "✅ Key pair generated")
+
+	if pwGenerated {
+		out.Print(ctx, color.MagentaString("Passphrase: ")+color.HiGreenString(passphrase))
```
AnomalRoil, line 122: It might be good to tell the user they need to remember this.

AnomalRoil: Also I'm not too fond of having the passphrase mixed with color symbols, since depending on the shell/terminal the color codes can render as characters.

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	kl, err := crypto.ListIdentities(gpg.WithUseCache(ctx, false))
+	if err != nil {
+		return errors.Wrapf(err, "failed to list private keys")
+	}
+	if len(kl) > 1 {
+		out.Print(ctx, "⚠ More than one private key detected. Make sure to use the correct one!")
+		return nil
+	}
+	if len(kl) < 1 {
+		debug.Log("Private Keys: %+v", kl)
```
AnomalRoil, line 135: We should never output private keys in logs, if you ask me... But in this case I'm not sure what that would output anyway since `len(kl)<0`

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	if err := s.initExportPublicKey(ctx, crypto, kl[0]); err != nil {
+		return err
+	}
+	out.Green(ctx, "✅ Key pair validated")
+	return nil
+}
+
+func (s *Action) initExportPublicKey(ctx context.Context, crypto backend.Crypto, key string) error {
+	fn := key + ".pub.key"
+	want, err := termio.AskForBool(ctx, fmt.Sprintf("Do you want to export your public key to %q?", fn), false)
```
AnomalRoil, line 149: maybe mention it will copy it as a file in the current directory?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	if err := ioutil.WriteFile(fn, pk, 06444); err != nil {
+		out.Error(ctx, "❌ Failed to export public key %q: %q", fn, err)
+		return err
+	}
+	out.Print(ctx, "✴ Public key exported to '%s'", fn)
+	return nil
+}
+
+func (s *Action) initHasUseablePrivateKeys(ctx context.Context, crypto backend.Crypto) bool {
+	kl, err := crypto.ListIdentities(gpg.WithAlwaysTrust(ctx, false))
```
AnomalRoil, line 169: Not clear why you need to set `gpg.WithAlwaysTrust` to false, maybe add a comment?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+	return len(kl) > 0
+}
+
+func (s *Action) initSetupGitRemote(ctx context.Context, team, remote string) error {
+	var err error
+	remote, err = termio.AskForString(ctx, "Please enter the git remote for your shared store", remote)
+	if err != nil {
+		return errors.Wrapf(err, "failed to read user input")
+	}
+	{
```
AnomalRoil, line 182: Why the scope?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+		ctx := ctxutil.WithHidden(ctx, true)
```
AnomalRoil, line 183: Why `WithHidden`?

internal/action/setup.go
```diff
@@ -0,0 +1,287 @@
...
+
+	out.Print(ctx, "Configuring the git remote ...")
+	if err := s.initSetupGitRemote(ctx, team, remote); err != nil {
+		return errors.Wrapf(err, "failed to setup git remote")
+	}
+	out.Print(ctx, "✅ Done. Created Team %q", team)
+	return nil
+}
+
+// initJoinTeam will create a local root store and clone and existing store to
```
AnomalRoil, line 257: "and clone **an** existing store "

internal/action/show.go
```diff
@@ -174,7 +174,7 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 		debug.Log("got(found: %t) key %s: %s", found, key, val)
 		return val, val, nil
 	} else if HasKey(ctx) {
-		out.Warning(ctx, "parsing is disabled but a key was provided.")
+		out.Warning(ctx, "⚠ Parsing is disabled but a key was provided.")
```
AnomalRoil, line 177: Maybe add "⚠" to the prefix of `out.Warning` instead?

AnomalRoil: You've actually already done so, so this ⚠ can be removed here.

AnomalRoil (changes requested):

internal/action/commands.go
dominikschulz: should be "leaked".

internal/action/edit.go
dominikschulz: actually it should be a warning, but otherwise: yes.

internal/action/delete.go
dominikschulz: the name of the secret or the name of the key?
adding the key would leak content of the secret to the plain commit history. not a good idea.
the name of the secret is visible in the commit anyway. if there is any use we can duplicate that information to the commit message, but I don't see why.

internal/action/edit.go
dominikschulz: yes, much better.

internal/action/repl.go
dominikschulz: Done.

internal/action/setup.go
dominikschulz: indeed.

internal/action/setup.go
dominikschulz: generated seems more apt.

internal/action/setup.go
dominikschulz: Done.

internal/action/setup.go
dominikschulz: Kernel 5.6? You seem to be living in the future ;)

Let's keep this around for a little longer.

internal/action/setup.go
dominikschulz: Done.

internal/action/setup.go
dominikschulz: Done.

internal/action/setup.go
dominikschulz: No, we must never log secret material, but these lists should only contain key IDs.
However in this case it must always be empty, so removing it.

internal/action/setup.go
dominikschulz: It prints the filename, that should be enough?

internal/action/setup.go
dominikschulz: Done.

internal/action/setup.go
dominikschulz: this became obsolete with one of the past refactorings

internal/action/setup.go
dominikschulz: IIRC we didn't want possible clutter from the RCS ops.

internal/action/setup.go
dominikschulz: Done.

internal/action/show.go
dominikschulz: Done.

internal/action/edit.go
dominikschulz: Done.

AnomalRoil (approved):

PR #1717, opened 2021-01-16
Implement signature verification for self updater

### LINKED ISSUE #1676: Implement gopass self-update

dominikschulz:
Fixes #1676

RELEASE_NOTES=[BUGFIX] Add signature verification for updater


internal/updater/verify.go
```diff
@@ -0,0 +1,281 @@
...
+	"bytes"
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	"github.com/gopasspw/gopass/pkg/debug"
+	"golang.org/x/crypto/openpgp"
+)
+
+var pubkey = []byte(`
```
AnomalRoil, line 14: Ideally we could have a secret key in our CI/DI and its pub key would be here.
I guess this correspond to a local key you have on your machine?

internal/updater/verify.go
```diff
@@ -0,0 +1,281 @@
...
+			continue
+		}
+		h, err := hex.DecodeString(p[0])
+		if err != nil {
+			return nil, err
+		}
+		return h, nil
+	}
+
+	return nil, fmt.Errorf("hash for file %q not found", filename)
```
AnomalRoil, line 280: Ideally the hashsum should be recomputed and compared to that in the hashsum file

internal/updater/verify.go
AnomalRoil: Also you could have used a ECDSA key, they are much shorter.

internal/updater/download.go
```diff
@@ -0,0 +1,119 @@
...
+	bar.Done()
+	debug.Log("Transferred %d bytes from %s to %s", count, url, dest)
+	return nil
+}
+
+type setter interface {
+	Set(int64)
+}
+
+type passThru struct {
```
AnomalRoil, line 108: What for?

internal/updater/download.go
```diff
@@ -0,0 +1,119 @@
...
+		case <-ctx.Done():
+			return backoff.Permanent(errors.New("user aborted"))
+		default:
+		}
+		return download(ctx, dest, url)
+	}, bo)
+}
+
+func download(ctx context.Context, dest, url string) error {
+	fh, err := os.OpenFile(dest, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0755)
```
AnomalRoil, line 76: You should defer close it, no?

internal/updater/github.go
```diff
@@ -0,0 +1,97 @@
...
+// FetchReleases fetches and returns all releases of gopass from GitHub
+func FetchReleases(pre bool) ([]ghrel.Release, error) {
+	if pre {
+		return ghrel.FetchAllReleases(gitHubOrg, gitHubRepo)
+	}
+	return ghrel.FetchAllStableReleases(gitHubOrg, gitHubRepo)
+}
+
+// LatestRelease fetches and returns the latest gopass release from GitHub
+func LatestRelease(pre bool) (ghrel.Release, error) {
```
AnomalRoil, line 88: This is available in https://github.com/dominikschulz/github-releases/blob/master/ghrel/releases.go#L111, no?

internal/updater/download.go
```diff
@@ -0,0 +1,119 @@
...
+		name := filepath.Base(header.Name)
+		if header.Typeflag == tar.TypeReg && name == "gopass" {
+			_, err := io.Copy(dfh, tarReader)
+			return errors.Wrapf(err, "Failed to read gopass from tar file")
+		}
+	}
+	return errors.Errorf("file not found in archive")
+}
+
+func tryDownload(ctx context.Context, dest, url string) error {
```
AnomalRoil, line 61: Isn't the download part done by `ghrel` already?

internal/updater/verify.go
dominikschulz: Yes, this is my key for now. That's needs to change eventually, but for implementing and testing that should be OK.

AnomalRoil:
I'm actually wondering whether you shouldn't be adding this capability directly into `ghrel` along with a README.

internal/updater/verify.go
dominikschulz: Yes, of course. This wasn't ready for review, yet. Sorry.

internal/updater/download.go
dominikschulz: This is for the download progressbar.

internal/updater/download.go
dominikschulz: No.

internal/updater/github.go
dominikschulz: No, I'm ditching that for now.

internal/action/version.go
```diff
@@ -76,30 +77,43 @@ func versionInfo(ctx context.Context, v versioner) string {
 func (s *Action) checkVersion(ctx context.Context, u chan string) {
 	if disabled := os.Getenv("CHECKPOINT_DISABLE"); disabled != "" {
 		u <- ""
+		debug.Log("remote version check disabled by CHECKPOINT_DISABLE")
 		return
 	}
 
-	if strings.HasSuffix(s.version.String(), "+HEAD") || protect.ProtectEnabled {
-		// chan not check version against HEAD or
+	force := os.Getenv("GOPASS_FORCE_CHECK") == "true"
```
AnomalRoil, line 84: `CHECKPOINT_DISABLE` can be any non-null string, but `GOPASS_FORCE_CHECK` must be `"true"`, maybe we should unify?

internal/updater/extract.go
```diff
@@ -0,0 +1,77 @@
...
+	switch filepath.Ext(filename) {
+	case ".gz":
+		gzr, err := gzip.NewReader(rd)
+		if err != nil {
+			return err
+		}
+		rd = gzr
+	case ".bz2":
+		rd = bzip2.NewReader(rd)
+	case ".zip":
```
AnomalRoil, line 36: So, the updater should exit early on Windows devices then.

AnomalRoil: There's https://golang.org/pkg/archive/zip/ that could handle it.

internal/updater/update.go
```diff
@@ -1,285 +1,101 @@
...
-func download(ctx context.Context, dest, url string) error {
-	fh, err := os.OpenFile(dest, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0755)
-	if err != nil {
-		return err
-	}
-	resp, err := http.Get(url)
-	if err != nil {
+	debug.Log("extracting binary from tarball ...")
+	if err := extractFile(buf, dlFilename, dest); err != nil {
 		return err
```
AnomalRoil, line 95: So, currently the Windows updates will fail here since `extractFile` doesn't support zip files.

AnomalRoil (dismissed):
LGTM, maybe it'd be best to fix the Windows path?

internal/updater/update.go
dominikschulz: I guess we should bail out early.

internal/updater/extract.go
dominikschulz: Yes, indeed. But I was too lazy do add windows support. I don't think that's much of a problem, but if this is working well on non-Windows OS we can still add Windows support later. Should be fairly easy to plug it in here.

internal/action/version.go
dominikschulz: Good idea.

internal/updater/extract.go
AnomalRoil: Sounds good.

dominikschulz:
Windows will now fail early until it's supported.

AnomalRoil (approved):

PR #1719, opened 2021-01-17
Do not exit on non-empty second pw

### LINKED ISSUE #1718: Invalid exit condition in AskForPass
## Current behaviour
We currently return correctly is someone enters the empty password and then enters a valid password, and we return the empty password in that case:
https://github.com/gopasspw/gopass/blob/774eb1d1a9f43ede92b6b6eaca6b28059ef99d64/pkg/termio/ask.go#L175-L177

## Expected behaviour
We should not accept the empty password unless it was entered twice.

AnomalRoil:
RELEASE_NOTES=[BUGFIX] The empty password must now be confirmed too

Fixes #1718

I simply removed the `|| pass == ""`.

Do we want to have an extra confirmation step in case of empty password or not?

### DIFF
--- a/pkg/termio/ask.go
+++ b/pkg/termio/ask.go
@@ -155,7 +155,7 @@ func AskForPassword(ctx context.Context, name string) (string, error) {
 
 	askFn := GetPassPromptFunc(ctx)
 	for i := 0; i < maxTries; i++ {
-		// check for context cancelation
+		// check for context cancellation
 		select {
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
LGTM

dominikschulz:
I guess it's a UX question if we should warn on empty passwords.
I'm inclined to say yes, but not sure.

PR #1720, opened 2021-01-18
Remove the ondisk backend

dominikschulz:
This backend has never been fully completed and shouldn't be used
as-is.

RELEASE_NOTES=[CLEANUP] Remove the ondisk backend


internal/action/commands.go
```diff
@@ -815,17 +814,17 @@ func (s *Action) GetCommands() []*cli.Command {
 				},
 				&cli.StringFlag{
 					Name:  "crypto",
-					Usage: "Select crypto backend (gpg, gpgcli, plain, xc)",
+					Usage: fmt.Sprintf("Select crypto backend %v", backend.CryptoBackends()),
 				},
 				&cli.StringFlag{
 					Name:  "rcs",
```
AnomalRoil, line 820: Why `rcs` here and `storage` for the other ones?

AnomalRoil (approved):

internal/action/commands.go
dominikschulz: Actually we'll have to go over all the command flags. I guess more than one will be inconsistent / ineffective.
I plan on doing that later.

PR #1724, opened 2021-01-18
Add windows support to the self updater

### LINKED ISSUE #1722: Windows support for gopass update
The new gopass updater should support windows as well.
Currently the implementation for that platform is stubbed out.

dominikschulz:
Fixes #1722

RELEASE_NOTES=[ENHANCEMENT] Add windows support to the self updater


internal/updater/extract.go
```diff
@@ -23,55 +24,94 @@ func extractFile(buf []byte, filename, dest string) error {
...
+		}
+
+		file, err := zrd.File[i].Open()
+		if err != nil {
+			return errors.Wrapf(err, "failed to read from zip file")
+		}
+
+		n, err := io.Copy(dfh, file)
+		if err != nil {
+			dfh.Close()
```
AnomalRoil, line 76: The `defer`ed Close above would return an error finding `dfh` is already closed, but we don't check the return value of all the Close anyway... Not sure what's best.

internal/updater/extract.go
```diff
@@ -23,55 +24,94 @@ func extractFile(buf []byte, filename, dest string) error {
...
+			return fmt.Errorf("unable to remove destination file: %q", err)
+		}
+	}
+
+	dfh, err := os.OpenFile(dest, os.O_WRONLY|os.O_CREATE|os.O_EXCL, mode)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to open file: %s", dest)
+	}
+	defer func() {
+		_ = dfh.Close()
```
AnomalRoil, line 38: Since this is a writable file it would be best to check Close for errors as these can happen upon closing with the sync and all.

internal/updater/extract.go
```diff
@@ -23,55 +24,94 @@ func extractFile(buf []byte, filename, dest string) error {
...
+		if header.Typeflag != tar.TypeReg {
+			continue
+		}
+		if name != "gopass" {
+			continue
+		}
+
+		n, err := io.Copy(dfh, tarReader)
+		if err != nil {
+			dfh.Close()
```
AnomalRoil, line 108: Same.

AnomalRoil (approved):

internal/updater/extract.go
dominikschulz: Yes, indeed. Doing this correctly under all circumstances is surprisingly difficult.
I really wonder if it's worth the hassle here.

internal/updater/extract.go
dominikschulz: Yes ...

internal/updater/extract.go
dominikschulz: :/

d-buchmann:
Hi, I just downloaded `gopass-1.12.8-windows-amd64.zip` and ran `gopass update`, but it tells me "gopass update is not supported on windows (#1722)".
Am I doing something wrong or is the fix not deployed as it should be?
I haven't looked at the code yet nor am I any familiar with `go`, hence the question...

AnomalRoil:
Are you sure you are running the correct gopass version @d-buchmann ?
What do these two commands tell you:
```
gopass version
where.exe gopass
```

d-buchmann:
quite sure...

`gopass version`:
gopass 1.12.8+a7e343465b4e45b0bb5b9c3dfe12032539506d6b (a7e343465b4e45b0bb5b9c3dfe12032539506d6b) go1.16.7 windows amd64
<root>     -  gpg 2.3.1 -  git 0.0.0
Available Storage Backends: fs, gitfs

`where.exe gopass`:
C:\dev\gopass-1.12.8-windows-amd64\gopass.exe (working dir, correct)

AnomalRoil:
That's really strange since we removed that message in #1790

d-buchmann:
Removed - well not completely; It's still in the windows binary 1.12.8 (at offset 0x007c9671, if you're interested).
Maybe one of you could have a look at the build/deployment process with regard to Windows?
Sorry I can't be of help though...

AnomalRoil:
Yeah, it's probably cached somewhere in our Github Actions. I'll have to dig into it.

d-buchmann:
I found `branch:fix/upwin` which is unfortunately deleted and logs are already purged as well...

AnomalRoil:
I think I've found the culprit:
https://github.com/gopasspw/gopass/blob/b494532de7a887d29d1c9076838f5243f9a508ec/internal/action/update.go#L25-L27

d-buchmann:
Exactly. I removed that, built it and now the update runs fine until I get
`Failed to update gopass: unable to remove destination file: "remove C:\\dev\\gopass\\gopass.exe: Access is denied."`
I kind of expected that on windows.

AnomalRoil:
I'm guessing there's no "fix" on our side, excepted to ask for admin privileges to do the update?

d-buchmann:
No, it's not an issue with privileges (I tested it), it's rather that the file to be removed is being executed and therefore locked.
I made a self-updater (in c#) on windows like this:
- Get new version of file `x`, save it under different name `y`
- run `y` from `x` (which is then terminated), copy the file `y` to `x`
- run `x` (the new version) and optionally/finally remove `y`
- ???
- Absolute madness.

PR #1725, opened 2021-01-18
Delete empty parent directories on gopass rm -r

### LINKED ISSUE #1723: Gopass does not list empty folders and rm -r allows to create them
### Summary

If one creates an entry in a subfolder of a subfolder and deletes the last subfolder using `rm -r`, even if the first subfolder is empty, it won't be deleted, whereas `gopass rm entry` does clean the folders in the entry path if they are empty.

## To reproduce

```
gopass generate new/folder/test/hello 10
gopass ls new
gopass rm -r new/folder/test
gopass ls new
> Error: Entry 'new' not found
ls $(gopass config path | cut -d ' ' -f 2)/new
> folder
gopass generate new/folder/test/hello 10
gopass rm new/folder/test/hello 
ls $(gopass config path | cut -d ' ' -f 2)/new
> ls: cannot access '/home/anomalroil/.local/share/gopass/stores/root/new': No such file or directory
```

dominikschulz:
This adds house keeping around empty parent directories. They would
have been left over in some cases. This adds more proetections to avoid
that, i.e. removing empty parents on rm -r (not only rm) as well as
fsck checks.

Fixes #1723

### DIFF
--- a/internal/action/delete.go
+++ b/internal/action/delete.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"github.com/gopasspw/gopass/pkg/ctxutil"
+	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/termio"
 
 	"github.com/urfave/cli/v2"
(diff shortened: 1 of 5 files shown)


AnomalRoil (approved):
LGTM

PR #1728, opened 2021-01-19
Adding release Github Action

AnomalRoil:
This is a basic action added through the Github interface to see if goreleaser works out of the box.

### DIFF
--- a/.github/workflows/autorelease.yml
+++ b/.github/workflows/autorelease.yml
@@ -1,0 +1,33 @@
+# This is a basic workflow to help you get started with Actions
+
+name: release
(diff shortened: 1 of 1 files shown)


AnomalRoil:
This should trigger when we push to a tag `v*`, I'll try that next to see how it works: I didn't understand what's happening with the releases once they are produced yet 😆 

Signing should be super easy to add once we've figured out how we want to handle the GPG key part:
https://github.com/goreleaser/goreleaser-action#signing

dominikschulz (approved):
LGTM

PR #1733, opened 2021-01-19
Release signing setup

dominikschulz:
This is adding release signing using the newly created dedicate release signing key 0x21FF4D8F35D1AEE5.
The private key is placed in a secret and [imported](https://github.com/marketplace/actions/import-gpg) into the gpg keyring of the release container.

The `GOPATH` issue is addressed by setting it to a [fixed value](https://laszlo.cloud/setting-gopath-in-github-actions). Let's see how that works. In doubt we might be able to remove trimpath from gcflags and asmflags, but then the path might still show up in some locations.

For a fully working GoReleaser-on-GHA setup at least `GITHUB_TOKEN` is still missing.


.github/workflows/autorelease.yml
```diff
@@ -23,11 +23,33 @@ jobs:
...
+          echo ${GOROOT}
+          env
       -
         name: Run GoReleaser
         uses: goreleaser/goreleaser-action@v2
         with:
           version: latest
           args: release --rm-dist
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
AnomalRoil, line 33: The `secrets.GITHUB_TOKEN` is an auto-generated token, no? Why change it?

AnomalRoil: See https://docs.github.com/en/actions/reference/authentication-in-a-workflow#about-the-github_token-secret

internal/updater/verify.go
```diff
@@ -11,237 +11,23 @@ import (
 	"golang.org/x/crypto/openpgp"
 )
 
+// gpg --expert --full-generate-key
+// (10) ECC (sign only)
+// (5) NIST P-512 (see https://github.com/golang/crypto/blob/eec23a3978ad/openpgp/packet/public_key.go#L227)
+// 2y
```
AnomalRoil, line 17: I'm totally not looking forward to the time where we'll have to rotate that key.

.goreleaser.yml
```diff
@@ -133,3 +133,4 @@ signs:
   -
     id: gopass
     artifacts: checksum
+    args: ["-u", "0x21FF4D8F35D1AEE5", "--armor", "--output", "${signature}", "--detach-sign", "${artifact}"]
```
AnomalRoil, line 136: Maybe use the "--batch" args here? I believe it is passed to `gpg`

AnomalRoil (approved):

.goreleaser.yml
dominikschulz: Shouldn't be necessary: https://goreleaser.com/customization/sign/

internal/updater/verify.go
dominikschulz: Yes ... but I think we can just add another key to the byte slice and it should just work. If I understood the golang.org/x/crypto/openpgp docs correctly.

.github/workflows/autorelease.yml
dominikschulz: I guess we can re-use that.

AnomalRoil (approved):

PR #1734, opened 2021-01-19
Avoid age keyring being recreated when a wrong passphrase is entered

### LINKED ISSUE #1678: age-keyring.age: secret key is regenerated after some use
Thanks for this awesome software, really appreciate your hard work.

### Summary
I'm was hoping to migrate to the `age` backend, but it seems that the secret key stored in `~/.config/gopass/age-keyring.age` is modified after decrypting `age`-encrypted secrets several times, making it such that one can no longer decrypt them.

### Steps To Reproduce
I cannot provide instructions to reproduce this directly, but I'll describe my setup and say this has happened about 5 times over the past few days.

1. I have my old `gpg` store mounted (not sure this is relevant)
2. My 'new' `age` store is the primary `gitfs` store
3. Generate/Insert a new password in the `age` store, which uses the keyring `~/.config/gopass/age-keyring.age` 
4. Make a copy of the the private key in `~/.config/gopass/age-keyring.age` (decrypt using passphrase)
4. Execute `gopass show -o <secret>` and enter the passphrase to open the keyring and decrypt the password (works as expected)
5. Repeat the previous step (4) several times until you encounter the error:
```
Decryption failed: no identity matched a recipient
Error: failed to retrieve secret '<secret>': Failed to decrypt
```
6. Compare the old private key with the new and notice that a new private key has been generated: one can no longer decrypt the secret.

### Expected behavior

`gopass show -o` should not affect `~/.config/gopass/age-keyring.age`.

### Environment

```
❯  gopass version
gopass 1.10.1 (2020-09-13 00:16:24) go1.15.2 linux amd64
<root>     -  age 0.0.1 - git 2.29.2
old        - gpg 2.2.25 - git 2.29.2
```
```
❯  uname -a
Linux x230 5.9.11-3-MANJARO #1 SMP PREEMPT Sat Nov 28 09:08:57 UTC 2020 x86_64 GNU/Linux
```
Installed via `yay -S gopass`.

dominikschulz:
While this is arguably secure the useability is terrible and of course
this is a bug. Before this fix we would threat every kind of error (including
one from a wrong passphrase) the same and just blindly generate
a new identity, overwriting the keyring. This wouldn't be able to decrypt
the previously encrypted secrets anymore.

Please note that the whole age backend must be still considered a proof-of-concept.
The concept of a keyring doesn't match age very well and if possible I'd like to
remove this pseudo keyring again in the future.

Fixes #1678

### DIFF
--- a/internal/backend/crypto/age/decrypt.go
+++ b/internal/backend/crypto/age/decrypt.go
@@ -41,7 +41,7 @@ func (a *Age) decrypt(ciphertext []byte, ids ...age.Identity) ([]byte, error) {
 }
 
 func (a *Age) decryptFile(ctx context.Context, filename string) ([]byte, error) {
-	pw, err := ctxutil.GetPasswordCallback(ctx)("index") // TODO should pass filename?
+	pw, err := ctxutil.GetPasswordCallback(ctx)(filename)
 	if err != nil {
 		return nil, err
 	}
--- a/internal/backend/crypto/age/age.go
+++ b/internal/backend/crypto/age/age.go
@@ -2,7 +2,9 @@ package age
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"path/filepath"
 	"sort"
 	"strings"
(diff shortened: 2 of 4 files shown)


internal/backend/crypto/age/decrypt.go
```diff
@@ -41,7 +41,7 @@ func (a *Age) decrypt(ciphertext []byte, ids ...age.Identity) ([]byte, error) {
 }
 
 func (a *Age) decryptFile(ctx context.Context, filename string) ([]byte, error) {
-	pw, err := ctxutil.GetPasswordCallback(ctx)("index") // TODO should pass filename?
+	pw, err := ctxutil.GetPasswordCallback(ctx)(filename)
```
AnomalRoil, line 44: Please change it in `encrypt.go` too then

AnomalRoil (changes requested):

internal/backend/crypto/age/decrypt.go
dominikschulz: Done.

PR #1741, opened 2021-01-20
Adding support for multiple same keys

AnomalRoil:
RELEASE_NOTES=[ENHANCEMENT] KV secrets are now key-values, supporting multiple same key with different values

### DIFF
--- a/pkg/gopass/secrets/kv.go
+++ b/pkg/gopass/secrets/kv.go
@@ -17,15 +17,15 @@ var _ gopass.Secret = &KV{}
 // NewKV creates a new KV secret
 func NewKV() *KV {
 	return &KV{
-		data: make(map[string]string, 10),
+		data: make(map[string][]string, 10),
 	}
 }
 
 // NewKVWithData returns a new KV secret populated with data
-func NewKVWithData(pw string, kvps map[string]string, body string, converted bool) *KV {
+func NewKVWithData(pw string, kvps map[string][]string, body string, converted bool) *KV {
 	kv := &KV{
 		password: pw,
-		data:     make(map[string]string, len(kvps)),
+		data:     make(map[string][]string, len(kvps)),
 		body:     body,
 		fromMime: converted,
 	}
@@ -72,7 +72,7 @@ func NewKVWithData(pw string, kvps map[string]string, body string, converted boo
 //   - body: "Yo\nHi"
 type KV struct {
 	password string
-	data     map[string]string
+	data     map[string][]string
 	body     string
 	fromMime bool
 }
@@ -87,10 +87,12 @@ func (k *KV) Bytes() []byte {
 		if !ok {
 			continue
 		}
-		_, _ = buf.WriteString(key)
-		_, _ = buf.WriteString(": ")
-		_, _ = buf.WriteString(sv)
-		_, _ = buf.WriteString("\n")
+		for _, v := range sv {
+			_, _ = buf.WriteString(key)
+			_, _ = buf.WriteString(": ")
+			_, _ = buf.WriteString(v)
+			_, _ = buf.WriteString("\n")
+		}
 	}
 	buf.WriteString(k.body)
 	return buf.Bytes()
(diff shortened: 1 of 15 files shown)


AnomalRoil:
This fixes #1576

pkg/gopass/secrets/yaml.go
```diff
@@ -71,6 +77,11 @@ func (y *YAML) Set(key string, value interface{}) error {
 	return nil
 }
 
+// Add doesn't work since as per YAML specification keys must be unique
+func (y *YAML) Add(key string, value interface{}) error {
+	return fmt.Errorf("not supported for PLAIN")
```
dominikschulz, line 82: this should be `YAML`

pkg/gopass/secrets/kv.go
```diff
@@ -102,28 +104,49 @@ func (k *KV) Keys() []string {
...
 	return keys
 }
 
-// Get returns a single key
+// Get returns the first value of that key
 func (k *KV) Get(key string) (string, bool) {
 	key = strings.ToLower(key)
+
+	if v, found := k.data[key]; found {
+		return v[0], true
```
dominikschulz, line 119: this is deterministic, isn't it?
the values should be in that slice in the order they are defined in the secret?

pkg/gopass/secrets/yaml.go
```diff
@@ -41,14 +41,14 @@ func (y *YAML) Keys() []string {
 	for key := range y.data {
 		keys = append(keys, key)
 	}
+	sort.Strings(keys)
 	if _, found := y.data["password"]; !found {
-		keys = append(keys, "password")
+		keys = append([]string{"password"}, keys...)
```
dominikschulz, line 46: You seem to always order the key `password` first, which is good.
But maybe we should write down that assumption somewhere?

dominikschulz (approved):
LGTM. Very good, thank you!

Some questions we might want to address before merging, though.

pkg/gopass/secrets/kv.go
AnomalRoil: Yes, they should

pkg/gopass/secrets/yaml.go
AnomalRoil: Yeah, since we said that the secret would always be on the first line, I figured that we might as well always have the password key first (unless there's a password key in the secret, but maybe we might want to change that to have multiple password keys in that case?)
We might also want to include a check that the password is non-null, no?

dominikschulz: Hmm, thinking more about it this is a bit weird.
The `key` `password` isn't necessarily the same as the `password` entry (the first line).

So we could very well have a password field (first line) and one - or even multiple - (KV or YAML) keys named `password`.

AnomalRoil: Yes, exactly. Notice that if the key `password` is in the entry, the extra `password` key at the beginning is skipped, the rational was that currently if we have multiple "same" key, we do not display them in the same order as they appear in the secret since we sort them first and so a `password` key in the body should be sorted with the other keys...
 
We could keep track of the key order and display multiple keys if we were to add a field to the KV type to retain the keys and their order upon parsing... WDYT?

dominikschulz: Uh, that sounds like a lot of extra complexity. I'd rather not do that.

pkg/gopass/secrets/kv.go
```diff
@@ -102,28 +104,49 @@ func (k *KV) Keys() []string {
 	for key := range k.data {
 		keys = append(keys, key)
 	}
+	sort.Strings(keys)
 	if _, found := k.data["password"]; !found {
-		keys = append(keys, "password")
+		keys = append([]string{"password"}, keys...)
 	}
```
dominikschulz, line 110: Drop this if block (line 108 - 110). We don't need it.

pkg/gopass/secrets/kv.go
```diff
@@ -102,28 +104,49 @@ func (k *KV) Keys() []string {
...
 	v, found := k.data[key]
 	return v, found
 }
 
 // Set writes a single key
 func (k *KV) Set(key string, value interface{}) error {
 	key = strings.ToLower(key)
-	k.data[key] = fmt.Sprintf("%s", value)
+	if v, ok := k.data[key]; ok && len(v) > 1 {
+		return fmt.Errorf("cannot set key %s: this entry contains multiple same keys. Please use edit instead", key)
```
dominikschulz, line 136: `... use 'gopass edit' instead`

pkg/gopass/secrets/yaml.go
dominikschulz: Thinking more about this: please drop the if block (line 45-47).
We don't need it and it doesn't make much sense, i.e. you won't get the content of the first line with a `Get("password")` anyway. And if there is one (or multiple) key with the name `password` we can handle it like any other key.

dominikschulz (changes requested):

dominikschulz (approved):
LGTM

PR #1743, opened 2021-01-21
Remove colored output

### LINKED ISSUE #1464: Colors selection is bad for white terminal background
Colors selection is bad if a black on white terminal is in use. Especially the yellow password output is nearly unreadable.

It would be nice to have a ``color`` option instead of ``nocolor`` which would e.g. be ``dark``, ``light`` or ``off``.

### LINKED ISSUE #1467: Cannot update root RCS backend after store creation
### Summary

I have an existing store (coming from pass) that, I believe, does not have an associated git repository (is there a way to check that?).
I want to initialize one (i.e. switch it to the gitcli RCS backend) to do synchronization, but it doesn't work.

### Steps To Reproduce

1. Run `gopass git init`. (This gives `Initializing git repository (noop) ...` and `Git initialized`, but I can re-run the same command and get the same result, so I'm not sure it works.)
2. Run `gopass git remote add origin …` (This prints nothing.)
3. Run `gopass sync`. (This prints `WARNING: Mount uses RCS backend 'noop'. Not syncing!` even though my path contains `gpgcli-gitcli-fs+file://…`.

### Expected behavior

Git repository should be initialized and synced.

### Environment

- OS: ArchLinux
- OS version: Linux antoyo-laptop 5.7.8-arch1-1 #1 SMP PREEMPT Thu, 09 Jul 2020 16:34:01 +0000 x86_64 GNU/Linux
- gopass Version:
```
gopass 1.9.2 (2020-07-13 14:34:42) go1.14.4 linux amd64
<root>     - gpg 2.2.21 - noop 0.0.0 -   fs 0.1.0
Available Crypto Backends: age, gpgcli, plain, xc
Available RCS Backends: gitcli, noop
Available Storage Backends: fs, inmem
```
- Installation method: ArchLinux repository

The line `<root>     - gpg 2.2.21 - noop 0.0.0 -   fs 0.1.0` seem to indicate that the root mount is using the `noop` backend, but I cannot figure out how to change it to `gitcli`.
### Additional context

If it's not possible to change the RCS backend of the root mount, how can I workaround this issue?

dominikschulz:
This confuses some terminals.

Also: fix setup.go condition and validate git setup.

Fixes #1467
Fixes #1464 

### DIFF
--- a/internal/store/leaf/write.go
+++ b/internal/store/leaf/write.go
@@ -97,6 +97,5 @@ func (s *Store) gitCommitAndPush(ctx context.Context, name string) error {
 		return errors.Wrapf(err, "failed to push to git remote")
 	}
 	debug.Log("synced with remote")
-	//out.Green(ctx, "Pushed changes to remote")
 	return nil
 }
--- a/internal/action/audit.go
+++ b/internal/action/audit.go
@@ -31,7 +31,7 @@ func (s *Action) Audit(c *cli.Context) error {
 	list := t.List(tree.INF)
 
 	if len(list) < 1 {
-		out.Yellow(ctx, "No secrets found")
+		out.Print(ctx, "No secrets found")
 		return nil
 	}
 
--- a/internal/action/binary.go
+++ b/internal/action/binary.go
@@ -267,7 +267,7 @@ func (s *Action) Sum(c *cli.Context) error {
 
 	h := sha256.New()
 	_, _ = h.Write(buf)
-	out.Yellow(ctx, "%x", h.Sum(nil))
+	out.Print(ctx, "%x", h.Sum(nil))
 
 	return nil
 }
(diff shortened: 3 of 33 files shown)


internal/store/leaf/write.go
```diff
@@ -97,6 +97,6 @@ func (s *Store) gitCommitAndPush(ctx context.Context, name string) error {
 		return errors.Wrapf(err, "failed to push to git remote")
 	}
 	debug.Log("synced with remote")
-	//out.Green(ctx, "Pushed changes to remote")
+	//out.Print(ctx, "Pushed changes to remote")
```
AnomalRoil, line 100: Should we remove this or uncomment it?

AnomalRoil (approved):
This also fixes #1464

internal/store/leaf/write.go
dominikschulz: Removed.

AnomalRoil (approved):

PR #1747, opened 2021-01-22
Confirm passphrase when saving age keyring

dominikschulz:

### DIFF
--- a/pkg/pinentry/pinentry.go
+++ b/pkg/pinentry/pinentry.go
@@ -12,6 +12,7 @@ import (
 	"os/exec"
 	"strings"
 
+	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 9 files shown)


pkg/pinentry/pinentry.go
```diff
@@ -125,13 +126,15 @@ func (c *Client) GetPin() ([]byte, error) {
 	if bytes.HasPrefix(buf, []byte("OK")) {
 		return nil, nil
 	}
-	if bytes.HasPrefix(buf, []byte("S PASSWORD_FROM_CACHE")) {
```
AnomalRoil, line 128: Not sure why you'd want to have a for loop instead of the more specific if statement here.
https://git.gnupg.org/cgi-bin/gitweb.cgi?p=pinentry.git;a=blob;f=doc/pinentry.texi;h=a4acd8b9f27bf2cdbd50aead2757872978bbe9f6;hb=HEAD#l483

AnomalRoil (approved):

pkg/pinentry/pinentry.go
dominikschulz: Since I'm not sure how many how many "headers" are being returned.

Maybe I'm reading the spec wrong, but this should work either way?

AnomalRoil: Yeah I think it should, but it would misinterpret a status message "S PIN_REPEATED"

AnomalRoil: Unless we want to do the same in case of such a status, actually. I haven't dug too much into how pinentry works

PR #1749, opened 2021-01-22
Attempt to handle non-master git default branches

### LINKED ISSUE #1742: Handle non-master default git branches
gopass has the hardcoded expectation that the git default branch is always called `master`.
This default is changing upstream and we should support whatever users decide their default branch should be called.

dominikschulz:
Fixes #1742

### DIFF
--- a/internal/action/rcs.go
+++ b/internal/action/rcs.go
@@ -7,6 +7,7 @@ import (
 	"github.com/gopasspw/gopass/internal/backend"
 	"github.com/gopasspw/gopass/internal/cui"
 	"github.com/gopasspw/gopass/internal/out"
+	si "github.com/gopasspw/gopass/internal/store"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/termio"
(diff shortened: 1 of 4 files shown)


internal/action/rcs.go
```diff
@@ -136,10 +137,10 @@ func (s *Action) RCSPush(c *cli.Context) error {
 	origin := c.Args().Get(0)
 	branch := c.Args().Get(1)
 
-	if origin == "" || branch == "" {
-		return ExitError(ExitUsage, nil, "Usage: %s git push <ORIGIN> <BRANCH>", s.Name)
-	}
 	if err := s.Store.RCSPush(ctx, store, origin, branch); err != nil {
+		if errors.Is(err, si.ErrGitNoRemote) {
+			return nil
```
AnomalRoil, line 142: We might want to log that out occurred, no?

internal/backend/storage/gitfs/git.go
```diff
@@ -235,7 +235,7 @@ func (g *Git) defaultRemote(ctx context.Context, branch string) string {
 func (g *Git) defaultBranch(ctx context.Context) string {
 	out, _, err := g.captureCmd(ctx, "defaultBranch", "rev-parse", "--abbrev-ref", "HEAD")
 	if err != nil || string(out) == "" {
-		return "master"
+		return "main"
```
AnomalRoil, line 238: Why change the default?

AnomalRoil (approved):

internal/action/rcs.go
dominikschulz: Why bother if no remote is configured?
But yes, we could log this.

internal/backend/storage/gitfs/git.go
dominikschulz: See e.g. https://github.com/github/renaming for context.

PR #1750, opened 2021-01-22
Replace github.com/pkg/errors by stdlib errors

dominikschulz:
RELEASE_NOTES=n/a


internal/store/root/init.go
```diff
@@ -2,22 +2,21 @@ package root
...
-	"github.com/pkg/errors"
 )
 
 // IsInitialized checks on disk if .gpg-id was generated and thus returns true.
 func (r *Store) IsInitialized(ctx context.Context) (bool, error) {
 	if r.store == nil {
 		debug.Log("initializing store and possible sub-stores")
 		if err := r.initialize(ctx); err != nil {
-			return false, errors.Wrapf(err, "failed to initialized stores: %s", err)
+			return false, fmt.Errorf("failed to initialized stores: %w", err)
```
AnomalRoil, line 19: typo: "failed to initialize stores: %w"

AnomalRoil (approved):
Really good idea, I was thinking about it when I coded the key-values code and saw we where using `errors` where `fmt.Errorf` was enough.

Also we might want to once have a discussion about third party dependencies and security. Maybe we should be using Dependabot too, wdyt?


Why the changed images at the beginning without any change to a doc file? Were these uncommitted changes from another commit or are they meant to be here?

dominikschulz:
I'd like to reduce the use of (unnecessary) external deps. Not sure about dependabot. What does it do?

The images shouldn't be changed here. Will revert them.

AnomalRoil:
Dependabot would submit automatically PRs when it detects that one of our dependencies as been upgraded, it allows us to have a clear trace of the dependencies updates and might later allow us to dig further into third party dep security/hardening, since supply chain attacks are becoming more and more of a concern lately in the field.

dominikschulz:
Sounds great.
Could you help to set it up?

AnomalRoil:
Yup, I'll dig into it in its own issue then

PR #1758, opened 2021-01-25
Use gpgconf(1) to locate the gpg binary

### LINKED ISSUE #1757: Searching path for `gpg` does not respect gnupg/gnupg2 side-by-side installation
### Summary

Many Linux distributions install GnuPG 2 as `/usr/bin/gpg`, but this is not guaranteed. Some offer legacy GnuPG 1 as `gpg` and install v2 as `gpg2`. The GnuPG 2 configure script even provides a flag, `--enable-gpg-is-gpg2`, to accommodate this use case.

Because I suspect most people would prefer defaulting to GnuPG2 when it is available, relying on a path search for `gpg` produces unexpected behavior when distributions install both versions side by side. Regardless of a preference for v1 or v2, the behavior will be unexpected if a user installs only v2 as `gpg2` and `gopass` is unable to find the executable it needs.

The right way to discover the path to GnuPG 2 is to query its `gpgconf` utility, which provides easily parsable output that provides the paths to components of the installation:
```
% gpgconf
gpg:OpenPGP:/usr/bin/gpg2
gpg-agent:Private Keys:/usr/bin/gpg-agent
scdaemon:Smartcards:/usr/libexec/scdaemon
gpgsm:S/MIME:/usr/bin/gpgsm
dirmngr:Network:/usr/bin/dirmngr
pinentry:Passphrase Entry:/usr/bin/pinentry
```

### Steps To Reproduce
Option 1:
1. Install only GnuPG 2 on a distribution that calls its main executable `gpg2`.

Option 2:
1. Install GnuPG 1 *and* 2 on the same distribution.

### Expected behavior
When only GnuPG 2 is available and is called `gpg2`, `gopass` should find the executable. When both v1 and v2 are available, `gopass` should probably prefer v2 unless the user explicitly overrides the preference.

### Environment
Void Linux is one such distribution that has a GnuPG 1 executable at `/usr/bin/gpg` and a GnuPG 2 executable at `/usr/bin/gpg2`.

Any version of `gopass` that includes commit https://github.com/gopasspw/gopass/commit/004eaa4a4c32fe6468404210677f06e9e30e8047 will display the unexpected behavior. I suspect older versions will also show unexpected behavior.

### Recommendations
The process for discovering GnuPG should probably go as follows:
1. If `gpgconf` exists in the path:
    a. Invoke `gpgconf`, capture output
    b. Parse the path to `gpg` as the third field in the record whose first field is `gpg`.
2. Otherwise, use current behavior of looking for `gpg` in the path.

dominikschulz:
Attempt to use gpgconf(1) for locating the path to the GPG binary.

Fixes #1757

RELEASE_NOTES=[ENHANCEMENT] Use gpgconf to the the gpg binary

### DIFF
--- a/internal/backend/crypto/gpg/cli/binary_others.go
+++ b/internal/backend/crypto/gpg/cli/binary_others.go
@@ -4,11 +4,25 @@ package cli
 
 import (
 	"os/exec"
+
+	"github.com/gopasspw/gopass/pkg/debug"
+	"github.com/gopasspw/gopass/pkg/pinentry/gpgconf"
 )
 
 func detectBinary(name string) (string, error) {
-	if name == "" {
-		name = "gpg"
+	// user supplied binaries take precedence
+	if name != "" {
+		return exec.LookPath(name)
 	}
-	return exec.LookPath(name)
+	// try to get the proper binary from gpgconf(1)
+	p, err := gpgconf.Path("gpg")
+	if err != nil || p == "" {
+		debug.Log("gpgconf failed, falling back to path lookup: %q", err)
+		// otherwise fall back to the default and try
+		// to look up "gpg"
+		return exec.LookPath("gpg")
+	}
+
+	debug.Log("gpgconf returned %q for gpg", p)
+	return p, nil
 }
--- a/pkg/pinentry/gpgconf/gpgconf.go
+++ b/pkg/pinentry/gpgconf/gpgconf.go
@@ -1,0 +1,39 @@
+package gpgconf
+
+import (
+	"bufio"
+	"bytes"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/gopasspw/gopass/pkg/debug"
+)
+
+// Path returns the path to a GPG component
+func Path(key string) (string, error) {
+	buf := &bytes.Buffer{}
+	cmd := exec.Command("gpgconf")
+	cmd.Stdout = buf
+	cmd.Stderr = os.Stderr
+
+	debug.Log("%s %+v", cmd.Path, cmd.Args)
+	if err := cmd.Run(); err != nil {
+		return "", err
+	}
+
+	key = strings.TrimSpace(strings.ToLower(key))
+	sc := bufio.NewScanner(buf)
+	for sc.Scan() {
+		p := strings.Split(strings.TrimSpace(sc.Text()), ":")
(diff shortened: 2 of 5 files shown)


internal/backend/crypto/gpg/cli/binary_others.go
```diff
@@ -4,11 +4,23 @@ package cli
...
 
 func detectBinary(name string) (string, error) {
-	if name == "" {
-		name = "gpg"
+	// user supplied binaries take precedence
+	if name != "" {
+		return exec.LookPath(name)
+	}
+	// try to get the proper binary from gpgconf(1)
+	if cfg, err := gpgConf(); err == nil {
```
AnomalRoil, line 17: We might want to log that error if we fallback to the old method instead.

AnomalRoil (approved):
We might want to also use gpgconf to get the `pinentry` binary, no?
https://github.com/gopasspw/gopass/blob/10e2e0ece4717a1be986ed8cb076d7da08d46486/pkg/pinentry/pinentry_others.go#L7

It might allow us to remove the special case for Darwin: https://github.com/gopasspw/gopass/blob/10e2e0ece4717a1be986ed8cb076d7da08d46486/pkg/pinentry/pinentry_darwin.go#L7

internal/backend/crypto/gpg/cli/binary_others.go
```diff
@@ -4,11 +4,24 @@ package cli
...
-		name = "gpg"
+	// user supplied binaries take precedence
+	if name != "" {
+		return exec.LookPath(name)
+	}
+	// try to get the proper binary from gpgconf(1)
+	if p, err := gpgconf.Path("gpg"); err == nil && p != "" {
+		debug.Log("gpgconf returned %q for gpg", p)
+		return p, nil
+	} else {
```
AnomalRoil, line 21: This will make our CI complain that there is a else statement before a return statement and thus it could be dropped.

AnomalRoil (dismissed):
LGTM

dshynkev:
Thanks for implementing this! Is there any chance 1.11.1 containing this commit will get tagged soon? Just asking to know whether to patch 1.11.0 or wait a bit.

dominikschulz:
I guess we're almost ready for 1.11.1.
One more PR waiting to get in and if no blockers show up we could tag it anytime.

PR #1759, opened 2021-01-25
Warn if throw-keyids is set

### LINKED ISSUE #1756: RFE: detect when encryption recipients aren't being set
### Summary

As reported in https://github.com/gopasspw/gopass/issues/1083, it is possible for settings in `gpg.conf` to result encryption recipients not being recorded for gopass entries.  While this doesn't stop gopass from working with gpg, it does break other use cases.  Furthermore, gopass doesn't currently detect that attempts to fix/repair the password entries are failing.

### Steps To Reproduce
* Set `throw-keyids` in your `gpg.conf`. 

```
~/.gnupg/gpg.conf
...
# Disable putting recipient key IDs into messages
throw-keyids
```
* Create a new password entry
* run `gopass fsck --decrypt`
* run `gopass fsck --decrypt`

### Expected behavior
For gopass to detect that attempts to include the encryption recipients are failing.  When detected, possibly mention the problematic `gpg.conf` setting.

### Environment

- OS: Arch Linux
- OS version: rolling
- gopass Version: 1.11.0
- Installation method: Arch repository

### Additional context
None

dominikschulz:
Fixes #1756

### DIFF
--- a/internal/backend/crypto/gpg/cli/gpg.go
+++ b/internal/backend/crypto/gpg/cli/gpg.go
@@ -33,6 +33,7 @@ type GPG struct {
 	pubKeys   gpg.KeyList
 	privKeys  gpg.KeyList
 	listCache *lru.TwoQueueCache
+	throwKids bool
 }
 
 // Config is the gpg wrapper config
@@ -56,8 +57,9 @@ func New(ctx context.Context, cfg Config) (*GPG, error) {
 	}
 
 	g := &GPG{
-		binary: "gpg",
-		args:   append(defaultArgs, cfg.Args...),
+		binary:    "gpg",
+		args:      append(defaultArgs, cfg.Args...),
+		throwKids: fileContains(gpgConfigLoc(), "throw-keyids"),
 	}
 
 	debug.Log("initializing LRU cache")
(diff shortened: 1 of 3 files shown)


AnomalRoil:
IMO this should be documented somewhere, because people that do not include key ids are typically doing it on purpose to try and stay anonymous and we don't want to undermine their effort if they think it is important for them, do we?

dominikschulz:
Good point, yes.
But this only affects gopass secrets.
Yes, people might want to apply this to secrets as well, but it just doesn't work well with our expectations.

AnomalRoil:
Well I disagree, this works well with Gopass since GPG will simply try ever secret key in the key ring during a decryption attempt if this option was used upon encryption.

So Gopass itself is not hindered in anyway by the throw-id setting.

dominikschulz:
The point is that some operations won't work as expected.

With age it's similar and there is not option to disable that.

So maybe we should only warn instead?
Or just ignore this setting and let the user figure it out?

dominikschulz:
Updated the PR to only warn about throw-keyids, not override it.

internal/backend/crypto/gpg/cli/gpg.go
```diff
@@ -33,6 +33,7 @@ type GPG struct {
 	pubKeys   gpg.KeyList
 	privKeys  gpg.KeyList
 	listCache *lru.TwoQueueCache
+	throwKids bool
```
AnomalRoil, line 36: Poor kids 😆

AnomalRoil (approved):

internal/backend/crypto/gpg/cli/gpg.go
dominikschulz: No kids were harmed in making this PR ;)

PR #1773, opened 2021-02-04
Patch goreleaser action

AnomalRoil:
RELEASE_NOTES=n/a

Signed-off-by: Yolan Romailler <yolan@romailler.ch>

### DIFF
--- a/helpers/changelog/main.go
+++ b/helpers/changelog/main.go
@@ -14,6 +14,12 @@ func main() {
 	}
 	defer fh.Close()
 
+	fw, err := os.Create("RELEASE_NOTES")
+	if err != nil {
+		panic(err)
+	}
+	defer fw.Close()
+
 	s := bufio.NewScanner(fh)
 	var in bool
 	for s.Scan() {
@@ -24,6 +30,10 @@ func main() {
 			}
 			in = true
 		}
-		fmt.Println(line)
+
+		_, err := fmt.Fprintln(fw, line)
+		if err != nil {
+			panic(err)
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


helpers/changelog/main.go
```diff
@@ -24,6 +29,11 @@ func main() {
...
 			in = true
 		}
-		fmt.Println(line)
+
+		_, err := fw.WriteString(line + "\n")
+		if err != nil {
+			panic(err)
+		}
 	}
+	fw.Sync()
```
dominikschulz, line 38: I've never used this method before. I don't think it's necessary.
But you can leave it as is if you want.

helpers/changelog/main.go
```diff
@@ -24,6 +29,11 @@ func main() {
 			}
 			in = true
 		}
-		fmt.Println(line)
+
+		_, err := fw.WriteString(line + "\n")
```
dominikschulz, line 33: I'd rather use `fmt.Fprintln(fw, line)`, that's more common.

helpers/changelog/main.go
```diff
@@ -14,6 +13,12 @@ func main() {
 	}
 	defer fh.Close()
 
+	fw, err := os.Create("RELEASE_NOTES")
```
dominikschulz, line 16: Hmm, if someone runs this helper locally in a checkout they might end up committing this along their changes.
But on the other hand I'm happy to argue if someone runs these helpers they should know what they are doing.

dominikschulz (approved):
LGTM. Some nits, but nothing blocking.

helpers/changelog/main.go
AnomalRoil: Yeah, let's try like this and see if we can figure out another way later.

dominikschulz: sg

dominikschulz (approved):
LGTM

PR #1774, opened 2021-02-04
Fix GHA

AnomalRoil:
RELEASE_NOTES=n/a


dominikschulz (approved):
Oh. Wow. Not it get's a bit ugly.
But well ... if it works.

PR #1776, opened 2021-02-04
fix GHA GPG fingerprint

AnomalRoil:
RELEASE_NOTES=n/a


dominikschulz (approved):
LGTM

Why do we need the `id` setting here?

AnomalRoil:
Because we use it later:
`GPG_FINGERPRINT: ${{ steps.import_gpg.outputs.fingerprint }}`

dominikschulz:
Got it, thanks!

PR #1778, opened 2021-02-05
Fix mixed case keys

### LINKED ISSUE #1777: Using caps in Keys causes issues
### Summary
It used to work.

### Steps To Reproduce
Create a new secret `t1` containing:
```
secret
User: test
```

Display it (with `safecontent` set to `true`):
```
$ gopass show t1
Secret: t1

User:
```

with safecontent set to `false` it works (but adds a trailing newline):
```
Secret: t1

secret
User: test

```

Furthermore the key `User` is not recognized:
```
$ gopass show t1 user

Error: key not found in entry
```
Both with `User` and with `user` as the queried keys and both with and without `safecontent` enabled

### Expected behavior
It should just not display the "secret" if safecontent is set (and the unsafe keys, if these are used)

AnomalRoil:
RELEASE_NOTES=[BUGFIX] fixed mixed case keys for key-value, all keys are lower case now

Fixes #1777

### DIFF
--- a/internal/action/show.go
+++ b/internal/action/show.go
@@ -204,7 +204,7 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	// everything but the first line
 	if ctxutil.IsShowSafeContent(ctx) && !ctxutil.IsForce(ctx) {
 		var sb strings.Builder
-		for _, k := range sec.Keys() {
+		for i, k := range sec.Keys() {
 			sb.WriteString(k)
 			sb.WriteString(": ")
 			// check if this key should be obstructed
@@ -218,7 +218,10 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 				}
 				sb.WriteString(strings.Join(v, "\n"+k+": "))
 			}
-			sb.WriteString("\n")
+			// we only add a final new line if the body is non-empty
+			if sec.Body() != "" || i < len(sec.Keys())-1 {
+				sb.WriteString("\n")
+			}
 		}
 
 		sb.WriteString(sec.Body())
--- a/pkg/gopass/secrets/kv.go
+++ b/pkg/gopass/secrets/kv.go
@@ -82,15 +82,22 @@ func (k *KV) Bytes() []byte {
 	buf := &bytes.Buffer{}
 	buf.WriteString(k.password)
 	buf.WriteString("\n")
-	for _, key := range k.Keys() {
+	for ik, key := range k.Keys() {
 		sv, ok := k.data[key]
 		if !ok {
 			continue
 		}
-		for _, v := range sv {
+		for iv, v := range sv {
 			_, _ = buf.WriteString(key)
 			_, _ = buf.WriteString(": ")
 			_, _ = buf.WriteString(v)
+			// the last one shouldn't add a newline, it's handled below
+			if iv < len(sv)-1 {
+				_, _ = buf.WriteString("\n")
+			}
+		}
+		// we must only add a final newline if the body is non-empty
+		if k.body != "" || ik < len(k.Keys())-1 {
 			_, _ = buf.WriteString("\n")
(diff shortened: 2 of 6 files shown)


internal/action/show.go
```diff
@@ -158,7 +158,7 @@ func (s *Action) showHandleOutput(ctx context.Context, name string, sec gopass.S
 		return nil
 	}
 
-	ctx = out.WithNewline(ctx, ctxutil.IsTerminal(ctx))
+	ctx = out.WithNewline(ctx, false && ctxutil.IsTerminal(ctx))
```
dominikschulz, line 161: Why are you always disabling the newline output?
If that's correct you could remove the second part of the condition.

But I guess this is not intentional?

internal/action/show.go
```diff
@@ -204,7 +204,7 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	// everything but the first line
 	if ctxutil.IsShowSafeContent(ctx) && !ctxutil.IsForce(ctx) {
 		var sb strings.Builder
-		for _, k := range sec.Keys() {
+		for l, k := range sec.Keys() {
```
dominikschulz, line 207: Usually one would expect n or i for the counter. l as in lines?

internal/action/show.go
```diff
@@ -218,7 +218,9 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 				}
 				sb.WriteString(strings.Join(v, "\n"+k+": "))
 			}
-			sb.WriteString("\n")
+			if sec.Body() != "" || l < len(sec.Keys())-1 {
```
dominikschulz, line 221: Ok, I get what you're doing but I think this is suble enough to ask for a short comment.

pkg/gopass/secrets/kv.go
```diff
@@ -82,15 +82,20 @@ func (k *KV) Bytes() []byte {
 	buf := &bytes.Buffer{}
 	buf.WriteString(k.password)
 	buf.WriteString("\n")
-	for _, key := range k.Keys() {
+	for lk, key := range k.Keys() {
```
dominikschulz, line 85: I expect this to be called `ik`

pkg/gopass/secrets/kv.go
```diff
@@ -82,15 +82,20 @@ func (k *KV) Bytes() []byte {
...
 		sv, ok := k.data[key]
 		if !ok {
 			continue
 		}
-		for _, v := range sv {
+		for lv, v := range sv {
```
dominikschulz, line 90: `iv`

pkg/gopass/secrets/kv.go
```diff
@@ -82,15 +82,20 @@ func (k *KV) Bytes() []byte {
...
 			_, _ = buf.WriteString(key)
 			_, _ = buf.WriteString(": ")
 			_, _ = buf.WriteString(v)
+			if lv < len(sv)-1 {
+				_, _ = buf.WriteString("\n")
+			}
+		}
+		if k.body != "" || lk < len(k.Keys())-1 {
```
dominikschulz, line 98: Short comment, please.

pkg/gopass/secrets/kv.go
```diff
@@ -202,6 +207,8 @@ func ParseKV(in []byte) (*KV, error) {
 		for i, part := range parts {
 			parts[i] = strings.TrimSpace(part)
 		}
+		// we only store lower caps keys for KV
```
dominikschulz, line 210: `lower case`, not `lower caps`?

dominikschulz (changes requested):
Looks good, only some minor comments.

internal/action/show.go
AnomalRoil: Ups, leftover from debugging!

dominikschulz (approved):
LGTM

PR #1792, opened 2021-02-12
Remember when fsck was last run and remind after 90d

dominikschulz:
This adds a reminder when fsck was not run within the last 90d (or
never).

RELEASE_NOTES=[ENHANCEMENT] Remind to run gopass fsck after 90d

### DIFF
--- a/internal/action/reminder.go
+++ b/internal/action/reminder.go
@@ -1,0 +1,39 @@
+package action
+
+import (
+	"context"
+	"os"
+
+	"github.com/gopasspw/gopass/internal/out"
+	"github.com/gopasspw/gopass/pkg/ctxutil"
+)
+
+func (s *Action) printReminder(ctx context.Context) {
+	if !ctxutil.IsInteractive(ctx) {
+		return
+	}
+	if !ctxutil.IsTerminal(ctx) {
+		return
+	}
+	if sv := os.Getenv("GOPASS_NO_REMINDER"); sv != "" {
+		return
+	}
+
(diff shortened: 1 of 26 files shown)


AnomalRoil (changes requested):
Let's have an extra case in `printReminder` that would allow us to disable this feature through an env var, shall we?

dominikschulz:
Yes, sounds good. I was worried that some people might feel annoyed, too.

internal/action/reminder.go
```diff
@@ -0,0 +1,34 @@
...
+	if !ctxutil.IsTerminal(ctx) {
+		return
+	}
+	if sv := os.Getenv("GOPASS_NO_REMINDER"); sv != "" {
+		return
+	}
+
+	if s.rem.Overdue("fsck") {
+		out.Notice(ctx, "You haven't run 'gopass fsck' in a while.")
+		return
```
AnomalRoil, line 24: Note that a return statement here means that we will warn only about fsck even if we should also warn about updates and audit.

People might ignore `fsck` but be more interested in the `update` one I guess.

AnomalRoil (dismissed):

AnomalRoil:
Please solve the conflicts too :)

internal/action/reminder.go
dominikschulz: Actually this is intentional! I did only want to print one notice per day.
Added a comment and put update first.

AnomalRoil (approved):

PR #1794, opened 2021-02-13
Add suffix f to formating out methods

### LINKED ISSUE #1793: Problem with passwords containing characters matching fmt format specifiers
### Summary

I have a couple of passwords that have character groups matching the fmt package format specifiers. 
As an example there's `%3U`, which when using go show appears as `!U(MISSING)`.

This problem has appeared with version 1.12.0

### Steps To Reproduce

Create a password containing a string matching a fmt format specifier.

When showing the password with gopass it will be treated as a format specifier that's missing its value.

### Expected behavior

The password should be shown as is.

### Environment

- OS: Linux - Archlinux
- OS version: `Linux myhostname 5.10.15-arch1-1 #1 SMP PREEMPT Wed, 10 Feb 2021 18:32:40 +0000 x86_64 GNU/Linux`
- gopass Version: 
```
gopass 1.12.0 (2021-02-11 21:15:26) go1.15.8 linux amd64
<root>     - gpg 2.2.27 - git 2.30.1
Available Storage Backends: fs, gitfs
```
- Installation method: From Archlinux's community repository

dominikschulz:
This commit renames the existing out methods that expect
a format string to include the common f suffix and introduces
new out methods without this suffix that don't accept a
format string or variadic arguments.

Note: Everything outside of `internal/out/` is based on a
search and replace operation and doesn't need much
attention.

Fixes #1793

RELEASE_NOTES=n/a

### DIFF
--- a/internal/action/aliases.go
+++ b/internal/action/aliases.go
@@ -12,7 +12,7 @@ import (
 
 // AliasesPrint prints all cofigured aliases
 func (s *Action) AliasesPrint(c *cli.Context) error {
-	out.Print(c.Context, "Configured aliases:")
+	out.Printf(c.Context, "Configured aliases:")
 	aliases := pwrules.AllAliases()
 	keys := make([]string, 0, len(aliases))
 	for k := range aliases {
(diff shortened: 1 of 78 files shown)


AnomalRoil (approved):
Welp, we should have had a test case for this, I never thought it could happen. Would you mind adding a test for show with a password that would trigger this bug?

dominikschulz:
Yes, I've added that. And had to fix some unrelated issues as well (e.g. one of the parsing disabled tests did actually expect the wrong output).

AnomalRoil (approved):
Perfect

AnomalRoil:
Can you solve the conflict and merge?

dominikschulz:
Done

PR #1828, opened 2021-03-06
Add gopass ln

### LINKED ISSUE #1820: [Feature Request] gopass ln command
It would be convenient to have the option to create symlinks inside the password store.
This would simplify maintenance, i.e. one could have github.com/accountname this will not be used by the git-credentials helper such that when changing secrets one would have to edit two files.
Creating the option to symlink the secret to the correct location would simplify tasks like the outlined one.

(I tried simply creating a symlink for the secret in the password-store but this did not work.)

dominikschulz:
This commit adds a new subcommand to support symlinks within the same
mount. The command is intentionally still hidden since the support
isn't fully fleshed out, yet. Neither audit nor list know of links
and cross mount symlinks aren't supported either (and maybe never will).

RELEASE_NOTES=[ENHANCEMENT] Add gopass ln

Fixes #1820

### DIFF
--- a/docs/commands/link.md
+++ b/docs/commands/link.md
@@ -1,0 +1,28 @@
+# `link` command
+
+The `link` (or `ln`) command is used to create a symlink from one secret in a
+store to a target in the same store.
+
+Note: Symlinks across different stores / mounts are currently not supported!
+
+Note: `audit` and `list` do not recognize symlinks, yet. They will treat
+symlinks as regular (different) entries.
+
+## Synopsis
+
+```
+$ gopass ln foo/bar bar/baz
+$ gopass show foo/bar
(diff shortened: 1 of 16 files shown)


AnomalRoil:
This one is going to take me some time to review and test locally on Windows, Macos and Linux

dominikschulz:
Yes, makes sense. Thanks for taking the time! I'll also give it some more testing on my end.

dominikschulz:
Actually I noticed some pre-existing bugs in the tree implementation that impacted (at least) the audit command with any filter argument. This surfaced during my tests of this PR so I've included them here.

dominikschulz:
My tests look good and I want to have the bugfixes merged.

PR #1834, opened 2021-03-12
Fixing typos

AnomalRoil:
Thanks to Roel Van de Paar


dominikschulz (approved):
Very good improvements. Thanks to both of you!

PR #1845, opened 2021-03-17
Install manpage when using `make install`

silkeh:
Add an `install-man` make target to install the manpage, and include it in the `install` target.

RELEASE_NOTES=[ENHANCEMENT] Install manpage when using `make install`

Signed-off-by: Silke Hofstra <silke@slxh.eu>


dominikschulz (approved):
Thanks!

PR #1865, opened 2021-03-22
Add getting started docs

dominikschulz:
This commit adds a getting started section to the readme and improves
the gopass clone UX a little it. It also fixes the call depth for nested
debug.Log invocations (e.g. during ExitError) and adds debug logging
for every out invocation.

RELEASE_NOTES=n/a

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 # gopass
 
 [![Build Status](https://travis-ci.org/gopasspw/gopass.svg?branch=master)](https://travis-ci.org/gopasspw/gopass)
+[![Packaging status](https://repology.org/badge/tiny-repos/gopass.svg)](https://repology.org/project/gopass/versions)
 [![Go Report Card](https://goreportcard.com/badge/github.com/gopasspw/gopass)](https://goreportcard.com/report/github.com/gopasspw/gopass)
 [![MIT License](https://img.shields.io/badge/license-MIT-blue.svg)](https://github.com/gopasspw/gopass/blob/master/LICENSE)
 [![Github All Releases](https://img.shields.io/github/downloads/gopasspw/gopass/total.svg)](https://github.com/gopasspw/gopass/releases)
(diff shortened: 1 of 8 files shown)


internal/out/print.go
```diff
@@ -34,6 +35,7 @@ func Printf(ctx context.Context, format string, args ...interface{}) {
 	if ctxutil.IsHidden(ctx) {
 		return
 	}
+	debug.LogN(1, format, args...)
```
AnomalRoil, line 38: I'm not too fond of this since it would add plaintext passwords to the debug log.

README.md
```diff
@@ -64,10 +65,66 @@ WARNING: Please prefer releases, unless you want to contribute to the
 development of gopass. The master branch might not be very well tested and
 can contain breaking changes without further notice.
 
+## Getting Started
+
+Either initialize a new git repository and clone an existing one.
```
AnomalRoil, line 70: "Either ... or"

README.md
```diff
@@ -64,10 +65,66 @@ WARNING: Please prefer releases, unless you want to contribute to the
...
+
+### New password store
+
+```bash
```
AnomalRoil, line 74: not sure bash is the best here: the output gets colored weirdly because of it.

README.md
```diff
@@ -64,10 +65,66 @@ WARNING: Please prefer releases, unless you want to contribute to the
...
+Configuring the git remote ...
+Please enter the git remote for your shared store []: git@gitlab.example.org:john/passwords.git
+✅ Configured
+```
+
+Hint: `gopass setup` will use `gpg` encryption and `git` storage by default.
+
+### Existing password store
+
+```bash
```
AnomalRoil, line 100: same ^

docs/setup.md
```diff
@@ -334,7 +334,36 @@ The new binary can be downloaded from the latest
...
+
+```bash
+$ gopass setup --crypto gpg --storage gitfs # used by default
+[...]
+# provide an existing, empty git remote, e.g. git@gitlab.example.org:john/passwords.git
+```
+
+#### Existing password store with git
+
+If you have created a password store with git `gopass` can easily clone it.
```
AnomalRoil, line 360: "with git, `gopass`" missing a comma

internal/out/print.go
dominikschulz: This shouldn't change what we log. It's supposed to only fix the originator shown in the log.
Why do you believe this will make us start logging plaintext passwords?

(Note: Yes, the debug can contain these, and yes we might want to change that. But this shouldn't be affected by this PR).

AnomalRoil: Because of https://github.com/gopasspw/gopass/blob/456c18a1c2f751746ce5900107fa1f62dbd4332c/internal/action/show.go#L174-L175
and `Print` is just using `Printf` behind the hood.

dominikschulz: Ack. This needs to be changed. But it's not the only place where we log secrets. The parser has a bunch of debug statements as well.

PR #1885, opened 2021-03-26
Adding postrel GHA

AnomalRoil:
Adding a post-release GHA

This is a WIP, to be refined because the current post-release helper isn't working without a specific folder structure

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,31 @@
+name: Post-release gopass deployement
+
+on:
+  release:
+    types:
+      - created
+
+jobs:
+  CD:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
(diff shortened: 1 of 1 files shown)


dominikschulz:
Do you have any idea how the folder structure should change?

AnomalRoil:
I was thinking that we have really two options:

- one GHA here that would be pulling all the right repo in one folder and then we can have the folder structure as it is currently, or tweak it as we see fit
- create a new "post-release packaging" repo that would have all other repo we need as submodules and trigger a GHA there upon completion of our release GHA here, and then we should use that folder structure in our post-release helper.

What would you prefer ?

dominikschulz:
@AnomalRoil Do you have plans to continue working on this?

I don't see how this could possibly work right now.
But feel free to prove me wrong.

AnomalRoil:
I'd love to have a fully automated release process, but I think this isn't the right approach. Instead we should have a GHA running on tags that does all the rest.

PR #1886, opened 2021-03-26
Avoid logging credentials

### LINKED ISSUE #1883: Do not log secrets
When debug logging is enabled we log secrets.

This should not happen. At least not by default.

dominikschulz:
This commit adds filtering to avoid logging credentials in the debug
logs. If logging of credentials, e.g. for debugging secret parsers,
is required GOPASS_DEBUG_LOG_SECRETS can be set to an non empty
string to enable logging of secrets.

Fixes #1883

RELEASE_NOTES=[BUGFIX] Avoid logging credentials

### DIFF
--- a/pkg/gopass/secrets/secparse/parse.go
+++ b/pkg/gopass/secrets/secparse/parse.go
@@ -1,6 +1,7 @@
 package secparse
 
 import (
+	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/gopass"
 	"github.com/gopasspw/gopass/pkg/gopass/secrets"
@@ -16,7 +17,8 @@ func Parse(in []byte) (gopass.Secret, error) {
 		debug.Log("parsed as MIME: %+v", s)
 		return s, nil
 	}
-	debug.Log("failed to parse as MIME: %s", err)
+	debug.Log("failed to parse as MIME: %s", out.Secret(err.Error()))
+
 	if _, ok := err.(*secrets.PermanentError); ok {
 		return secrets.ParsePlain(in), err
 	}
@@ -25,14 +27,16 @@ func Parse(in []byte) (gopass.Secret, error) {
 		debug.Log("parsed as YAML: %+v", s)
 		return s, nil
 	}
-	debug.Log("failed to parse as YAML: %s\n%s", err, string(in))
+	debug.Log("failed to parse as YAML: %s\n%s", err, out.Secret(string(in)))
+
 	s, err = secrets.ParseKV(in)
 	if err == nil {
 		debug.Log("parsed as KV: %+v", s)
 		return s, nil
 	}
 	debug.Log("failed to parse as KV: %s", err)
+
 	s = secrets.ParsePlain(in)
-	debug.Log("parsed as plain: %s", string(s.Bytes()))
+	debug.Log("parsed as plain: %s", out.Secret(s.Bytes()))
 	return s, nil
 }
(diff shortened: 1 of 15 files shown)


pkg/gopass/secrets/secparse/parse.go
```diff
@@ -25,14 +26,14 @@ func Parse(in []byte) (gopass.Secret, error) {
 		debug.Log("parsed as YAML: %+v", s)
 		return s, nil
 	}
-	debug.Log("failed to parse as YAML: %s\n%s", err, string(in))
+	debug.Log("failed to parse as YAML: %s\n%s", err, out.Secret(string(in)))
 	s, err = secrets.ParseKV(in)
 	if err == nil {
 		debug.Log("parsed as KV: %+v", s)
```
AnomalRoil, line 32: That seems like a lot of logging too, no?

AnomalRoil (changes requested):
I'm not sure if this one could be logging passwords too:
https://github.com/gopasspw/gopass/blob/9349415d4759e3aa754fa3c757e171da386b333c/pkg/gopass/secrets/secparse/mime.go#L26

The first line is usually the password, except for Mime, and we are running it first too in the parser:
https://github.com/gopasspw/gopass/blob/9349415d4759e3aa754fa3c757e171da386b333c/pkg/gopass/secrets/secparse/parse.go#L14

But I think it's currently logging all passwords upon parsing, isn't it?

pkg/gopass/secrets/secparse/parse.go
dominikschulz: Great catch! The MIME parser would indeed still leak secrets.

Error logging is quite verbose, yes. But it can be controlled with `GOPASS_DEBUG_FUNCS` and `GOPASS_DEBUG_FILES`.

AnomalRoil: I'm not sure how this is resolved? 

This is when `err == nil`, so we have the whole secret `s` that's logged here upon success.

pkg/gopass/secrets/secparse/parse.go
```diff
@@ -25,14 +27,16 @@ func Parse(in []byte) (gopass.Secret, error) {
 		debug.Log("parsed as YAML: %+v", s)
```
AnomalRoil, line 27: Here too actually, if there is no error, we log the secret, don't we?

pkg/gopass/secrets/secparse/parse.go
dominikschulz: But it's wrapped in an `out.Secret`. In https://github.com/gopasspw/gopass/pull/1886/files#diff-2769ebba150a804d5515d28f28b7a8e06f05d914e9781b4e58beac1dc4d2ed52R203 the string is rewritten to not contain ... anything.

pkg/gopass/secrets/secparse/parse.go
dominikschulz: No, `secrets.YAML` implements the `Safer` interface, thus it's being rewritting inside the debug package: https://github.com/gopasspw/gopass/pull/1886/files#diff-2769ebba150a804d5515d28f28b7a8e06f05d914e9781b4e58beac1dc4d2ed52R203

pkg/gopass/secrets/secparse/parse.go
AnomalRoil: Oh right. That's a nice way to do it.

AnomalRoil:
Is the `mime` parser logging resolved? I don't feel so, since it's the error itself that would contain the plaintest `line` that's probably going to be the password if it's not a Mime secret.

dominikschulz:
Yes, the MIME vector should be resolved, too. See https://github.com/gopasspw/gopass/pull/1886/files?diff=split&w=1#diff-a3f960f5e161571c71dc434416d9be495d7e0799c7528e088c213db56c28d2fcR20

I've tested it. I'm not sure if we missed any other places, but the most obvious ones should be addressed now.

Issue #1878, opened 2021-03-25
Can auto push be disabled for team sharing store


morningspace:
### Summary

It looks the current gopass behavior is that it will auto sync all the time where there is a change. However, in a team sharing scenario, I'd like to control that only privilege member is allowed to push to the remote, while others can only pull.

It appears I can achieve this by configure Collaborators & teams on GitHub side, i.e. grant read only permission, but the gopass will still keep auto pushing w/ errors.

dominikschulz:
Yes, this is true.

I have to admit I've never considered such a use case.

morningspace:
Thanks @dominikschulz for your prompt reply. Any chance that it could be considered. I think it's a common case for team sharing. I haven't checked the gopass code, but if that's something kind of easy to add. I'd be happy to spend some time to add that.

dominikschulz:
Maintaining config options is very tedious so I aim for as little config options as possible.

The AutoSync issue has been discussed before in #1567 and the general config handling approach in #1819 .
I understand your use case and I think this is something we could / should support, but I'm not yet sure what the least toilsome approach would be.

dominikschulz:
To support the use case (i.e. a read only password store client) we would need:

* a use case description in `docs/usecases/` describing the use case
* a way to disable autosync-on-writes (maybe, read-only clients shouldn't even attempt to write to their store)
* changes to `gopass sync` to only pull, never push changes
* an integration test

Please note that we don't want to add new features without a clear usecase definition. Otherwise we'll accidentally break these in the future (this has happened a few times before, because nobody remembers why a specific feature was as it was).

morningspace:
Sure, understood. What can I do or can help in this case? Maybe I can start w/ submitting a use case into `docs/usecases/`? Or if someone else should take that per the process.

> maybe, read-only clients shouldn't even attempt to write to their store
I think so. Since it's ready-only, there's no write and sync-on-write, but it needs pull to sync the remote changes from time to time.

@dominikschulz

dominikschulz:
If you could contribute a use case definition that would help.

Then we can figure out what else we'd need.

morningspace:
Sure, I will submit a PR later this week.

morningspace:
PR #1892 ready for review. @dominikschulz

AnomalRoil:
So, this would really require us to tackle #1567 as well, since this should probably be a per-store setting and not a global one, as said in your PR.

[...]

PR #1892, opened 2021-03-29
Proposed use case: Readonly Store

morningspace:
Background of this use case, see: #1878.

### DIFF
--- a/docs/usecases/readyonly-store.md
+++ b/docs/usecases/readyonly-store.md
@@ -1,0 +1,34 @@
+# Use case: Readonly Store
+
+## Summary
+
+Allow a password store or a set of sub-stores to be configured in readonly mode in team sharing scenario.
+
+## Background
+
+In a team sharing scenario that we share password store among team members, usually we want each member to be able to pull and push so that anyone can share secret data with others. However, typically in a large size team, we may want the sharing to be more restricted where only priviledged users are allowed to push secret data to the remote store while others can only pull from the remote store.
+
+The current gopass behavior is that it will auto sync (pull and push) betweeen local store and remote store any time when there is a change at local. This means anyone can push their personal data to the centrally controlled remote store which will be polluted with arbitrary data unexpectedly.
+
+To workaround this, we can configure "Collaborators & teams" on GitHub side to grant read only permission to those who do not necessarily need push, but the gopass on their machines will still keep auto pushing and prompt with errors which is annoying.
+
+## Proposal
+
+Ultimately it turns out that this scenario requires a feature such as a store in readonly mode, where people can configure their local store or a set of sub-stores in readonly mode, to disable the writes and the autosync-on-writes to the store, but they can still pull to sync the latest changes from the remote store. This is not a one-stop solution for the RBAC model of the team sharing store, because we still need GitHub to setup the store access at server-side, but it will provide better usage experience from gopass client side.
+
+Configuration examples:
+```console
+# To print the config
+$ gopass config readonly
+readonly: false
+# To setup the config
+$ gopass config readonly true
+readonly: true
+# To apply the config to a sub-store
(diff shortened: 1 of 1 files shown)


dominikschulz (commented):
I'm not yet sold on the specific implementation details, e.g. having a `readonly` option.
Maybe there is an easier way to support this use case.

docs/usecases/readyonly-store.md
```diff
@@ -0,0 +1,34 @@
...
+Configuration examples:
+```console
+# To print the config
+$ gopass config readonly
+readonly: false
+# To setup the config
+$ gopass config readonly true
+readonly: true
+# To apply the config to a sub-store
+$ gopass config readly true --store team-sharable
```
dominikschulz, line 28: This should read `gopass config readonly`.

But I don't think I want to bring back nested configs. They were very painful.
There needs to be a better way.

docs/usecases/readyonly-store.md
```diff
@@ -0,0 +1,34 @@
...
+
+The current gopass behavior is that it will auto sync (pull and push) betweeen local store and remote store any time when there is a change at local. This means anyone can push their personal data to the centrally controlled remote store which will be polluted with arbitrary data unexpectedly.
+
+To workaround this, we can configure "Collaborators & teams" on GitHub side to grant read only permission to those who do not necessarily need push, but the gopass on their machines will still keep auto pushing and prompt with errors which is annoying.
+
+## Proposal
+
+Ultimately it turns out that this scenario requires a feature such as a store in readonly mode, where people can configure their local store or a set of sub-stores in readonly mode, to disable the writes and the autosync-on-writes to the store, but they can still pull to sync the latest changes from the remote store. This is not a one-stop solution for the RBAC model of the team sharing store, because we still need GitHub to setup the store access at server-side, but it will provide better usage experience from gopass client side.
+
+Configuration examples:
```
dominikschulz, line 19: I'm not convinced this is the best way to do it, yet.

Maybe we should drop the `Configuration examples` section and have a more open ended `Usage examples` one?

docs/usecases/readyonly-store.md
morningspace: My bad, let me fix it. Re: nested configs, do you mean the per-store config? I wonder why that's painful. Understand that can introduce complexity such as the possible conflict between per-store config and global config.

docs/usecases/readyonly-store.md
morningspace: Sure, I'm open to make any change as long as it can address the issue nicely.

morningspace:
> Maybe there is an easier way to support this use case.

Yeah, I hope so too. Right now, I am not sure whether there's simpler option. If we consider this as a configurable thing, then I think no matter what name it is, readonly, offline, or whatever else, it is essentially a way to control the writes and autosync-on-writes behavior conditionally. Also, it seems that has to be a per-store config, because practically, to make the local store completely readonly is not a good idea as people may have both private and sharable stuff stored on local.

On the other side, if we can consider it as a non-configurable thing, maybe a better idea, since people can explicitly turn off the readonly config for a team sharing store which will bring back the painful things again. With that, it looks to declare a store to be readonly during the init phase would make much more sense. However, the store owners or privileged users still need write access to the store. This makes the idea slip to some kind of RBAC model which is also something I may want to avoid since it may also be complex or even more complex than a config option.

dominikschulz:
I'd still want to merge this and eventually support the use case but I don't like to proposed configuration section, yet.

So either @morningspace changes this or we merge it as is and update it afterwards.

AnomalRoil:
@dominikschulz I think you can edit it by sending extra commits directly on the branch morningspace/gopass, and it will accept commits from the original repo maintainers.

dominikschulz (approved):
LGTM

PR #1907, opened 2021-04-17
Do not recurse when deleting a key

### LINKED ISSUE #1906: `gopass rm -r folder foo` deletes folder right away (non interactive)
### Summary

`gopass rm -r folder` ask for confirmation but `gopass rm -r folder foo` does not.

### Steps To Reproduce

`gopass rm -r folder foo`

### Expected behavior

`gopass rm` without the flag `-f` should reliably prompt for confirmation before deleting anything.

### Environment

- OS: Arch Linux
- gopass Version: 1.12.5
- Installation method: OS package manager pacman

Thank you very much for the work you put into this project.

dominikschulz:
Fixes #1906

RELEASE_NOTES=[BUGFIX] Do not recurse with a key

### DIFF
--- a/internal/action/delete.go
+++ b/internal/action/delete.go
@@ -29,6 +29,10 @@ func (s *Action) Delete(c *cli.Context) error {
 	// specifying a key is optional
 	key := c.Args().Get(1)
 
+	if recursive && key != "" {
+		return ExitError(ExitUsage, nil, "Can not use -r with a key. Invoke delete either with a key or with -r")
+	}
+
 	if !force { // don't check if it's force anyway
 		recStr := ""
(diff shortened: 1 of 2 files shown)


internal/action/delete.go
```diff
@@ -39,7 +43,7 @@ func (s *Action) Delete(c *cli.Context) error {
 		}
 	}
 
-	if recursive {
+	if recursive && key == "" {
```
AnomalRoil, line 46: This is unnecessary since it would have returned earlier otherwise, but I guess it won't hurt.

AnomalRoil (approved):

PR #1932, opened 2021-05-25
Add fish completion for `otp` subcommand

valer-cara:
Fill in entries on `gopass otp <TAB>` in fish shell, just like `gopass show <TAB>`.

Funny testing this on Arch. Fish checks and loads `/usr/share/fish/vendor_completions.d/gopass.fish` which cleared my local mods. Took a while to figure out why i had to double source this file. Alas, once this is merged it'll make it's way into the Arch vendor files.


dominikschulz:
Sorry, but this is a generated file.
We'll need to fix the generator.

valer-cara:
@dominikschulz - I've fixed the generator and also ran `make completion` which put my initial change in place

dominikschulz (approved):
LGTM. Thanks!

PR #1937, opened 2021-05-29
Remove GOPASS_NOCOLOR

### LINKED ISSUE #1936: Remove GOPASS_NOCOLOR
As suggested in https://github.com/gopasspw/gopass/pull/1931#pullrequestreview-669995780

dominikschulz:
Not sure if we need provide backwards compatiblity by mapping GOPASS_NOCOLOR to NO_COLOR. I'd argue not since we did support NO_COLOR for a long time already.

Also while working on this I noticed that we had both `Color` as well as `NoColor` keys in `ctxutil`. Well, that's why I like remove config options so much ...

Fixes #1936

RELEASE_NOTES=[CLEANUP] Remove GOPASS_NOCOLOR in favor of NO_COLOR


docs/config.md
```diff
@@ -33,6 +32,7 @@ Variables not exclusively used by gopass
 | `PAGER`                | `string` | the pager program used for `gopass list`. See [Features](features.md#auto-pager) for details           |
 | `GIT_AUTHOR_NAME`      | `string` | name of the author, used by the rcs backend to create a commit                                         |
 | `GIT_AUTHOR_EMAIL`     | `string` | email of the author, used by the rcs backend to create a commit                                        |
+| `NO_COLOR`             | `bool`   | disable color output
```
AnomalRoil, line 35: I would mention the https://no-color.org/ thing, because setting a global NO_COLOR env can have unexpected side effects on their system if people are not aware it's a "common" variable.

docs/features.md
```diff
@@ -444,7 +444,7 @@ Running `gopass [generate|insert] foo/bar` on an existing entry `foo/bar` will o
 
 ### Disabling Colors
 
-Disabling colors is as simple as setting `GOPASS_NOCOLOR` to `true`.
+Disabling colors is as simple as setting `NO_COLOR` to `true`.
```
AnomalRoil, line 447: Same, mentioning https://no-color.org/ seems like a good plan.

main_test.go
```diff
@@ -164,8 +164,4 @@ func TestInitContext(t *testing.T) {
 
 	assert.NoError(t, os.Setenv("GOPASS_DEBUG", "true"))
 	ctx = initContext(ctx, cfg)
```
AnomalRoil, line 166: You can also remove this line I think, since it was needed for the next ones.

AnomalRoil (changes requested):
LGTM after just a few extra changes.

It's nice to remove such things, indeed. I hadn't noticed we had both "HasColor" and "HasNoColor" stuff xD

AnomalRoil (approved):
LGTM, but this shouldn't be merged as part of 1.12.7 IMO since it's somewhat breaking for people relying on the GOPASS_NOCOLOR env var.

I propose to have this as part of a 1.13.0 release.

dominikschulz:
Sounds good.

PR #1952, opened 2021-07-01
Allow --strict to be chained with --symbols

### LINKED ISSUE #1941: Please allow `--strict` to be chained with `--symbols`
### Summary
Right now it seems that `--strict` and `--symbols`/`-s` are mutually exclusive, with strict taking priority. These should have the option to be chained together in order to form passwords that have strictly have one member from each char group, symbols included. 

The reason these should be chained rather than overriding the behavior of `--strict` is there are websites which require uppercase letters, lowercase letters, digits, **and special characters**, just as there are websites that require strict passwords **that do not allow special characters**(for some stupid reason).

the former currently requires continuously generating passwords with the `-s` flag until one meets the requirements. 

### Steps To Reproduce
`gopass generate --symbols --strict any.given.website/user@email`
 
### Expected behavior
a password with a member from each class `a-z`,`A-Z`,`0-9` `!@#$%^&*()-_+=[{]},<.>/?;:'"\`~\\|`

### Environment

- OS: fedora 34
- OS version: 5.12.8-300.fc34.x86_64 #1 SMP Fri May 28 15:20:54 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: gopass 1.12.5 
- Installation method: copr `daftaupe/gopass`

dominikschulz:
Fixes #1941

RELEASE_NOTES=[BUGFIX] Allow --strict to be chained with --symbols

### DIFF
--- a/internal/action/create.go
+++ b/internal/action/create.go
@@ -396,7 +396,7 @@ func (s *Action) createGeneratePassword(ctx context.Context, hostname string) (s
 		return "", err
 	}
 	if corp {
-		return pwgen.GeneratePasswordWithAllClasses(length)
+		return pwgen.GeneratePasswordWithAllClasses(length, symbols)
 	}
 
 	return pwgen.GeneratePassword(length, symbols), nil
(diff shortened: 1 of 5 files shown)


AnomalRoil (approved):
LGTM

PR #1953, opened 2021-07-01
Normalize recipient IDs before comparsion

### LINKED ISSUE #1900: fsck reports false-positives when comparing recipients
`gopass fsck` can report false positives if the `.gpg-id` files don't contain full size hashes but either short ones or recipient names.

The [comparison](https://github.com/gopasspw/gopass/blob/master/internal/store/leaf/fsck.go#L104) must first normalize to long hashes before comparing them.

dominikschulz:
Fixes #1900

RELEASE_NOTES=[BUGFIX] Normalize recipient IDs before comparison

### DIFF
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/gopasspw/gopass/internal/backend"
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/internal/store"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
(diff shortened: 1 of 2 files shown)


AnomalRoil (dismissed):
Nice, I hadn't even noticed we had already a `fingerprint` method.

AnomalRoil:
But this broke some tests it seems.

internal/store/leaf/fsck.go
```diff
@@ -127,6 +127,14 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) error {
 	return nil
 }
 
+func fingerprints(ctx context.Context, crypto backend.Crypto, in []string) []string {
+	out := make([]string, 0, len(in))
+	for _, r := range in {
+		out = append(out, crypto.Fingerprint(ctx, r))
+	}
+	return out
```
AnomalRoil, line 135: We might want to sort `out` before return, no? 
Are we guaranteed to receive the keys in the same order every time?
Also `itemRecps` might need to be sorted, no?

dominikschulz: Shouldn't be necessary. The comparison uses maps: https://github.com/dominikschulz/gopass/blob/57bb81f2af87f2b92b96f535461f3aba84eb5404/internal/store/leaf/fsck.go#L138

AnomalRoil: Nice, I should have checked that function too.

Issue #1819, opened 2021-02-28
RFC: Simplify config handling


dominikschulz:
The current config implementation is pretty strict so it can provide proper backwards compatibility.

However this makes it harder to edit the configuration manually since we don't allow extra fields and don't support comments, cf. #1654 

Should we change the config format to a simple key-value format (ini-style, no YAML) that supports comments?

It's not trivial since then programmatic changes, e.g. through `gopass config foo bar` become more difficult and we'd need to figure out backwards compatibility. But I wonder if there is any need for that at all.

AnomalRoil:
We might also want to discuss #1567 since we're discussing config handling :) 

In my opinion:
- We should support a config file per store, that's stored at the root of the said store and fallback to the "main" config file if the store doesn't have a specific config. (This can be annoying in teams, since the config file per store might be shared? Maybe have a setting in the main one to disable secondary config files?) Ideally the logic to handle it should be the same, it would just be its location that would influence which file is used.
- Partial config files should be interpreted and non-specified fields should default to Gopass' default values. I believe this should actually already work with the `mostRecent` config we've got, but maybe we're missing some default values? https://github.com/gopasspw/gopass/blob/e54b2a5cfd17cda7b353f51d0316b4c6ab5a7ae7/internal/config/io.go#L106-L109
Seems we are missing `autoimport`, `cliptimeout` and `notifications` default values.   
In any case we should make sure to honor the `$PASSWORD_STORE_DIR` value in case the path is the default one, no?
I guess the issue in #1654 is probably that we do not correctly handle an empty path in the config.

- We should support comments   
Actually I'm surprised that we do not already support comments since it seems we are using a YAML parser, so it should work with the comments as per YAML format, no?

morningspace:
Just a couple of comments re: the comments idea.

When we say support comments, it looks the implication is that we encourage user to modify the configuration somewhere in a file, so that they can make comments alongside a certain config item. It's intuitive to modify a YAML file and add comments to it. 

On the other side, the current `gopass config` is more like a way that encourage people to change the configuration from the command line which does not need to touch the YAML file directly. If we do not have too many configurable items and people have already been familiar the way as such, I don't see benefits to add the comment support at least for now.

P.S.

Maybe we can take `git config` as reference, which has both the command and the config file w/ ini-like format. YAML parser should have comment supported. So, if we did not expose the config file format to the end user, this seems to be a nice way that aligns w/ how git user uses `git config`. To me, it's just not a must have for now.

[...]

PR #1968, opened 2021-08-02
Using sane default for partial config

### LINKED ISSUE #1654: Setting `mime:false` in config blanks out `path`, ignores `$PASSWORD_STORE_DIR`, changes other defaults
### Summary

I don't want to define the path in my `config.yml`, so I only added `mime: false` there. (Need that because of #1600).

Afterwards, any `gopass show` instead asked me to run an 'onboarding wizard'. It seems this is because running `gopass config` now shows it has picked up `mime: false` but reset everything else to different defaults than it had chosen when I had no `config.yml`.

### Steps To Reproduce

1. Use pass, have `$PASSWORD_STORE_DIR` set, etc.
1. Install gopass
1. `test "$(gopass config path)" = "path: $PASSWORD_STORE_DIR"` (0)
1. echo 'mime: false' > "$XDG_CONFIG_HOME/gopass/config.yml"
1. `test "$(gopass config path)" = "path: $PASSWORD_STORE_DIR"` (0; it's `""`; other settings also changed)

### Expected behavior
The only config that changes is the `mime: false` we changed in creating `config.yml`.

### Environment

- OS: Arch Linux
- OS version: 5.9.8-arch1-1 
- gopass Version: [gopass version] 
```
It seems you are new to gopass. Do you want to run the onboarding wizard? [Y/n/q]: q
gopass 1.10.1 (2020-09-13 00:16:24) go1.15.2 linux amd64
<root>     -     <none> - git 2.29.2
```
- Installation method: Arch package `community/gopass`

### Additional context

`gopass config` without `mime: false` in `config.yml` (before creating it, or after deleting it):
```
autoclip: false
autoimport: true
cliptimeout: 45
exportkeys: true
mime: true
nocolor: false
nopager: false
notifications: true
path: [value of $PASSWORD_STORE_DIR]
safecontent: false
```

`gopass config` with `mime: false` in `config.yml`:
```
autoclip: false
autoimport: false
cliptimeout: 0
exportkeys: true
mime: false
nocolor: false
nopager: false
notifications: false
path:
safecontent: false
```

`autoimport`, `cliptimeout`, `notifications`, and `path` were all changed despite not being specified. Deleting `config.yml` restores them to 'previous defaults'.

Presumably a problem for / would be caught by #1622.

### LINKED ISSUE #1967: Default path config
### Summary

It would be nice to have a default for the `path` config parameter. I would suggest having it default to `~/.local/share/gopass/stores/root` if the config parameter is not set.

AnomalRoil:
RELEASE_NOTES=[BUGFIX] Use same default for partial config files

This is also upgrading the documentation regarding setup as we didn't reflect the move to the .local dir in there yet.

I've also removed most special cases for config in the tests, since now our defaults are as they should be.
I let the `exportkey` one set to false since I figured we didn't want to export keys during testing unless testing explicitly that.

This fixes #1654 and fixes #1967
Notice this doesn't remove the need for #1819.

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -61,7 +61,7 @@ gopass init
 ```
 
 This will prompt you for which GPG key you want to associate the store with.
-Then it will create a `.password-store` directory in your home directory.
+Then it will create a `.local/share/gopass/stores/root` directory in your home directory.
 
 If you don't want gopass to use this default directory, you can instead initialize a password store with:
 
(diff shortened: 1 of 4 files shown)


dominikschulz (approved):
Great work. Thanks!

PR #1983, opened 2021-08-28
Don't prompt to retype password unnecessarily with age backend

duxovni:
When gopass can't find a pinentry binary to use for prompting for a password to unlock an age keyring, it uses its own barebones fallback pinentry mechanism to prompt for a password in the terminal.  This fallback pinentry always asked the user to retype their password and required the passwords to match, even when decrypting a file using an already-existing password.  I've updated it to only prompt for a repeat when necessary, and also made the password prompt messages less awkward.

RELEASE_NOTES=[BUGFIX] Don't prompt to retype password unnecessarily with age backend when pinentry binary is unavailable.

Signed-off-by: Faye Duxovni <duxovni@duxovni.org>

### DIFF
--- a/pkg/pinentry/cli/fallback.go
+++ b/pkg/pinentry/cli/fallback.go
@@ -7,11 +7,13 @@ import (
 )
 
 // Client is pinentry CLI drop-in
-type Client struct{}
+type Client struct {
+	repeat bool
+}
 
 // New creates a new client
 func New() (*Client, error) {
-	return &Client{}, nil
+	return &Client{repeat: false}, nil
 }
 
 // Close is a no-op
@@ -22,8 +24,11 @@ func (c *Client) Confirm() bool {
 	return true
 }
 
-// Set is a no-op
-func (c *Client) Set(string, string) error {
+// Set is a no-op unless you're requesting a repeat
+func (c *Client) Set(key string, _ string) error {
+	if key == "REPEAT" {
+		c.repeat = true
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


pkg/pinentry/cli/fallback.go
```diff
@@ -7,11 +7,13 @@ import (
...
 // Client is pinentry CLI drop-in
-type Client struct{}
+type Client struct {
+	repeat bool
+}
 
 // New creates a new client
 func New() (*Client, error) {
-	return &Client{}, nil
+	return &Client{repeat: false}, nil
```
AnomalRoil, line 16: It's not clear to me why we would want to disable repeat by default for new clients.
What's the motivation there?

AnomalRoil (approved):

pkg/termio/ask_test.go
```diff
@@ -214,25 +214,30 @@ func TestAskForPasswordNonInteractive(t *testing.T) {
...
 	assert.NoError(t, err)
 	assert.Equal(t, "foo", sv)
 
-	sv, err = AskForPassword(ctx, "test")
+	sv, err = AskForPassword(ctx, "test", false)
+	assert.NoError(t, err)
+	assert.Equal(t, "foobar", sv)
+
+	sv, err = AskForPassword(ctx, "test", true)
 	assert.NoError(t, err)
```
AnomalRoil, line 241: Is this really the behaviour we want, btw, not erroring out but instead outputting the empty string is the two passwords aren't matching?

pkg/pinentry/cli/fallback.go
duxovni: This module is being used as a drop-in replacement for pinentry, so the motivation is to replicate pinentry's behavior.

pkg/termio/ask_test.go
duxovni: I don't know, but that's the original functionality and isn't changing here, so it seems outside the scope of this PR.

AnomalRoil: Agreed, but I was raising this for discussion with @dominikschulz to see if we needed to address this in another PR

dominikschulz (approved):

PR #1985, opened 2021-09-03
Add capitalized words to memorable passwords

### LINKED ISSUE #1984: generate memorable password with --strict
### Summary

When you `gopass generate -g memorable` , it produces only lowercase passwords. When combined with `--strict` , uppercase some letters please?

### Steps To Reproduce

`gopass generate -g memorable -s --strict test 15`

outputs like

`alphabet5]soup4,`

### Expected behavior

a password with a member from each class, including uppercase `A-Z`, like `Alphabet5]soup4,`

### Environment

- OS: MacOS
- OS version: Big Sur 11.5
- gopass Version: 1.12.8
- Installation method: brew

elijahjpassmore:
Memorable passwords generated with --strict will now have at least one
capitalized word.

Closes #1984

RELEASE_NOTES=[FEATURE] Add capitalized words to memorable passwords

Signed-off-by: Elijah J. Passmore <elijah@elijahjpassmore.com>

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -219,7 +219,10 @@ func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, n
 	case "xkcd":
 		return s.generatePasswordXKCD(ctx, c, length)
 	case "memorable":
-		return pwgen.GenerateMemorablePassword(pwlen, symbols), nil
+		if c.Bool("strict") {
+			return pwgen.GenerateMemorablePassword(pwlen, symbols, true), nil
+		}
+		return pwgen.GenerateMemorablePassword(pwlen, symbols, false), nil
 	case "external":
 		return pwgen.GenerateExternal(pwlen)
 	default:
--- a/pkg/pwgen/memorable.go
+++ b/pkg/pwgen/memorable.go
@@ -4,15 +4,33 @@ import "strings"
 
 // GenerateMemorablePassword will generate a memorable password
 // with a minimum length
-func GenerateMemorablePassword(minLength int, symbols bool) string {
+func GenerateMemorablePassword(minLength int, symbols bool, capitals bool) string {
 	var sb strings.Builder
+	var upper = false
 	for sb.Len() < minLength {
-		sb.WriteString(randomWord())
+		if capitals {
+			if randomInteger(2) == 0 {
+				sb.WriteString(strings.Title(randomWord()))
+				upper = true
+			} else {
+				sb.WriteString(randomWord())
+			}
+		} else {
+			sb.WriteString(randomWord())
+		}
 		sb.WriteByte(Digits[randomInteger(len(Digits))])
 		if !symbols {
 			continue
 		}
 		sb.WriteByte(Syms[randomInteger(len(Syms))])
+	}
+	// If there isn't already a capitalized word, capitalize the first letter
+	if capitals {
+		if upper {
+			return sb.String()
+		}
+		var str = sb.String()
+		return strings.Title(string(str[0])) + str[1:]
 	}
 	return sb.String()
 }
--- a/pkg/pwgen/memorable.go
+++ b/pkg/pwgen/memorable.go
@@ -22,10 +22,7 @@ func GenerateMemorablePassword(minLength int, symbols bool, capitals bool) strin
 		sb.WriteByte(Syms[randomInteger(len(Syms))])
 	}
 	// If there isn't already a capitalized word, capitalize the first letter
-	if capitals {
-		if upper {
-			return sb.String()
-		}
+	if capitals && !upper {
 		var str = sb.String()
 		return strings.Title(string(str[0])) + str[1:]
 	}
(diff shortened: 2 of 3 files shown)


pkg/pwgen/memorable.go
```diff
@@ -4,16 +4,34 @@ import "strings"
 
 // GenerateMemorablePassword will generate a memorable password
 // with a minimum length
-func GenerateMemorablePassword(minLength int, symbols bool) string {
+func GenerateMemorablePassword(minLength int, symbols bool, capitals bool) string {
```
AnomalRoil, line 7: That's a public API change, but it seems only Gopass and Gopass JSON-API (which don't use that function) is using this package currently, so that should be fine.
We might want to increment a minor version instead of a patch version because of that tho.

pkg/pwgen/memorable.go
```diff
@@ -4,16 +4,34 @@ import "strings"
...
+		if capitals {
+			if randomInteger(2) == 0 {
+				sb.WriteString(strings.Title(randomWord()))
+				upper = true
+			} else {
+				sb.WriteString(randomWord())
+			}
+		} else {
+			sb.WriteString(randomWord())
+		}
```
AnomalRoil, line 20: the double if with both having the same else clause could be rewritten:
```suggestion
		// when requesting uppercase, we randomly uppercase words
		if capitals && randomInteger(2) == 0 {
			sb.WriteString(strings.Title(randomWord()))
			upper = true
		} else {
			sb.WriteString(randomWord())
		}
```

pkg/pwgen/memorable.go
```diff
@@ -4,16 +4,34 @@ import "strings"
...
 		if !symbols {
 			continue
 		}
 		sb.WriteByte(Syms[randomInteger(len(Syms))])
 	}
+	// If there isn't already a capitalized word, capitalize the first letter
+	if capitals {
+		if upper {
+			return sb.String()
+		}
```
AnomalRoil, line 31: Could be instead:
```suggestion
	if capitals && !upper {
```

pkg/pwgen/pwgen_test.go
```diff
@@ -93,7 +93,12 @@ func TestGeneratePasswordWithAllClasses(t *testing.T) {
...
 
 func TestGenerateMemorablePassword(t *testing.T) {
-	pw := GenerateMemorablePassword(20, false)
+	pw := GenerateMemorablePassword(20, false, false)
+	assert.GreaterOrEqual(t, len(pw), 20)
+}
+
+func TestGenerateMemorablePasswordCapital(t *testing.T) {
+	pw := GenerateMemorablePassword(20, false, true)
 	assert.GreaterOrEqual(t, len(pw), 20)
```
AnomalRoil, line 102: It would be nice to assert it contains a uppercase. You can compare the string with the lowercase version `strings.ToLower(pw)` as it shouldn't be the same in that case.

Also maybe add in the previous test that it is only lowercase if set to `false`?

internal/action/generate.go
```diff
@@ -219,7 +219,10 @@ func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, n
 	case "xkcd":
 		return s.generatePasswordXKCD(ctx, c, length)
 	case "memorable":
-		return pwgen.GenerateMemorablePassword(pwlen, symbols), nil
+		if c.Bool("strict") {
+			return pwgen.GenerateMemorablePassword(pwlen, symbols, true), nil
+		}
+		return pwgen.GenerateMemorablePassword(pwlen, symbols, false), nil
```
AnomalRoil, line 225: @dominikschulz actually we're using `WriteString` in there which can return an error... Why aren't we returning errors in the Generate*Password functions and instead always return `nil` here??

AnomalRoil (changes requested):
@dominikschulz I'd love to have your input regarding the possibility of `WriteString` to error out and us not forwarding it currently. Should we also refactor the Generate*Password functions to return `(string, error)` and get rid of the `nil ` in the `return ..., nil` instead?

AnomalRoil:
Thank you for the contribution @elijahjpassmore, that's really appreciated. I've just got a few comments, if you could review and change things accordingly, that'd be great.

elijahjpassmore:
Thanks @AnomalRoil, I've made the changes you suggested. If you have any other issues or questions, feel free to let me know.

AnomalRoil (approved):

internal/action/generate.go
dominikschulz: It's signature has an error argument, yes.
But at least with the current implementation it will never return an error: https://cs.opensource.google/go/go/+/refs/tags/go1.17.1:src/strings/builder.go;l=125;drc=43652dc46f770253b3603f47165b1568b439b0b5

pkg/pwgen/memorable.go
dominikschulz: We should probably make it a minor bump, yes. But otherwise I wouldn't be concerned about this change. This package is only used by gopass itself and gopass-jsonapi: https://pkg.go.dev/github.com/gopasspw/gopass@v1.12.8/pkg/pwgen?tab=importedby

That being said I also wouldn't set the bar too high here. This package doesn't have a very good interface anyway and is likely due for a major redesign at some point. The interfaces aren't very consistent. Cryptic uses an object while the others can only be called directly. Either all methods should be static or all object methods where the configuration goes into the object. Maybe with static helpers that use the most common settings, like the http package does.

dominikschulz (approved):

PR #1996, opened 2021-09-28
Update setup.md

astark13:
Added the "apt-get update" command in the "Pre-Installation Steps / Download and Install Dependencies / Ubuntu & Debian" in order to avoid "Package ... is not available" error when running "apt-get install gnupg2 git rng-tools"


AnomalRoil (approved):

AnomalRoil:
Thanks

PR #2004, opened 2021-10-04
Move prompts and progress bars to stderr rather than stdout

duxovni:
Having password prompts/etc on stdout is inconvenient in any context where the output is being piped to something other than a terminal.  This change makes gopass more scripting-friendly, especially when using the age backend.  It also removes a redundant password prompt message when the age backend asks for a password, since pinentry will display its own password prompt immediately afterward.

RELEASE_NOTES=[ENHANCEMENT] Move password prompts to stderr

Signed-off-by: Faye Duxovni <duxovni@duxovni.org>

### DIFF
--- a/internal/action/create_test.go
+++ b/internal/action/create_test.go
@@ -76,11 +76,11 @@ func TestCreateWebsite(t *testing.T) {
 	act.cfg.ClipTimeout = 1
 
 	buf := &bytes.Buffer{}
-	out.Stdout = buf
-	termio.Stdout = buf
+	out.Stderr = buf
+	termio.Stderr = buf
 	defer func() {
-		out.Stdout = os.Stdout
-		termio.Stdout = os.Stdout
+		out.Stderr = os.Stderr
+		termio.Stderr = os.Stderr
 	}()
 
 	// provide values on redirected stdin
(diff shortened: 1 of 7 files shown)


dominikschulz (approved):
LGTM, but I'd like to have second pair of eyes on this.

duxovni:
I can believe that everything works fine for you with the GPG backend; the age backend is way worse in this regard. Even just `gopass show` gives me password prompts in the output. There's the redundant prompt I removed, and then all the prompts generated by the CLI pinentry fallback; in order to fix those, I also needed to fix all of the other termio prompts, which then required fixing the progress bars too. There may well be a lot of other prompts that are still on stdout, but this covers all the prompts that might come up during `gopass show` with the age backend, and then all the prompts that use the same codepaths as those prompts.

dominikschulz:
Sorry for the sub-par experience with the age backend. It was really only a proof of concept for me (mostly to verify that Filos initial API was suitable for us). Thanks for your help in improving that!

dominikschulz:
Oh, no worries. That didn't come across too critical or anything. I just would love to spend more time on the age backend but couldn't get myself to do that so far.

I gave the changes a short try and I think we can merge this as is and then fix any other occurences as we go.

PR #2024, opened 2021-11-08
Fix AutoClip handling on generate

### LINKED ISSUE #2023: Autoclip on `generate` broken
### Summary
Currently, autoclip on `generate` does not work as I beieve was intended. 

Clip-on-generate currently works by checking whether one of two conditions are met:
1. Is the `-c` flag `true`
2. Is `autoclip` true and is output _being redirected_  

I belive (2.) to be incorrect autoclipping behavior, and discordant with the comment preceding it:

```
// copy to clipboard if:
// - explicitly requested with -c
// - autoclip=true, but only if output is not being redirected
if IsClip(ctx) || (s.cfg.AutoClip && !ctxutil.IsTerminal(ctx)) 
```

As can be seen, the code actually checks that output _is_ being redirected, i.e. the output file descriptor is _not_ a TTY. 

### Steps To Reproduce

1. `gopass config autoclip true` 
2. `gopass generate -f foobar 24`
3. Prints "Not printing secrets by default. Use 'gopass show foobar' to display the password.". Contents of `foobar` will not be copied to clipboard 

*Unexpected behavior*
1. `gopass config autoclip true`
2. `gopass generate -f foobar 24 | cat`
3. Prints "Copied to clipboard". Contents of `foobar` _will_  be copied to clipboard 

### Expected behavior
In the above scenarios, I exect the contents of `foobar` to be copied to the clipboard in the first scenario, but not the second. In other words, the inverse of the actual behavior. 

### Environment
- OS: Ubuntu 20.04
- OS version: 5.13.0-7614-generic #14~1631647151~20.04~930e87c~dev-Ubuntu SMP Thu Sep 16 19:26:07  x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: gopass 1.12.8-git+HEAD go1.17.1 linux amd64
- Installation method: `go install` of latest 

### Additional context
I'm happy to submit a PR for the fix if this is confirmed to be unexpected behavior.

dominikschulz:
Fixes #2023

RELEASE_NOTES=[BUGFIX] Fix AutoClip handling on generate

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -120,7 +120,7 @@ func (s *Action) generateCopyOrPrint(ctx context.Context, c *cli.Context, name, 
 	// copy to clipboard if:
 	// - explicitly requested with -c
 	// - autoclip=true, but only if output is not being redirected
-	if IsClip(ctx) || (s.cfg.AutoClip && !ctxutil.IsTerminal(ctx)) {
+	if IsClip(ctx) || (s.cfg.AutoClip && ctxutil.IsTerminal(ctx)) {
 		if err := clipboard.CopyTo(ctx, name, []byte(password), s.cfg.ClipTimeout); err != nil {
 			return ExitError(ExitIO, err, "failed to copy to clipboard: %s", err)
 		}
--- a/internal/action/generate_test.go
+++ b/internal/action/generate_test.go
@@ -150,6 +150,32 @@ func TestGenerate(t *testing.T) {
 		assert.NoError(t, act.Generate(gptest.CliCtxWithFlags(ctx, t, map[string]string{"force": "true", "xkcd": "true", "print": "true", "lang": "en"}, "foobar", "baz")))
 		buf.Reset()
 	})
+
+	// generate --force foobar 24 w/ autoclip and output redirection
+	t.Run("generate --force foobar 24", func(t *testing.T) {
+		ov := act.cfg.AutoClip
+		defer func() {
+			act.cfg.AutoClip = ov
+		}()
+		act.cfg.AutoClip = true
+		ctx := ctxutil.WithTerminal(ctx, false)
+		assert.NoError(t, act.Generate(gptest.CliCtxWithFlags(ctx, t, map[string]string{"force": "true"}, "foobar", "24")))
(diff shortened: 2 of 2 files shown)


internal/action/generate.go
```diff
@@ -120,7 +120,7 @@ func (s *Action) generateCopyOrPrint(ctx context.Context, c *cli.Context, name,
 	// copy to clipboard if:
 	// - explicitly requested with -c
 	// - autoclip=true, but only if output is not being redirected
-	if IsClip(ctx) || (s.cfg.AutoClip && !ctxutil.IsTerminal(ctx)) {
+	if IsClip(ctx) || (s.cfg.AutoClip && ctxutil.IsTerminal(ctx)) {
```
AnomalRoil, line 123: Maybe I should be replacing the content of my new function `out.IsOutputRedirected`  by this then.
I'm not sure how `ctxutil.IsTerminal(ctx)` behaves when we are piping stuff but in a terminal.

internal/action/generate_test.go
```diff
@@ -150,6 +150,32 @@ func TestGenerate(t *testing.T) {
...
+	// generate --force foobar 24 w/ autoclip and output redirection
+	t.Run("generate --force foobar 24", func(t *testing.T) {
+		ov := act.cfg.AutoClip
+		defer func() {
+			act.cfg.AutoClip = ov
+		}()
+		act.cfg.AutoClip = true
+		ctx := ctxutil.WithTerminal(ctx, false)
+		assert.NoError(t, act.Generate(gptest.CliCtxWithFlags(ctx, t, map[string]string{"force": "true"}, "foobar", "24")))
+		assert.Contains(t, buf.String(), "Not printing secrets by default")
```
AnomalRoil, line 163: Are we printing this on `StdOut`? If so we might want to consider doing that on `StdErr` instead in a future PR.

AnomalRoil (approved):
LGTM

internal/action/generate.go
AnomalRoil: Now in #2025

internal/action/generate_test.go
dominikschulz: Yes, I think this goes to stdout right now.

PR #2034, opened 2021-11-12
[Go 1.18] Introduce type parameters

### LINKED ISSUE #2030: [Go 1.18] Consider type parameters
Go 1.18 brings type parameters (aka generics). I'm not overly enthusiastic about that change but I'm still curious to try them and if we have any real opportunities to simplify the code by using them I'd be happy to.

dominikschulz:
This commit replaces some overly complex constructs with type
parameters.

Fixes #2030

RELEASE_NOTES=[ENHANCEMENT] Generics


internal/backend/registry.go
```diff
@@ -3,30 +3,103 @@ package backend
...
+	}
+	sort.Strings(names)
+	return names
+}
+
+func (r *Registry[K, V]) Prioritized() []V {
+	bes := make([]V, 0, len(r.backends))
+	for _, be := range r.backends {
+		bes = append(bes, be)
+	}
```
AnomalRoil, line 77: Looking forward the new `maps` package in 1.19 for that :)

internal/backend/registry.go
```diff
@@ -3,30 +3,103 @@ package backend
...
+		return be, nil
+	}
+	var zero V
+	return zero, ErrNotFound
+}
+
+func (r *Registry[K, V]) Backend(name string) (K, error) {
+	if name == "gpg" {
+		name = "gpgcli"
+	}
```
AnomalRoil, line 95: Why do we have a hardcoded "exception" like that here? Wouldn't it be best to change the name on the backend side?

internal/backend/registry.go
```diff
@@ -3,30 +3,108 @@ package backend
...
+		return bes[i].Priority() < bes[j].Priority()
+	})
+	return bes
+}
+
+func (r *Registry[K, V]) Get(key K) (V, error) {
+	if be, found := r.backends[key]; found {
+		return be, nil
+	}
+	var zero V
```
AnomalRoil, line 88: Nice one, I'll remember it when I'll play with Generics too.

AnomalRoil (approved):
Nothing special to say, I'm yet to get used to Generics, but LGTM

internal/backend/registry.go
dominikschulz: yeah, probably. need to clean this up but it's still hardcoded in a bunch of places.

PR #2040, opened 2021-11-14
Fix autoclip-on-generate behavior.

### LINKED ISSUE #2023: Autoclip on `generate` broken

acaloiaro:
Fix #2023 

Previously, secrets would autoclip only when output was being
redirected, which is incorrect. Secrets should autoclip when output is
not being redirected (i.e. we're on a TTY).

RELEASE_NOTES=[BUGFIX] Fixed autoclip when generating secrets.

Signed-off-by: Adriano Caloiaro <github@adriano.fyi>

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -120,7 +120,7 @@ func (s *Action) generateCopyOrPrint(ctx context.Context, c *cli.Context, name, 
 	// copy to clipboard if:
 	// - explicitly requested with -c
 	// - autoclip=true, but only if output is not being redirected
-	if IsClip(ctx) || (s.cfg.AutoClip && !ctxutil.IsTerminal(ctx)) {
+	if IsClip(ctx) || (s.cfg.AutoClip && ctxutil.IsTerminal(ctx)) {
 		if err := clipboard.CopyTo(ctx, name, []byte(password), s.cfg.ClipTimeout); err != nil {
 			return ExitError(ExitIO, err, "failed to copy to clipboard: %s", err)
 		}
--- a/internal/action/generate_test.go
+++ b/internal/action/generate_test.go
@@ -150,6 +150,38 @@ func TestGenerate(t *testing.T) {
 		assert.NoError(t, act.Generate(gptest.CliCtxWithFlags(ctx, t, map[string]string{"force": "true", "xkcd": "true", "print": "true", "lang": "en"}, "foobar", "baz")))
 		buf.Reset()
 	})
+
+	// regression test for #2023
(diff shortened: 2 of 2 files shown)


AnomalRoil:
I think that was actually already fixed in #2024, no?

But you have more tests, so I'd love to still get the extra tests if you could rebase on master and fix the conflict?

acaloiaro:
I see. I'll go ahead and shut this one down. If anyone wants to grab the extra test and apply it, please feel free.

acaloiaro:
@AnomalRoil The existing test appears to be nearly identical in coverage to mine: https://github.com/dominikschulz/gopass/commit/c53a985819831ffac10986aac886b63cba1adc85#diff-0ed14f4fcc141393a29d562b970e0f3efa80ae0608ed2a3d00ceb37bba8918fd

PR #2019, opened 2021-10-27
Better progress bar handling for OTP

AnomalRoil:
RELEASE_NOTES=[BUGFIX] Do not print OTP progress bar if not in terminal
RELEASE_NOTES=[UX] Use new progress bar for OTP expiry time

This is making the expiry time for OTP visually go from 0 to expiry time in terminal, but prevents printing the progress bar when not in a terminal.

image

I also replaced the special handling of the remaining time until expiry to use `Truncate` instead.

Sadly, I'm yet again adding some "special handling" to detect whether we are in a terminal or not. We might want to heavily refactor all such things.

This fixes #2017.

PR #2041, opened 2021-11-17
Adding option to exit OTP progress bar

AnomalRoil:
As discussed in #2019


internal/action/otp.go
```diff
@@ -65,25 +66,52 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
 	} else { // if not then we want to print a progress bar with the expiry time
 		out.Printf(ctx, "%s", token)
-		out.Warningf(ctx, "This OTP password still lasts for:", nil)
+		out.Warningf(ctx, "([q] to stop. -o flag to avoid.) This OTP password still lasts for:", nil)
 		bar := termio.NewProgressBar(int64(secondsLeft))
 		bar.Hidden = ctxutil.IsHidden(ctx)
 		if bar.Hidden {
 			skip = true
 		} else {
+			cancel := make(chan bool)
```
dominikschulz, line 75: I wonder if I'm missing something, but can't you just use `ctx, cancel := context.WithCancel(ctx)` and then use the standard [cancel](https://github.com/gopasspw/gopass/blob/master/pkg/hibp/dump/scanner.go#L175) idiom there?

internal/action/otp.go
```diff
@@ -65,25 +66,52 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
 				for tt := range ticker.C {
-					if tt.After(expiresAt) {
+					select {
+					case <-cancel:
+						cancelled = true
+					default:
+						// we don't want to block if not cancelled
+					}
+					if tt.After(expiresAt) || cancelled {
 						bar.Done()
```
dominikschulz, line 89: Can we just move that to the end of the goroutine or in a defer? I think we always want to call this?

internal/action/otp.go
```diff
@@ -65,25 +66,52 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
 						done <- true
 						return
 					}
 					bar.Inc()
 				}
 			}()
+			go func() {
+				tty, err := tty.Open()
```
dominikschulz, line 97: Could you remind me why we need the tty lib here?
Can we not just do a regular [read](https://github.com/gopasspw/gopass/blob/master/pkg/termio/ask.go#L43)?

dominikschulz (changes requested):
I wonder why we need the go-tty lib. Also I think gocycle rightfully complains. If you need to keep the nesting at least move the implementation of these goroutines to explicit functions.

AnomalRoil:
Yeah, this is still a work in progress since the Linux CI isn't passing, but I was trying to code on Windows for once, which made my life a little bit more complicated.

Excellent idea @dominikschulz about putting the Done in a defer.

For the go-tty, we need it in order to read a keystroke on all platforms, without requiring to press Enter after entering our input...
If we're fine with requiring a Enter after the keystroke I could get rid of it and use the Context Cancel syntax, I guess. Wdyt?

dominikschulz:
@AnomalRoil Hmm, I thought I had replied to this one. But pressing enter after q sounds very un-intuitive to me. So let's go with go-tty instead.

AnomalRoil:
At long last! I shouldn't try and dev when I cannot run the full test suite and codequality checks locally xD

internal/action/otp.go
```diff
@@ -36,6 +37,48 @@ func (s *Action) OTP(c *cli.Context) error {
...
+		cancel()
+	}
+	defer tty.Close()
+	for {
+		select {
+		case <-ctx.Done():
+			return // returning not to leak the goroutine
+		default:
+		}
+		r, err := tty.ReadRune()
```
AnomalRoil, line 71: Actually I wonder if this is not waiting until the next keystroke, and thus this goroutine isn't returning when it should... (basically it's always in iteration 1 of the for loop...)

AnomalRoil: Yeah, the goroutine is leaking when the OTP Bar finishes on line 52, I think. 
Related: https://benjamincongdon.me/blog/2020/04/23/Cancelable-Reads-in-Go/

Seems like there isn't an easy way out...

AnomalRoil: I think I'll have to refactor to use a cancelable reader, using a `io.Reader` on `os.Stdin` and test it on Windows hoping it will work fine [without all that boilerplate](https://github.com/mattn/go-tty/blob/72ed86c4554d4bf5afd278468f2eb74df8de8f97/tty_windows.go#L200) since we are just caring about key presses...

dominikschulz (approved):
The change looks fine and seems to work well.
But I was immedeately thinking that we should maybe keep generating new OTP codes, just like the Autenticator apps do.

dominikschulz:
Don't feel obliged to add that. I'm not a huge fan of the OTP feature anyway, but since we have it we could try to make it as nice and useful as it can be. Up to you.

AnomalRoil:
It was actually easier to "fix" the Go routine leak by not fixing it and having a loop so that it's mandatory to cancel the loop and there's no risk of having a character being eaten by the leaked go routine when we're back to gopass prompt view.
So I've implemented your suggestion, let me know what you think.

dominikschulz:
Seems to work fine. Thank you.

dominikschulz (approved):

colemickens:
Looks like another major behavior change, change of default, and no global config for this? Or maybe I skimmed the diff too quickly?

AnomalRoil:
@colemickens I'm not sure how that's a change of behaviour?
We used to have a "ugly progress bar" for OTP, and we now have a nicer looking one, which has a prompt to quit it.

colemickens:
It's a big change for anyone that had scripts that relied on "show --clip" and "totp --clip" being one-shot behavior. AFAICT there's no way for me to have the same keyboard shortcut working before without doing something like chaining `until` with gopass so that it gets killed after a few seconds. 

Not a huge deal but it is a change in UX and sadly I just hit another (unrelated) change in `gopass` yesterday and I'm just tired. No other mission critical tool "moves" under me as much as gopass has over the past couple of years.

dominikschulz:
@colemickens I still don't get it.

Before this PR we had a progress bar that you couldn't cancel.
After this PR we still have a progress bar. That you can quit by pressing `q`.

How would that break `show --clip`?

We can discuss if we should print the progress bar when `--clip` is given or not - but that didn't change this PR.

colemickens:
Unless I'm bad at reading Go lately. the PR seems to add a loop that wasn't there before that causes the OTP generation to loop. I have a keybinding that pops up a terminal, runs `gopass [show|totp] --clip`  and then expects it to exit. Maybe the entirety of this behavior isn't just encapsulated in this PR, but it definitely exacerbated it.

It does seem like there's a short-circuit for when it's being piped, so I can probably workaround this.

dominikschulz:
The actual loop was already present before this PR. Try it yourself: `git checkout adf6254add0cfd5dd56141bcb14247e43089898b^` (this will get you the state before this was merged) and then `go build`.

I'm not sure if we should be printing the progress bar at all when `--clip` is specified, but I don't think this PR should be able to break any workflow.

AnomalRoil:
Yeah, the fact the progress bar now prints when using `--clip` is totally a bug.

AnomalRoil:
BTW, any unexpected behaviour change warrants an issue, don't hesitate to open issue whenever you run into something breaking. Don't hesitate to open issues if you run into anything that breaks your workflow after any update.

PR #2047, opened 2021-11-25
Fix: 1917 key caps

TM2500:
Adds GPG-Key Capabilities Parsing & and Key-IsUsable return False when no sufficient Key-Caps.

Sorry, that there are no test-cases included in this PR.
I could not figure out a nice and easy way to add invalid-keys, without breaking everything else.

We ran this Feature-branch in prod for quite a while and thought it might be time to merge into upstream.

### DIFF
--- a/internal/backend/crypto/gpg/cli/encrypt.go
+++ b/internal/backend/crypto/gpg/cli/encrypt.go
@@ -26,7 +26,7 @@ func (g *GPG) Encrypt(ctx context.Context, plaintext []byte, recipients []string
 		if err != nil {
 			debug.Log("Failed to check key %s. Adding anyway. %s", err)
 		} else if len(kl.UseableKeys(gpg.IsAlwaysTrust(ctx))) < 1 {
-			out.Printf(ctx, "Not using expired key %s for encryption", r)
+			out.Printf(ctx, "Not using invalid key %s for encryption. (Check its expiration date or its encryption capabilities.)", r)
 			continue
 		}
 		args = append(args, "--recipient", r)
--- a/internal/backend/crypto/gpg/colons/parse_colons.go
+++ b/internal/backend/crypto/gpg/colons/parse_colons.go
@@ -114,21 +114,20 @@ func Parse(reader io.Reader) gpg.KeyList {
 
 func parseKeyCaps(field string) gpg.Capabilities {
 	keycaps := gpg.Capabilities{}
-	caps := strings.ToLower(field)
 
-	if strings.Contains(caps, "S") {
+	if strings.Contains(field, "S") {
 		keycaps.Sign = true
 	}
-	if strings.Contains(caps, "E") {
+	if strings.Contains(field, "E") {
 		keycaps.Encrypt = true
 	}
-	if strings.Contains(caps, "C") {
+	if strings.Contains(field, "C") {
 		keycaps.Certify = true
 	}
-	if strings.Contains(caps, "A") {
+	if strings.Contains(field, "A") {
 		keycaps.Authentication = true
 	}
-	if strings.Contains(caps, "D") {
+	if strings.Contains(field, "D") {
 		keycaps.Deactivated = true
 	}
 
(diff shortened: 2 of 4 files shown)


dominikschulz:
Thanks, that PR looks good. I think in this case it's acceptable to not have tests.

Once the checks are happy we can merge this.

AnomalRoil:
If you could maybe also pull master and rebase on it to just include your 4 new commits, that'd be great and should allow you to easily solve the merge conflict.

TM2500:
Thank you for the quick reply.
I don't know why, but one dependabot commit got into this PR.
I hope this doesn't mess things up.

Have a pleasant evening.

AnomalRoil:
You'll need to fix the test by adding the "Encryption" capability to the test keys here:
https://github.com/gopasspw/gopass/blob/1ccdf941d11771c447e0257819773a4dd65b227e/internal/backend/crypto/gpg/key_test.go#L43-L60

and also here:
https://github.com/gopasspw/gopass/blob/1ccdf941d11771c447e0257819773a4dd65b227e/internal/backend/crypto/gpg/key_test.go#L107-L118

TM2500:
Thanks for the hints.
Extended also the failed cases, so it fails at points where it's supposed to fail.

TM2500:
At the point of the integration-test, I don't have a clue where to begin and how to setup the gpg-env for these.

That was what I ment with the missing test-cases.

AnomalRoil:
This is really strange: our test key `0x82EBD945BE73F104` should be working:
```
sec   dsa2048 2017-01-25 [SC] [expires: 2027-01-23]
      AB919DBF9BF0DE74896397F282EBD945BE73F104
      Keygrip = C26D35176FC642692D621EA3D61481B425649C5C
uid           [ultimate] John Doe <john.doe@gopass.pw>
ssb   elg2048 2017-01-25 [E] [expires: 2027-01-23]
      Keygrip = 09B835E6A569C9F01DC27352147E5D332EC0A6A7
```

It's not yet expired... And yet the tests are failing saying it expired.

TM2500:
> This is really strange: our test key `0x82EBD945BE73F104` should be working:
> 
> ```
> sec   dsa2048 2017-01-25 [SC] [expires: 2027-01-23]
>       AB919DBF9BF0DE74896397F282EBD945BE73F104
>       Keygrip = C26D35176FC642692D621EA3D61481B425649C5C
> uid           [ultimate] John Doe <john.doe@gopass.pw>
> ssb   elg2048 2017-01-25 [E] [expires: 2027-01-23]
>       Keygrip = 09B835E6A569C9F01DC27352147E5D332EC0A6A7
> ```
> 
> It's not yet expired... And yet the tests are failing saying it expired.

probably because only the subkey has the Encrypt-Cap and it's only set in the subkey, but not in the parent-key.

AnomalRoil:
Yeah, I guess the problem is now how to handle subkeys graciously from a master key fingerprint? 

As a workaround for the test, we could try and replace:
https://github.com/gopasspw/gopass/blob/a4e2a777b0d45da0d269769ac1ce2422cdb55c97/tests/tester.go#L24
with the keygrip of the encryption subkey. Not sure if that works then?

But that would be SUPER breaking for a lot of our users who put their "parent key fingerprint" in their recipient list... So we cannot ship it as is, we need to graciously handle subkeys, IMO.

Also, please change the text here:
https://github.com/gopasspw/gopass/blob/1ccdf941d11771c447e0257819773a4dd65b227e/internal/backend/crypto/gpg/cli/encrypt.go#L29
to read instead "Not using invalid key %s for encryption. (Check its expiration date or its encryption capabilities.)"

TM2500:
Yeah, I totally understand.
I'm looking into that.
Strange that we had only keys without subkeys, but nevermind.

AnomalRoil:
Ah, I did miss the ToLower issue too!

A proof that tests are a really good thing.

AnomalRoil (approved):

Issue #2046, opened 2021-11-25
gopass otp progressbar formatting bug when invoked with 0 seconds left


defhacks:
### Summary

gopass otp prints mangled progress bar when token has 0 seconds left at time of invocation.
e.g.

```
⚠ This OTP password still lasts for:
] %!(NOVERB)%!(EXTRA int64=0, int64=0)[Goooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooopass] 100.00%
```

### Steps To Reproduce

run gopass otp <secret> when time left for token is 0. I just looped using Ctrl+C and re-running manually.

### Expected behavior

non-corrupt output in the progress bar

### Environment

- OS:  Ubuntu 20.04
- gopass Version: gopass 1.13.0+c574e920b532606894b3e64db059f3e3a838b0c9
- Installation method: gopass update, iirc ?

### Additional context

I tracked it down to the nested `Sprintf()` calls here: https://github.com/gopasspw/gopass/blob/master/pkg/termio/progress.go#L133

I was curious why this was happening and ended up being able to figure out what's going by building an isolated demo of the problematic code path, but I'm not clear what the best fix is, likely some special casing of `max = 0` ?

Should be `go run`-able, just add imports
```
func main() {
        max := 1
        digits := int(math.Log10(float64(max))) + 1
        fmt.Println(digits)
        fmt_string := fmt.Sprintf(" %%%dd / %%%dd ", digits, digits)
        fmt.Println(fmt_string)

        max = 0
        digits = int(math.Log10(float64(max))) + 1
        fmt.Println(digits)
        fmt_string = fmt.Sprintf(" %%%dd / %%%dd ", digits, digits)
        fmt.Println(fmt_string)
}
```
output:
```
1
 %1d / %1d
-9223372036854775807
 %-9223372036854775807d / %-9223372036854775807d
```

AnomalRoil:
Shoot, I guess we probably have a edge-case there. 

Do you want to add the special handling for the case where it's 0 yourself in a PR or shall I?

AnomalRoil:
Fixed in #2048

PR #2048, opened 2021-11-26
handle progressbar having 0 seconds left at startup

defhacks:
Attempt to address https://github.com/gopasspw/gopass/issues/2046

I'm not sure if the release notes commit message instructions are for my commit, or for the merge commit, but if I need change thing, I'm than happy to.

### DIFF
--- a/pkg/termio/progress.go
+++ b/pkg/termio/progress.go
@@ -130,6 +130,10 @@ func (p *ProgressBar) doPrint() {
 
 	barWidth := uint(termWidth)
 	digits := int(math.Log10(float64(max))) + 1
+	// Log10(0) is undefined
+	if max < 1 {
+		digits = 1
+	}
 	text := fmt.Sprintf(fmt.Sprintf(" %%%dd / %%%dd ", digits, digits), cur, max)
 	if p.Bytes {
 		curStr := humanize.Bytes(uint64(cur))


pkg/termio/progress.go
```diff
@@ -129,7 +129,14 @@ func (p *ProgressBar) doPrint() {
 	}
 
 	barWidth := uint(termWidth)
-	digits := int(math.Log10(float64(max))) + 1
+	var digits int
+	// Log10(0) is undefined
+	if max > 0 {
```
dominikschulz, line 134: I think we can make this a little bit shorter.

Could you leave the initial assignment as is and only add three lines to check max < 1 and then set the value to 1.

AnomalRoil:
For your commit to pass CI, you need to sign it off using the syntax 
`git commit -s` plus what you usually do.

And also to add a Release note message, so the easy way is to do :
`git commit - - amend -s -m "RELEASE_NOTES=[BUGFIX] Fixing progress bar to handle zero properly"`
And then `git push --force` to correct your branch.

defhacks:
Hopefully that's all correct now, it probably ended up being a bit more work having me do it, but I do appreciate the guidance and also I learned about `git commit -s`  :)

dominikschulz (approved):
LGTM

PR #2052, opened 2021-11-26
Update dependencies

dominikschulz:
Also adds a workaround for syntax errors in the password-rules.json
file.


pkg/pwgen/pwrules/gen.go
```diff
@@ -99,12 +103,51 @@ func fetchRules() (map[string]jsonRule, error) {
...
+		line := scanner.Text()
+		skip := false
+		// skip two broken entries
+		for _, needle := range []string{"fidelity.com", "hkexpress.com"} {
+			if strings.Contains(line, "\""+needle+"\"") {
+				skip = true
+			}
+		}
+		if skip {
+			scanner.Scan()
```
AnomalRoil, line 134: Are these two Scan really meant to be here?

AnomalRoil (dismissed):

pkg/pwgen/pwrules/gen.go
dominikschulz: Yes, I should add a comment.

This is my very naive way of working around seriously broken JSON.

It looks for the first line of a three line struct and skips the two other lines. Naive and maybe stupid, but effective.

pkg/pwgen/pwrules/gen.go
```diff
@@ -99,12 +103,55 @@ func fetchRules() (map[string]jsonRule, error) {
...
+	}
+
+	var buf bytes.Buffer
+	scanner := bufio.NewScanner(c.src)
+	for scanner.Scan() {
+		line := scanner.Text()
+		skip := false
+		// skip two broken entries. this is a terrible hack because
+		// the JSON is not valid.
+		for _, needle := range []string{"fidelity.com", "hkexpress.com"} {
```
AnomalRoil, line 130: Should we had a negative test that tests that `json.NewDecoder` fails when decoding these entries, so that once it won't fail any more we'll be able to remove the hardcoded exceptions??

AnomalRoil (approved):

pkg/pwgen/pwrules/gen.go
dominikschulz: I don't have the motivation to do that, I'm afraid.

But feel free to add something.
But remember that we don't want to download the source on every build so it would need to be part the genrule

AnomalRoil: I guess we could have a "monthly" GHA that checks if the data is still rotten or not.
I'll had an issue about it.

PR #2053, opened 2021-12-01
illegal base32 data could be lower case

vv01f:
so documenting it should be upper case to prevent people wondering why converting zu base32 doesnt help


dominikschulz (approved):
LGTM. Thank you!

PR #2054, opened 2021-12-01
Fixing Repology badge with new Gopass URL

AnomalRoil:
The badge was pointing to the old Gopass name that was mixing Gopass by aviau with ours.


dominikschulz (approved):
LGTM

PR #2064, opened 2021-12-23
Add wizard templates to gopass create

dominikschulz:
RELEASE_NOTES=[ENHANCEMENT] Template support for the create wizard


dominikschulz:
I guess this is still missing some docs to be really useful.

internal/action/config.go
```diff
@@ -37,6 +37,12 @@ func (s *Action) Config(c *cli.Context) error {
 func (s *Action) printConfigValues(ctx context.Context, needles ...string) {
 	m := s.cfg.ConfigMap()
 	for _, k := range filterMap(m, needles) {
+		// if only a single key is requested, print only the value
+		// useful for scriping, e.g. `$ cd $(gopass config path)`
```
AnomalRoil, line 41: Good idea! Was sometimes using `cut` or `tr`  to go to that path 😂

internal/create/helpers.go
```diff
@@ -0,0 +1,41 @@
...
+// way for use in the name of a secret
+func extractHostname(in string) string {
+	if in == "" {
+		return ""
+	}
+	// help url.Parse by adding a scheme if one is missing. This should still
+	// allow for any scheme, but by default we assume http (only for parsing)
+	urlStr := in
+	if !strings.Contains(urlStr, "://") {
+		urlStr = "http://" + urlStr
```
AnomalRoil, line 32: Why not assume `https` by default? Passwords are usually going over TLS secured connections.

internal/create/wizard.go
```diff
@@ -0,0 +1,310 @@
...
+				sec.SetPassword(password)
+			}
+		}
+
+		// select store
+		if store == "" {
+			store = cui.AskForStore(ctx, s)
+		}
+
+		// generate name, ask for override if already taken
```
AnomalRoil, line 221: I don't understand that comment, where does it ask for override?

AnomalRoil (commented):
Haven't tried it yet, but lgtm so far.

internal/create/helpers.go
dominikschulz: Because that's the shortest valid scheme. Please note the comment and look at the next few lines: This is really only used to temporarily construct a valid URL. We extract the hostname and throw away everything, i.e. the scheme doesn't escape the method.

Of course we could use https too, but I wouldn't make a difference.

internal/create/wizard.go
dominikschulz: Clarified.

PR #2065, opened 2021-12-24
Improve clipboard safety

### LINKED ISSUE #1816: Hide passwords from MacOS clipboard history
Passwords I copy appear in Alfred's clipboard history, and I imagine they would appear in other MacOS clipboard history managers like Maccy as well.

The standard seems to be that apps can set `org.nspasteboard.ConcealedType` when copying (which also requires using the Mac API). From an issue on bitwarden's tracker, I found a link to this PR on keepassxc which implemented the flag: https://github.com/keepassxreboot/keepassxc/pull/1202/files.

I know gopass uses attoto's module for clipboard access - so it might be worth adding the option upstream. However the module is made tiny by plugging directly to `pbcopy`, so using the clipboard api instead would be adding a lot of code; it might be worth adding mac-specific clipboard implementation here. Either way, I'm happy to code up a draft.

The alternative would be something like [this open PR on the alfred integration](https://github.com/gopasspw/gopass-alfred/pull/1) that uses Alfred to do the copying, but this only fixes the issue for those who use Alfred to copy passwords.

dominikschulz:
This commit finally incorporates the changes suggested by rianadon.

Additionally it switches the unclip checksum from SHA256 to Argon2id.
Although the risk is close to zero (i.e. if someone can access our
process env we're doomed anyway) but there is just no reason to not
do it properly.

RELEASE_NOTES=[ENHANCEMENT] Hide password on MacOS clipboards

Fixes #1816

### DIFF
--- a/pkg/clipboard/copy_darwin.go
+++ b/pkg/clipboard/copy_darwin.go
@@ -1,0 +1,77 @@
+//go:build darwin
+// +build darwin
+
+package clipboard
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+
+	"github.com/atotto/clipboard"
+	"github.com/gopasspw/gopass/pkg/debug"
+)
+
+func copyToClipboard(ctx context.Context, content []byte) error {
+	password := string(content)
+
+	// first try to copy via osascript, if that fails fallback to pbcopy
+	if err := copyViaOsascript(ctx, password); err != nil {
+		debug.Log("failed to copy via osascript: %s", err)
+		return clipboard.WriteAll(password)
+	}
+	return nil
+}
+
+func copyViaOsascript(ctx context.Context, password string) error {
+	args := []string{
+		// The Foundation library has the Objective C symbols for pasteboard
+		"-e", `use framework "Foundation"`,
+		// Need to use scripting additions for access to "do shell script"
+		"-e", "use scripting additions",
+		// type = a reference to the ObjC constant NSPasteboardTypeString
+		// which is needed to indentify clioboard contents as text
+		"-e", "set type to current application's NSPasteboardTypeString",
+		// pb = a reference to the system's pasteboard
+		"-e", "set pb to current application's NSPasteboard's generalPasteboard()",
+		// Must clear contents before adding a new item to pasteboard
+		"-e", "pb's clearContents()",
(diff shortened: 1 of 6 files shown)


pkg/clipboard/copy_darwin.go
```diff
@@ -0,0 +1,72 @@
...
+	r, w, err := os.Pipe()
+	if err != nil {
+		return err
+	}
+
+	cmd.ExtraFiles = []*os.File{r} // Receiving end of pipes is connected to fd#3
+	go func() {
+		defer w.Close()
+		io.WriteString(w, password) // Write the password to fd#3
+	}()
```
AnomalRoil, line 58: This could deserve a comment mentioning how line 60 is going to be waiting on a write on its pipe because of the above `cat 0<&3` and is therefore blocking until the `WriteString` is done, maybe. 
Also, don't we have a risk of the write "starting too soon" and the pipe not getting it all? Seems like it could be a race condition issue leading to "missing parts of password" or even it "blocking with no reason when copying" if the string is fully written before the command is actually run on line 60.

From: https://groups.google.com/g/alt.comp.lang.applescript/c/1v0s0Tkn1bM/m/b6VesyBKv6YJ
it seems we could be relying on `cat` and Stdin instead, which would allow us to get rid of the goroutine and use the "nicer":
```
cmd.Stdin = strings.NewReader(password)
```

AnomalRoil (dismissed):

pkg/clipboard/copy_darwin.go
dominikschulz: I couldn't get it to work with the `cmd.Stdin` approach but feel free to send a PR if you succeed.

wheleph:
I'm sorry for the seemingly dumb question but I'm wondering if this change ended up in 1.13.1. 

Comparing the dates of the merge and the release, I assume that it did. But on the other hand, it's not mentioned in the [release notes](https://github.com/gopasspw/gopass/releases/tag/v1.13.1) and more importantly, when I test it locally with [Maccy](https://maccy.app/) I still can see my passwords there although I did add `org.nspasteboard.ConcealedType` in the app's settings:

Screenshot 2022-02-20 at 01 06 55

wheleph:
I'm sorry for my ignorance. I've just noticed that the milestone for this PR is 1.14.0.

PR #2072, opened 2021-12-27
GCI-ing imports

AnomalRoil:
Running gci in order to make sure we have properly deterministic imports with a nice and consistent formatting.

See: https://github.com/daixiang0/gci

### DIFF
--- a/internal/action/action.go
+++ b/internal/action/action.go
@@ -5,12 +5,11 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/blang/semver/v4"
 	"github.com/gopasspw/gopass/internal/config"
 	"github.com/gopasspw/gopass/internal/reminder"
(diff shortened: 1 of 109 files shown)


AnomalRoil:
I'll rebase, I wanted to work without having to install go 1.18 beta 1, but I see I won't be able to do that without having conflicts and annoying merge commits.

dominikschulz:
Looks good. Can you please add it to our Makefile as well?
Either as a check in the `codequality` target, as a fix in the `fmt` target or both. As you prefer.

Thank you!

AnomalRoil:
Yes, this is actually part of my new ongoing effort to migrate from our current Makefile codequality check to a more streamlined golangci-lint based one, which integrates nicely with Github :) 
More PRs to come soon.

dominikschulz (approved):
LGTM

Issue #2073, opened 2021-12-28
Migrate our Codequality checks to GHA and Golangci-lint


AnomalRoil:
We could use golangci-lint that integrates nicely with Github Actions instead of our current Makefile-based codequality check.

Main issue: most linters aren't yet supporting Go 1.18, and the current GHA for golangci-lint is built with Go 1.17, whose `go fmt` doesn't like generics.

Todo, test all linters and see if they work or not:
 - [x] asciicheck
 - [x] dogsled (issue with two things)
 - [x] dupl (some dups in tests)
 - [x] forcetypeassert
 - [x] gci: #2072 
 - [x] gochecknoinits (we use a lot of init, prolly to be kept disabled)
 - [x] goconst (annoying?)
 - [x] gocyclo (one issue to correct in mkActFunc)
 - [x] godot (annoying)
 - [x] godox (we have a few todo...)
 - [x] gofmt
 - [x] goimports
 - [x] gomoddirectives
 - [x] gomodguard
 - [x] goprintffuncname
 - [x] ifshort
 - [x] ineffassign
 - [x] misspell
 - [x] nakedret
 - [x] nestif (some work to refactor)
 - [x] nolintlint
 - [x] paralleltest (lots of work to pass)
 - [x] prealloc
 - [x] revive
 - [x] rowserrcheck
 - [x] predeclared
 - [x] staticcheck
 - [x] structcheck
 - [x] stylecheck
 - [x] tagliatelle
 - [x] testpackage
 - [x] thelper
 - [x] tparallel
 - [x] typecheck
 - [x] unconvert
 - [x] varcheck
 - [x] wastedassign
 - [x] wrapcheck

Currently not working with Go 1.18:
 - deadcode
 - depguard
 - durationcheck
 - errcheck
 - errorlint
 - exhaustive
 - exportloopref
 - gocritic
 - goerr113
 - gosec
 - gosimple
 - govet
 - importas
 - makezero
 - nilerr
 - rowserrcheck
 - staticcheck
 - structcheck
 - stylecheck
 - thelper
 - tparallel
 - typecheck
 - unconvert
 - unparam
 - unused
 - varcheck
 - wastedassign
 - wrapcheck

PR #2075, opened 2021-12-28
Migrating to Golangci-Lint

AnomalRoil:
Adding a GHA that is currently not running automatically until it's supporting Go 1.18 (can be launched manually to test it).

Adding the proper config files for Golangci-lint, with the linters whitelisted in #2073, still some extra work to test them all. 

Removing codequality from CI in Makefile, since it's now meant to be handled in a GHA.
Phasing out some of the codequality linters too since they are in golangci already.


dominikschulz (approved):
LGTM. Thanks!

Issue #1162, opened 2019-08-10
[FR] Use flag to select identity when using gopass clone


M-Gregoire:
First of all, thank you for the awesome tool that Gopass is!

### Summary
I'd like to clone an existing password store (git) using one command without the need for extra input.

### Steps To Reproduce
```
gopass clone git@my-git-repo.git
Configuring git repository ...
Use Identity (my@email.com) for password store git config? [Y/n/q]: Y
Your password store is ready to use! Have a look around: `gopass list`
```

### Expected behavior
```
gopass clone --id my@email.com git@my-git-repo.git
Configuring git repository ...
Your password store is ready to use! Have a look around: `gopass list`
```

A flag that would allow to specify either:
- The email address associated with the GPG key
- The 8 character ID found by typing "gpg --list-keys" and looking at the "pub" line
- The GPG key fingerprint found by typing "gpg --fingerprint" and removing all of the spaces

just like `gopass init` would be awesone.

### Note
I've been looking through the documentation and the source code, it seems like it's note implemented but might have missed it.

M-Gregoire:
Just noticed one can simply do the following:
```
gopass --yes clone git@my-git-repo.git
```
To clone the repo in one line. However, this doesn't allow to choose which key to use if said repo has multiple recipients and you have multiple keys associated with the repo.

I don't see any real use case for this so closing this issue but feel free to open it again if needed.

M-Gregoire:
Sorry for this long post, seems like I'm a bit confused about the use of `git clone`.
I'm trying to create `~/.password-store` based on an already existing store on git.

I first tried 
```
gopass --yes clone git@my-git-repo.git
```
this returns
```
Configuring git repository ...
Your password store is ready to use! Have a look around: `gopass list`
```

But nothing gets created (I guess this is an issue, an error message should probably be returned). 

I then noticed the docs says:
> Please note that all cloned repositories must already have been initialized with gopass. (See the previous section for more details.)

So I tried the following:
```
# This commands works fine and create an empty password store at ~/.password-store
gopass --yes init <key-id>

gopass --yes clone git@my-git-repo.git
```
But this last commands returns: `Error: Can not clone git@my-git-repo.git to the root store, as this store is already initialized. Please try cloning to a submount: `gopass clone git@my-git-repo.git sub`

I'm confused, the docs says 
> all cloned repositories must already have been initialized 


and this commands says


> Can not clone ... to the root store, as this store is already initialized

Am I missing something?  
[...]

PR #2077, opened 2021-12-28
Enable additional golangci-lint godot checks

dominikschulz:
Enable additional golangci-lint checks

Refactors code to make them pass.

Only comment changes.

RELEASE_NOTES=n/a

### DIFF
--- a/internal/backend/crypto/age/identities.go
+++ b/internal/backend/crypto/age/identities.go
@@ -19,7 +19,7 @@ var (
 	idRecpCacheKey = "identity"
 )
 
-// Identities returns all identities, used for decryption
+// Identities returns all identities, used for decryption.
 func (a *Age) Identities(ctx context.Context) ([]age.Identity, error) {
 	if !ctxutil.HasPasswordCallback(ctx) {
 		debug.Log("no password callback found, redirecting to askPass")
@@ -71,7 +71,7 @@ func (a *Age) IdentityRecipients(ctx context.Context) ([]age.Recipient, error) {
 	return r, nil
 }
 
-// GenerateIdentity creates a new identity
+// GenerateIdentity creates a new identity.
 func (a *Age) GenerateIdentity(ctx context.Context, _ string, _ string, pw string) error {
 	if pw != "" {
 		ctx = ctxutil.WithPasswordCallback(ctx, func(prompt string, confirm bool) ([]byte, error) {
@@ -82,7 +82,7 @@ func (a *Age) GenerateIdentity(ctx context.Context, _ string, _ string, pw strin
 	return err
 }
 
-// ListIdentities lists all identities
+// ListIdentities lists all identities.
 func (a *Age) ListIdentities(ctx context.Context) ([]string, error) {
 	debug.Log("checking existing identities")
 	ids, err := a.getAllIdentities(ctx)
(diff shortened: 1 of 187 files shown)


internal/backend/crypto/age/identities.go
```diff
@@ -201,14 +201,14 @@ func (a *Age) getAllIdentities(ctx context.Context) (map[string]age.Identity, er
...
+	// merge both.
 	for k, v := range ssh {
 		native[k] = v
 	}
 	debug.Log("got %d merged identities", len(native))
 
-	// TODO add passage identities, too
-	// $HOME/.passage/identities
+	// TODO add passage identities, too.
+	// $HOME/.passage/identities.
```
AnomalRoil, line 211: This trailing dot seems strange...

internal/backend/crypto/gpg/colons/parse_colons.go
```diff
@@ -10,11 +10,11 @@ import (
 )
 
 var (
-	// John Doe (user) <john.doe@example.com>
+	// John Doe (user) <john.doe@example.com>.
```
AnomalRoil, line 13: The regex isn't matching the point, not sure we should have one there. 
Same for the two other comment below.
Maybe add a //nolint:godot for that part?

internal/backend/storage.go
```diff
@@ -49,22 +49,23 @@ type Storage interface {
...
-				return st, nil
-			}
-			be, err := StorageRegistry.Get(FS)
-			if err != nil {
-				return nil, err
-			}
-			return be.Init(ctx, path)
+	// The call to HasStorageBackend is important since GetStorageBackend will always return FS
+	// if nothing is found in the context.
+	if be, err := StorageRegistry.Get(GetStorageBackend(ctx)); HasStorageBackend(ctx) && err == nil {
```
AnomalRoil, line 56: Could it be that the weird `clone` behaviour where it's not properly initializing the Git config because it's probably detecting the file system as being `fs` instead of `gitfs` be related to that part and the below call to Init? See the issue in my comment: https://github.com/gopasspw/gopass/issues/1162#issuecomment-998829861

internal/backend/storage/fs/loader.go
```diff
@@ -37,20 +37,26 @@ func (l loader) Init(ctx context.Context, path string) (backend.Storage, error)
...
 	return fmt.Errorf("dir not found")
 }
 
+// Priority returns the priority of this backend. Should always be higher than
+// the more specific ones, e.g. gitfs.
 func (l loader) Priority() int {
 	return 50
 }
+
 func (l loader) String() string {
```
AnomalRoil, line 60: Exported functions should be commented, no?

AnomalRoil (commented):
75 files out of 185 for now, a few initial comments already.

internal/backend/storage/fs/loader.go
dominikschulz: They implement an interface, so IIRC that's an exception.

AnomalRoil (approved):
LGTM, that was a hella lot of files and dots to check xD

dominikschulz:
Thanks a lot!

Once we enable golangci-lint enforcement we won't need to do that many again.

PR #2084, opened 2021-12-30
Fix some TODOs

dominikschulz:
Fix some TODOs


internal/backend/storage/fs/rcs.go
```diff
@@ -62,14 +62,17 @@ func (s *Store) Revisions(context.Context, string) ([]backend.Revision, error) {
 		}}, nil
 }
 
-// GetRevision is not implemented.
-func (s *Store) GetRevision(context.Context, string, string) ([]byte, error) {
-	return []byte("foo\nbar"), nil
```
AnomalRoil, line 67: Nice catch xD

AnomalRoil (approved):
LGTM

internal/backend/storage/fs/rcs.go
dominikschulz: I put that in during early development. And never removed it :(

PR #2085, opened 2021-12-31
Enforce TLSv1.3 for outgoing HTTPS connections

dominikschulz:
Only applies to connections to GitHub that supports it.
So no compatibility concerns. Middleboxes might cause
trouble but they are terrible and must go away.

RELEASE_NOTES=[ENHANCEMENT] Enforce TLSv1.3


AnomalRoil (approved):
LGTM

PR #2091, opened 2022-01-09
Add env variables for custom clipboard commands

### LINKED ISSUE #2042: Add configuration options for custom clipboard commands
## Problem

*This is not a feature request but a request for discussion. I can implement it myself.*

Currently, `gopass` only handles unclipping for a small set of clipboard managers, for example  [KDE Klipper](https://github.com/gopasspw/gopass/blob/8586105f2704c2d02e515c154c8e0d0c0a2757d5/pkg/clipboard/unclip_linux.go#L20). For other clipboard managers the `unclip` procedure just doesn't work. Some clipboard managers support passwords out of the box (masking them), but `gopass` would need to support the way passwords are added to the clipboard manager.

For example, [GPaste](https://github.com/Keruspe/GPaste):
```bash
$ echo -n "This is a test" | wl-copy
$ gpaste-client
254de975-949b-4d8e-8f71-8ddf566ee97c: This is a test
$ gpaste-client add-password google.com secretString
$ gpaste-client
c6ef91f8-235f-4b6d-b8c7-108acd9c0f77: [Password] google.com
254de975-949b-4d8e-8f71-8ddf566ee97c: This is a test
$ gpaste-client delete-password google.com
$ gpaste-client
254de975-949b-4d8e-8f71-8ddf566ee97c: This is a test
```

## Suggested solution implementation

It would be great if one could specify custom commands for the `clip` and `unclip` procedure.

To archive that, I would add two new optional config options to [`internal/config/config.go`](https://github.com/gopasspw/gopass/blob/master/internal/config/config.go#L19):
* `clipcmd`: `gpaste-client add-password "{.Name}" "{.Password}"`
* `unclipcmd`: `gpaste-client delete-password "{.Name}"`

Both are template strings.

If `clipcmd` is set I would try to execute it in [`pkg/clipboard/clipboard.go`](https://github.com/gopasspw/gopass/blob/master/pkg/clipboard/clipboard.go#L30), if it fails the normal procedures are used.
Same for `unclipcmd`: If it is set I would try to execute it in [`pkg/clipboard/unclip.go`](https://github.com/gopasspw/gopass/blob/master/pkg/clipboard/unclip.go#L28), if it fails the normal procedures are used.

---

Does that sound reasonable?

hashworks:
Adds `GOPASS_CLIPBOARD_COPY_CMD` and `GOPASS_CLIPBOARD_CLEAR_CMD` environment variables which are called instead of the normal implementation if set. The commands receive the name of the password as their first parameter and the password or its checksum on `STDIN`.

Resolves #2042.

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -4,35 +4,37 @@
 
 Some configuration options are only available through setting environment variables.
 
-| **Option**              | **Type** | **Description**                                                                                              |
-|-------------------------|----------|--------------------------------------------------------------------------------------------------------------|
-| `CHECKPOINT_DISABLE`    | `bool`   | Set to any non-empty value to disable calling the GitHub API when running `gopass version`.                  |
-| `GOPASS_DEBUG`          | `bool`   | Set to any non-empty value to enable verbose debug output                                                    |
-| `GOPASS_DEBUG_LOG` | `string` | Set to a filename to enable debug logging |
-| `GOPASS_DEBUG_LOG_SECRETS` | `bool` | Set to any non-empty value to enable logging of credentials |
-| `GOPASS_DEBUG_FUNCS` | `string` | Comma separated filter for console debug output (functions) |
-| `GOPASS_DEBUG_FILES` | `string` | Comma separated filter for console debug output (files) |
-| `GOPASS_UMASK`          | `octal`  | Set to any valid umask to mask bits of files created by gopass                                               |
-| `GOPASS_GPG_OPTS`       | `string` | Add any extra arguments, e.g. `--armor` you want to pass to GPG on every invocation                          |
-| `GOPASS_EXTERNAL_PWGEN` | `string` | Use an external password generator. See [Features](features.md#using-custom-password-generators) for details |
-| `GOPASS_CHARACTER_SET`  | `bool`   | Set to any non-empty value to restrict the characters used in generated passwords                            |
-| `GOPASS_CONFIG`         | `string` | Set this to the absolute path to the configuration file                                                      |
-| `GOPASS_HOMEDIR`        | `string` | Set this to the absolute path of the directory containing the `.config/` tree                                |
-| `GOPASS_FORCE_UPDATE`   | `bool`   | Set to any non-empty value to force an update (if available)                                                 |
-| `GOPASS_NO_NOTIFY`      | `bool`   | Set to any non-empty value to prevent notifications                                                          |
-| `GOPASS_NO_REMINDER`      | `bool`   | Set to any non-empty value to prevent reminders                                                          |
+| **Option**                   | **Type** | **Description**                                                                                                  |
+| ---------------------------- | -------- | ---------------------------------------------------------------------------------------------------------------- |
+| `CHECKPOINT_DISABLE`         | `bool`   | Set to any non-empty value to disable calling the GitHub API when running `gopass version`.                      |
+| `GOPASS_DEBUG`               | `bool`   | Set to any non-empty value to enable verbose debug output                                                        |
+| `GOPASS_DEBUG_LOG`           | `string` | Set to a filename to enable debug logging                                                                        |
+| `GOPASS_DEBUG_LOG_SECRETS`   | `bool`   | Set to any non-empty value to enable logging of credentials                                                      |
+| `GOPASS_DEBUG_FUNCS`         | `string` | Comma separated filter for console debug output (functions)                                                      |
+| `GOPASS_DEBUG_FILES`         | `string` | Comma separated filter for console debug output (files)                                                          |
+| `GOPASS_UMASK`               | `octal`  | Set to any valid umask to mask bits of files created by gopass                                                   |
+| `GOPASS_GPG_OPTS`            | `string` | Add any extra arguments, e.g. `--armor` you want to pass to GPG on every invocation                              |
+| `GOPASS_EXTERNAL_PWGEN`      | `string` | Use an external password generator. See [Features](features.md#using-custom-password-generators) for details     |
+| `GOPASS_CHARACTER_SET`       | `bool`   | Set to any non-empty value to restrict the characters used in generated passwords                                |
+| `GOPASS_CONFIG`              | `string` | Set this to the absolute path to the configuration file                                                          |
+| `GOPASS_HOMEDIR`             | `string` | Set this to the absolute path of the directory containing the `.config/` tree                                    |
+| `GOPASS_FORCE_UPDATE`        | `bool`   | Set to any non-empty value to force an update (if available)                                                     |
+| `GOPASS_NO_NOTIFY`           | `bool`   | Set to any non-empty value to prevent notifications                                                              |
+| `GOPASS_NO_REMINDER`         | `bool`   | Set to any non-empty value to prevent reminders                                                                  |
+| `GOPASS_CLIPBOARD_COPY_CMD`  | `string` | Use an external command to copy a password to the clipboard. See [GPaste](usecases/gpaste.md) for an example     |
+| `GOPASS_CLIPBOARD_CLEAR_CMD` | `string` | Use an external command to remove a password from the clipboard. See [GPaste](usecases/gpaste.md) for an example |
 
 Variables not exclusively used by gopass
 
 | **Option**             | **Type** | **Description**                                                                                        |
-|------------------------|----------|--------------------------------------------------------------------------------------------------------|
+| ---------------------- | -------- | ------------------------------------------------------------------------------------------------------ |
 | `PASSWORD_STORE_DIR`   | `string` | absolute path containing the password store (a directory). Only supported during initialization!       |
 | `PASSWORD_STORE_UMASK` | `string` | Set to any valid umask to mask bits of files created by gopass (GOPASS_UMASK has precedence over this) |
 | `EDITOR`               | `string` | command name to execute for editing password entries                                                   |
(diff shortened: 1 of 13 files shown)


docs/config.md
```diff
@@ -4,35 +4,37 @@
...
+| `GOPASS_CONFIG`              | `string` | Set this to the absolute path to the configuration file                                                          |
+| `GOPASS_HOMEDIR`             | `string` | Set this to the absolute path of the directory containing the `.config/` tree                                    |
+| `GOPASS_FORCE_UPDATE`        | `bool`   | Set to any non-empty value to force an update (if available)                                                     |
+| `GOPASS_NO_NOTIFY`           | `bool`   | Set to any non-empty value to prevent notifications                                                              |
+| `GOPASS_NO_REMINDER`         | `bool`   | Set to any non-empty value to prevent reminders                                                                  |
+| `GOPASS_CLIPBOARD_COPY_CMD`  | `string` | Use an external command to copy a password to the clipboard. See [GPaste](usecases/gpaste.md) for an example     |
+| `GOPASS_CLIPBOARD_CLEAR_CMD` | `string` | Use an external command to remove a password from the clipboard. See [GPaste](usecases/gpaste.md) for an example |
```
hashworks, line 25: The above two lines are the only changes, the rest are reformattings.

pkg/clipboard/clipboard_test.go
```diff
@@ -3,21 +3,34 @@ package clipboard
...
+
+	buf := &bytes.Buffer{}
+	out.Stdout = buf
+
+	assert.Error(t, CopyTo(ctx, "foo", []byte("bar"), 1))
+}
+
+func TestUnsupportedCopyToClipboard(t *testing.T) {
 	_ = os.Setenv("GOPASS_NO_NOTIFY", "true")
+	_ = os.Setenv("GOPASS_CLIPBOARD_COPY_CMD", "")
```
hashworks, line 33: Since we might set the environment variable in a previous test we need to "unset" it.

tests/show_test.go
```diff
@@ -106,13 +106,11 @@ func TestShow(t *testing.T) {
 		assert.NoError(t, err)
 		assert.NotContains(t, out, "***")
 		assert.NotContains(t, out, "safecontent=true")
-		assert.NotContains(t, out, "and")
```
hashworks, line 109: Had to adjust the integration test, please confirm that this is safe.

pkg/clipboard/clipboard_test.go
```diff
@@ -3,21 +3,34 @@ package clipboard
 import (
 	"bytes"
 	"context"
+	"github.com/atotto/clipboard"
```
AnomalRoil, line 6: The move here is unnecessary and will make [gci](https://github.com/daixiang0/gci) unhappy IMO.

pkg/clipboard/clipboard.go
```diff
@@ -16,27 +17,58 @@ var (
...
+			return fmt.Errorf("failed to close STDIN of clipboard copy process: %w", err)
+		}
+
+		if err := clipboardProcess.Wait(); err != nil {
+			_ = notify.Notify(ctx, "gopass - clipboard", "failed to execute $CLIPBOARD_COPY_CMD")
+			return fmt.Errorf("failed to execute $CLIPBOARD_COPY_CMD: %w", err)
+		}
+	} else {
+		if clipboard.Unsupported {
+			out.Printf(ctx, "%s", ErrNotSupported)
```
AnomalRoil, line 57: This should probably be a `out.Errorf` instead.

tests/show_test.go
AnomalRoil: This is really strange, it should not happen: when safecontent is set to `true`,  the "and" in the twoliner should not be displayed.

AnomalRoil: Did that fail by matching the "and" in this error?
> ErrNotSupported = fmt.Errorf("WARNING: No clipboard available. Install " + Helpers + ", provide $GOPASS_CLIPBOARD_COPY_CMD and $GOPASS_CLIPBOARD_CLEAR_CMD

Please, revert this and change the content of the twoliner test to have something else that `and\nmore stuff` to see if it works then.

AnomalRoil (changes requested):

pkg/clipboard/clipboard.go
hashworks: I can adjust it, but please note that I didn't change this.

AnomalRoil: This could also solve the below issue about the integration test not working anymore.

hashworks: It won't – the test currently looks at `cmd.CombinedOutput()`, which combines `STDOUT` and `STDERR`.

pkg/clipboard/clipboard_test.go
hashworks: Fixed in 7ffa1037

pkg/clipboard/clipboard.go
hashworks: Fixed in 7ffa1037

tests/show_test.go
hashworks: Fixed in ecc2333b

pkg/clipboard/clipboard.go
```diff
@@ -16,27 +17,58 @@ var (
...
 // CopyTo copies the given data to the clipboard and enqueues automatic
 // clearing of the clipboard.
 func CopyTo(ctx context.Context, name string, content []byte, timeout int) error {
-	if clipboard.Unsupported {
-		out.Printf(ctx, "%s", ErrNotSupported)
-		_ = notify.Notify(ctx, "gopass - clipboard", fmt.Sprintf("%s", ErrNotSupported))
-		return nil
-	}
+	clipboardCopyCMD := os.Getenv("GOPASS_CLIPBOARD_COPY_CMD")
+	if clipboardCopyCMD != "" {
```
dominikschulz, line 27: This function is getting too big to comprehend quickly.

Please turn the if into a single line statement, e.g. `if cmd := os.Getenv("..."); cmd != "" {`and then call into another function, e.g. `if err := copyToCmd(ctx, ..., cmd);`.

pkg/clipboard/clipboard_test.go
```diff
@@ -16,8 +16,21 @@ import (
...
+
+	buf := &bytes.Buffer{}
+	out.Stdout = buf
+
+	assert.Error(t, CopyTo(ctx, "foo", []byte("bar"), 1))
+}
+
+func TestUnsupportedCopyToClipboard(t *testing.T) {
+	_ = os.Setenv("GOPASS_NO_NOTIFY", "true")
+	_ = os.Setenv("GOPASS_CLIPBOARD_COPY_CMD", "")
```
dominikschulz, line 33: We have a nice [helper](https://github.com/gopasspw/gopass/blob/master/internal/action/clone_test.go#L121) to temporarily unset env vars and restore them later. Maybe we should have that also for setting vars in tests.

pkg/clipboard/unclip.go
```diff
@@ -11,37 +13,68 @@ import (
 )
 
 // Clear will attempt to erase the clipboard.
-func Clear(ctx context.Context, checksum string, force bool) error {
-	if clipboard.Unsupported {
-		return ErrNotSupported
-	}
+func Clear(ctx context.Context, name string, checksum string, force bool) error {
+	clipboardClearCMD := os.Getenv("GOPASS_CLIPBOARD_CLEAR_CMD")
```
dominikschulz, line 17: Similar to `CopyTo`: I think this is getting too long to comprehend.

hashworks:
`TestShowAutoClip` seems to test STDOUT, not STDERR. Fixing…

dominikschulz (changes requested):

pkg/clipboard/clipboard.go
hashworks: Fixed in a0986a93

pkg/clipboard/unclip.go
hashworks: Fixed in a0986a93

pkg/clipboard/clipboard_test.go
hashworks: Introduced in bfee59e9

AnomalRoil:
Once all the tests pass even on Windows, I'd say it's ready to be merged 😄 

Thank you for the nice addition.

AnomalRoil (dismissed):

pkg/clipboard/unclip.go
```diff
@@ -11,37 +12,43 @@ import (
 )
 
 // Clear will attempt to erase the clipboard.
-func Clear(ctx context.Context, checksum string, force bool) error {
-	if clipboard.Unsupported {
+func Clear(ctx context.Context, name string, checksum string, force bool) error {
```
dominikschulz, line 15: The nesting in this method is not very idiomatic. In Go you'd try to avoid else and else if constructs when possible.

Here you could just copy line 54/55 in the `clipboardClearCMD` branch and have it return there.

Then the unsupported case can just be a simple if and the rest doesn't need to be in an if statement at all.

dominikschulz (changes requested):

pkg/clipboard/unclip.go
hashworks: Resolved in a0c81ae6

dominikschulz (approved):
LGTM. Thanks a lot!

PR #2092, opened 2022-01-09
Fossil support

### LINKED ISSUE #2022: Support Fossil SCM
We should consider adding support for the fossil SCM.
It has a much simpler model than git and handles a few things ootb which we need to work around with git (e.g. autosync, no staging area, ...): https://fossil-scm.org/home/doc/trunk/www/quickstart.wiki

dominikschulz:
This commit adds HIGHLY EXPERIMENTAL support for the Fossil SCM.
It's barely tested but seems to support the most basic use cases
already. A lot of Fossil specifics aren't supported, yet. Cloning
for examples isn't supported so far.

This should be threated as an experiment only so far. If there is some interest we
will probably built this out further in the future. Fossil definitely has some
very nice properties and should be much simpler to support than git.

Fixes #2022

RELEASE_NOTES=[EXPERIMENTAL] Support the Fossil SCM

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 )
 
 require (
+	bitbucket.org/creachadair/stringset v0.0.10 // indirect
 	filippo.io/edwards25519 v1.0.0-rc.1 // indirect
 	github.com/cpuguy83/go-md2man/v2 v2.0.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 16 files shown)


AnomalRoil (approved):
I never tried fossil-scm so far, so not much to say. 

Overall, I'm a bit sad we don't have a nice go lib to handle that and have to resort to cmd.Exec, but I guess that's fine.

Also no tests, but it's difficult to do that kind of tests when you're calling external tools and all.

dominikschulz:
If we ever find something that's pure Go that'd be great.
But fossil is based on SQLite and that isn't available in Go without CGO.

I'm primarily interested in this to keep exploring possible alternatives to git.
I understand that many users won't migrate away from git. But if I find something
that works better I'll happily move over myself.

dertuxmalwieder:
Just yesterday I complained on the IRC (of a different community) that my attempts at getting my own Git server running were fruitless. I would have to *migrate to* Git first... :-)

Thank you for your commit! From what I can see, it **should** work fine...

dominikschulz:
@dertuxmalwieder Please note that this is really meat as an early preview. A lot of things don't work yet - or are at least untested (e.g. cloning a repo).

PR #2095, opened 2022-01-10
Use github.com/twpayne/go-pinentry

dominikschulz:
This commit replaces our own pinentry implementation with go-pinentry
since that is more complete and has much better test coverage. This
change will allow us to deprecate our own pinentry implementation.

RELEASE_NOTES=[CLEANUP] Deprecate gopasspw/pinentry


AnomalRoil (approved):
LGTM

PR #2102, opened 2022-01-12
Pass git commands directly to git

### LINKED ISSUE #2097: Make `gopass git` pass commands directly to git
Instead of adding another subcommand like in https://github.com/gopasspw/gopass/issues/2088 I'd consider just bringing back the original `gopass git` implementation that would just invoke git in the right place.

dominikschulz:
This allows using all git commands.

RELEASE_NOTES=[ENHANCEMENT] gopass git invokes git directly

Fixes #2097

### DIFF
--- a/internal/action/git.go
+++ b/internal/action/git.go
@@ -1,0 +1,32 @@
+package action
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/gopasspw/gopass/internal/out"
+	"github.com/gopasspw/gopass/pkg/ctxutil"
+	"github.com/urfave/cli/v2"
+)
+
+// Git passes the git command to the underlying backend.
+func (s *Action) Git(c *cli.Context) error {
+	ctx := ctxutil.WithGlobalFlags(c)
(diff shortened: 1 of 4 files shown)


internal/action/git.go
```diff
@@ -0,0 +1,35 @@
...
+func (s *Action) Git(c *cli.Context) error {
+	ctx := ctxutil.WithGlobalFlags(c)
+	store := c.String("store")
+
+	sub, err := s.Store.GetSubStore(store)
+	if err != nil {
+		return ExitError(ExitGit, err, "failed to get sub store %s: %s", store, err)
+	}
+	if sub == nil {
+		panic("boom")
```
AnomalRoil, line 23: This seems not super descriptive. Could it happen on a non-initialized setup, maybe? Or is it really supposed to never happen?

AnomalRoil (approved):

internal/action/git.go
dominikschulz: Haha, sorry. Forgot this

PR #2103, opened 2022-01-12
Add age subcommand

### LINKED ISSUE #2098: Add `gopass age` or simliar
When using `gopass` with the `age` backend we manage identities automatically on setup but don't expose an easy way to manipulate (not even list) the identities store in an encrypted file.

It's possible to decrypt it with age directly and manipulate the content - it's only private key strings - but that shouldn't be necessary.
We should expose at least:

* a list operation
* a generate operation to create new keys
* maybe delete, too

dominikschulz:
Fixes #2098

RELEASE_NOTES=[ENHANCEMENT] Add age subcommand


internal/backend/crypto/age/commands.go
```diff
@@ -0,0 +1,94 @@
...
+								if err != nil {
+									return err
+								}
+								victim := c.Args().First()
+
+								ids, _ := a.Identities(ctx)
+								newIds := make([]string, 0, len(ids))
+
+								for _, id := range ids {
+									if x, ok := id.(*age.X25519Identity); ok && x.Recipient().String() == victim {
```
AnomalRoil, line 80: I've probably missed it in an earlier PR, but didn't we use to support more than just X25519 identities? 

Is that going to be "backward compatible" with SSH identities that we used to support and that could be using P-256 for instance? Or even worse, `age` is compatible with `ssh-rsa` IIRC.

internal/backend/crypto/age/identities.go
```diff
@@ -159,7 +159,7 @@ func (a *Age) addIdentity(ctx context.Context) ([]age.Identity, error) {
 }
 
 func (a *Age) saveIdentities(ctx context.Context, ids []string, newFile bool) error {
-	if !ctxutil.HasPasswordCallback(ctx) {
+	if !ctxutil.HasPasswordCallback(ctx) && ctxutil.IsInteractive(ctx) {
```
AnomalRoil, line 162: So, if:
- we are NOT interactive
- and we do NOT have a password callback function

Then the `encryptFile` below would call the non-existent password callback:
https://github.com/gopasspw/gopass/blob/f6237d4003955fa808285eab9c91fc14d9d20fef/internal/backend/crypto/age/encrypt.go#L70-L74

And that would fail with 
https://github.com/gopasspw/gopass/blob/f6237d4003955fa808285eab9c91fc14d9d20fef/pkg/ctxutil/ctxutil.go#L431

This is not obvious at first, maybe a comment would be nice.

internal/backend/storage/gitfs/commands.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func (l loader) Commands(i func(*cli.Context) error, s func(string) (string, error)) []*cli.Command {
+	return []*cli.Command{
+		{
+			Name:  "git",
+			Usage: "Run a git command inside a password store",
+			Description: "" +
+				"If the password store is a git repository, execute a git command " +
+				"specified by git-command-args.",
```
AnomalRoil, line 21: BTW, `git-command-args` is mentioned nowhere in the output.

Maybe change the `Usage` part to contain it: 
```
gopass git store git-command-args...
```

AnomalRoil (dismissed):

internal/action/commands.go
```diff
@@ -980,4 +965,42 @@ func (s *Action) GetCommands() []*cli.Command {
 			Action: s.Version,
 		},
 	}
+
+	// crypto and storage backends can add their own commands if they need to
```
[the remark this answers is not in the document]
dominikschulz: :smile:

internal/backend/crypto/age/commands.go
dominikschulz: added a comment to clarify. let me know if that doesn't explain it well enough.

AnomalRoil (approved):

dylangerdaly:
Can this be considered stable enough to migrate to? I'd like to retire my Password Store GPG Key and migrate to `age`.

However if something happens and the format changes I'll wait.

PR #2109, opened 2022-01-14
Body only entries are detected now by show -o

### LINKED ISSUE #2106: Gopass `show -o` has misleading error for body-only entries
Assuming we have an entry `x` that only has a body and no password on its first line, e.g.:
```

---
some: data
```
 then running `gopass show -o x` will return: `Error: empty secret`, which could be an issue if a user assume the entire entry is empty and deletes it for example.

But the entry could actually contain for instance a binary blob encoded using `cat file | gopass cat x`, or just a body made of yaml data as in the above.

## Expected behaviour

The error message should probably be different in the case where the `Body` of the entry is non-empty and only the `Password` is empty.
Maybe `Error: empty password, entry has data only in its body, or contains a binary blob accessible with gopass cat.`

AnomalRoil:
Fixes #2106

### DIFF
--- a/internal/action/show.go
+++ b/internal/action/show.go
@@ -203,6 +203,9 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 		return pw, "", nil
 	}
 	if IsPasswordOnly(ctx) {
+		if pw == "" && fullBody != "" {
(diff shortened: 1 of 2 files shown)


dominikschulz (approved):
LGTM

PR #2110, opened 2022-01-14
Handle from prefix correctly during move

### LINKED ISSUE #2079: Gopass `mv` moves fuzzy names even on exact match and doesn't rename directories
### Steps To Reproduce
```
$ GOPASS_HOMEDIR=~/test gop generate comm/test 12 
$ GOPASS_HOMEDIR=~/test gop generate comm/test2 12
$ GOPASS_HOMEDIR=~/test gop generate communication/t1 12
$ GOPASS_HOMEDIR=~/test gop ls
gopass
 ├── communication/
 │          └── t1
 └── comm/
        ├── test
        └── test2
       
$ GOPASS_HOMEDIR=~/test gop mv comm email 
$ GOPASS_HOMEDIR=~/test gop ls
gopass
└── email/
    └── comm/
        ├── test
        ├── test2
        └── unication/
            └── t1
```

### Expected behavior

Should get instead:
```
$ GOPASS_HOMEDIR=~/test gop ls
gopass
 ├── communication/
 │          └── t1
 └── email/
        ├── test
        └── test2
       
```

Notice also that the folder `comm` wasn't renamed but actually moved into a new folder instead.

This is not similar to how Unix's `mv`works, which is what we aim for with `gopass mv` IIRC.

dominikschulz:
Fixes #2079

RELEASE_NOTES=[BUGFIX] Handle from prefix correctly on mv

### DIFF
--- a/internal/store/root/move.go
+++ b/internal/store/root/move.go
@@ -17,6 +17,7 @@ import (
 // supported. Each entry has to be decoded and encoded for the destination
 // to make sure it's encrypted for the right set of recipients.
 func (r *Store) Copy(ctx context.Context, from, to string) error {
+	debug.Log("Copy %s to %s", from, to)
 	return r.move(ctx, from, to, false)
 }
 
@@ -25,6 +26,7 @@ func (r *Store) Copy(ctx context.Context, from, to string) error {
 // for the destination store with the right set of recipients and remove it
 // from the old location afterwards.
 func (r *Store) Move(ctx context.Context, from, to string) error {
+	debug.Log("Move %s to %s", from, to)
 	return r.move(ctx, from, to, true)
 }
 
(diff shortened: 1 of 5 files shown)


internal/store/root/move.go
```diff
@@ -156,6 +147,43 @@ func (r *Store) moveFromTo(ctx context.Context, subFrom *leaf.Store, from, to, f
...
+		dstIsDir = false
+		to = ""
+	}
+
+	// are we moving into an existing directory? Then we just need to prepend
+	// it's name to the source.
+	// a -> b
+	// - a/f1 -> b/a/f1
+	// a -> b
+	// - a -> b/a
```
AnomalRoil, line 162: Wow, I've no idea what that means.

`mv a b` on a tree with a/f1 inside of a?
And same but for just an entry named a?

internal/store/root/move.go
```diff
@@ -156,6 +147,43 @@ func (r *Store) moveFromTo(ctx context.Context, subFrom *leaf.Store, from, to, f
...
+	// are we moving a simple file? that's easy
+	if !srcIsDir {
+		// otherwise we just rename a file to another name
+		return to
+	}
+
+	// move a/ b, where a is a directory with a trailing slash and b
+	// does not exist, i.e. move a to b
+	if strings.HasSuffix(from, "/") {
+		return path.Join(to, strings.TrimPrefix(src, from))
```
AnomalRoil, line 179: I don't see any difference with the "else" path that returns the same thing, no?

Is this if clause necessary?

AnomalRoil (approved):

PR #2117, opened 2022-01-17
Change saltLen to unit8 and add some tests

### LINKED ISSUE #2116: saltLen should return uint8
Minor cleanup, but flagged by the security scanner.

[saltLen](https://github.com/gopasspw/gopass/blob/f2cac9f3b37b1731b7d3fff967bbdae010ffca9e/internal/tpl/funcs.go#L53) could just return an uint8. And avoid these ugly bare returns.

dominikschulz:
Fixes #2116

RELEASE_NOTES=n/a

### DIFF
--- a/internal/tpl/funcs.go
+++ b/internal/tpl/funcs.go
@@ -48,33 +48,30 @@ func sha1sum() func(...string) (string, error) {
 }
 
 // saltLen tries to parse the given string into a numeric salt length.
-// NOTE: This is on of the rare cases where I think named returns
-// are useful.
-func saltLen(s []string) (saltLen int) {
-	defer func() {
-		debug.Log("using saltLen %d", saltLen)
-	}()
-
+func saltLen(s []string) uint8 {
 	// default should be 32bit
-	saltLen = 32
+	sl := uint8(32)
+
+	defer func(sl uint8) {
+		debug.Log("using saltLen %d", sl)
+	}(sl)
 
 	if len(s) < 2 {
-		return
+		return sl
 	}
 
-	i, err := strconv.Atoi(s[0])
-	if err == nil && i > 0 {
-		saltLen = i
-	}
+	i, err := strconv.ParseUint(s[0], 10, 8)
 	if err != nil {
 		debug.Log("failed to parse saltLen %+v: %q", s, err)
+		return sl
 	}
-	return
+	sl = uint8(i)
+	return sl
 }
 
 func md5cryptFunc() func(...string) (string, error) {
 	return func(s ...string) (string, error) {
-		sl := uint8(saltLen(s))
+		sl := saltLen(s)
 		if sl > 8 || sl < 1 {
 			sl = 4
 		}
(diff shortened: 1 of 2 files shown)


internal/tpl/funcs.go
```diff
@@ -48,33 +48,30 @@ func sha1sum() func(...string) (string, error) {
...
 	if len(s) < 2 {
-		return
+		return sl
 	}
 
-	i, err := strconv.Atoi(s[0])
-	if err == nil && i > 0 {
-		saltLen = i
-	}
+	i, err := strconv.ParseUint(s[0], 10, 8)
```
AnomalRoil, line 63: Is this really supposed to be `s[0]` and not `s[1]` ? Looking below at the usage with `ssha.Generate`, it seems `s[0]` is meant to be the password already...

AnomalRoil (dismissed):
Are we sure we have the proper tests for `saltLen` as well?

internal/tpl/funcs.go
dominikschulz: I'm very confused right now. It looks like it's been wrong all the time (i.e. the salt being the first parameter). So the hashes shouldn't work. But I'm quite sure I did use them for real. Which can't be true.

Let me try to add proper tests and then come back to this question.

AnomalRoil:
It does seem like it was wrong already, but then isn't that really a breaking change that could change the resulting hashes for people?

dominikschulz:
Do you mean "breaking change" in the context of semver?
If so I'd argue no, because it was broken before and did not work as intended. So this is a fix, not a breaking change.

The previously generated hashes were just invalid. They would use a random salt of length 32 (depending on the function) and either an empty string or the requested salt length as the password.

I don't know why this did never occur to anyone. Maybe nobody is using this?

I certainly did use it before, but not sure when. If I have some time I might bisect to find the breaking change. But not right now.

AnomalRoil:
Yeah, I meant "breaking" as in "this could lead to different hashes now that we have patched".
But having now looked at its references in the code, it seems we are only using it to generate new values, so it's not going to cause different values on existing entries.

My guess is that it's dating from when we stopped using fixed size hashes to use this saltLen function instead last year in [that commit](https://github.com/gopasspw/gopass/commit/8d227c9e3858112cdafd4bf449fde454141107ae).

AnomalRoil (approved):

dominikschulz:
Wow, good find.

Yes, indeed. That must be it.

I don't think I have used the feature since then.

Leaves the question how useful it is.
But I need it from time to time ...

AnomalRoil:
@dominikschulz  Wow, seems this just broke our CI on master 😱

dominikschulz:
Why would that fail after the tests ran successfully on the PR? 🤔

AnomalRoil:
I just ran it twice locally and it failed once and worked once...

AnomalRoil:
```
$ go test ./internal/tpl/... -v   
    template_test.go:290: 
        	Error Trace:	template_test.go:290
        	Error:      	Received unexpected error:
        	            	can't validate: hash does not match password
        	Test:       	TestVars/{{.Content_|_md5crypt_"7"}}
        	Messages:   	{{.Content | md5crypt "7"}}
--- FAIL: TestVars (0.00s)
```

dominikschulz:
So it's a flaky test. Hmm, I didn't expect that. Need to take a look later. https://github.com/gopasspw/gopass/issues/2121

PR #2119, opened 2022-01-17
Always re-encrypt when fsck is invoked with --decrypt

### LINKED ISSUE #2015: Add a `--reencrypt` option to `fsck`
It can happen that we are doing a "key rotation" or have changed some things related to PGP that require us to fully re-encrypt our store.

Currently the only process to do so it adding a dummy recipient and then removing it.
This is not really desirable as it can lead to side effects such as committing secrets with a "weaker" key or so to our pw store.

Ideally we should be able to use `fsck --reencrypt` in order to decrypt all secrets and "re-encrypt" it with the current set of recipients.
 
This would also help solving issues such as "My process failed while I was adding a recipient", "I ^C Gopass in the middle of the removal of a recipient" or other things like that and would fit nicely under `fsck` responsabilites, IMO.

This would also have helped for example in #1977 when it was needed to fully re-encrypt the store using a downgraded PGP version because the latest PGP version wasn't being properly supported and cross-compatible for our usecase.

This also help with "expired subkeys": I can just add a new subkey to my key and "re-encrypt" the whole store. 
So it would relate to: #1969, #1107, #1430

What it should do: re-encrypt in place all secrets for the current set of recipients, using the "main" encryption subkey of a key (or the specified subkeys) for each recipient, and then commit/push/sync the whole as per existing config.

dominikschulz:
Fixes #2015

RELEASE_NOTES=[BUGFIX] Always re-encrypt when fsck is invoked with
--decrypt.

### DIFF
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -87,8 +87,15 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) error {
 	if err != nil {
 		return fmt.Errorf("failed to decode secret %s: %w", name, err)
 	}
+
+	// check if this is still an old MIME secret.
+	// Note: the secret was already converted when it was parsed during Get.
+	// This is just checking if it was converted from MIME or not.
+	// This branch is pretty much useless right now, but I'd like to add some
+	// reporting on how many secrets were converted from MIME to new format.
+	// TODO: report these stats
 	if cs, ok := sec.(convertedSecret); ok && cs.FromMime() {
-		debug.Log("leftover Mime secret: %s\nYou should consider editing it to re-encrypt it.", name)
+		debug.Log("leftover Mime secret: %s", name)
 	}
 
 	out.Printf(ctx, "Re-encrypting %s to fix recipients and storage format.", name)
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -71,20 +71,35 @@ type convertedSecret interface {
 }
 
 func (s *Store) fsckCheckEntry(ctx context.Context, name string) error {
-	// make sure we can actually decode this secret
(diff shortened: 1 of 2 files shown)


internal/store/leaf/fsck.go
```diff
@@ -71,20 +71,35 @@ type convertedSecret interface {
...
+	}
+
+	// we need to make sure Parsing is enabled in order to parse old Mime secrets
+	ctx = ctxutil.WithShowParsing(ctx, true)
+	sec, err := s.Get(ctx, name)
+	if err != nil {
+		return fmt.Errorf("failed to decode secret %s: %w", name, err)
+	}
+	if cs, ok := sec.(convertedSecret); ok && cs.FromMime() {
+		debug.Log("leftover Mime secret: %s\nYou should consider editing it to re-encrypt it.", name)
```
AnomalRoil, line 91: Aren't we now re-encrypting without Mime just below?

AnomalRoil (changes requested):

internal/store/leaf/fsck.go
dominikschulz: No, but this might deserve a comment.

We're only checking if it was converted.
The type assertion doesn't change it.

The question is more of need the check at all.

AnomalRoil: I meant the new code on line 102, is the Set actually encrypting that back as a Mime secret?

dominikschulz: Set is using `sec.Bytes()`. And secparse (inside Get) never returns a MIME secret, there is not MIME secret type anymore. There is a parser, but that's returning a KV secret. So this should work as intended (i.e. write as a KV secret).

AnomalRoil (approved):

PR #2122, opened 2022-01-18
Disable flaky tests and add an example

dominikschulz:
Fixes# 2121

RELEASE_NOTES=n/a

Signed-off-by: Dominik Schulz <dominik.schulz@gauner.org>

PR #2128, opened 2022-01-24
Prevent MD5Crypt to generate invalid salts

### LINKED ISSUE #2121: Fix flaky template tests
https://github.com/gopasspw/gopass/pull/2117#issuecomment-1015261881

AnomalRoil:
This fixes #2121 without disabling it for now, unlike #2122.

I'm using rejection sampling, and my tests show that the error has a ~2-3% chance of being triggered, so the rejection sampling seems fine and isn't at risk of falling into an infinite loop.

### DIFF
--- a/internal/tpl/funcs.go
+++ b/internal/tpl/funcs.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha1"
 	"fmt"
 	"strconv"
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
LGTM. Thank you.

PR #2135, opened 2022-02-09
extend setup doc to freebsd and csh/tcsh

r0lh:
Extends the setup documentation for FreeBSD and adopts the "Inappropriate ioctl"-error-workaround for csh/tcsh


dominikschulz (approved):
LGTM. 👍

PR #2138, opened 2022-02-19
docs: mention tessen in the list of integrations

ayushnix:
tessen is a bash script that can autotype and copy data from
passwordstore and gopass on wayland compositors on Linux

RELEASE_NOTES=n/a

Signed-off-by: Ayush Agarwal <ayushnix@fastmail.com>


dominikschulz (approved):
LGTM

PR #2145, opened 2022-02-26
Minor generate and show fixes

### LINKED ISSUE #2140: Make it more obvious when we apply a rule on password generation
> Should we make it more obvious if we apply a canned rule?

_Originally posted by @dominikschulz in https://github.com/gopasspw/gopass/discussions/2123#discussioncomment-2235877_

The best way would probably to add an `info` log saying "Matched domain name with password rule [{add the rule here}].

We might also want to consider adding a flag/env var to bypass rule checking, and test how it behaves when generating a key-value instead of the main password.

dominikschulz:
Fixes #2140

Addresses a few minor issues found when trying to reproduce #2140.

RELEASE_NOTES=n/a

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -111,7 +111,7 @@ func keyAndLength(args argList) (string, string) {
 func (s *Action) generateCopyOrPrint(ctx context.Context, c *cli.Context, name, key, password string) error {
 	entry := name
 	if key != "" {
-		entry += ":" + key
+		entry += " " + key
 	}
 
 	out.OKf(ctx, "Password for entry %q generated", entry)
@@ -161,7 +161,7 @@ func hasPwRuleForSecret(name string) (string, pwrules.Rule) {
 
 // generatePassword will run through the password generation steps.
 func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, name string) (string, error) {
-	if domain, rule := hasPwRuleForSecret(name); domain != "" {
+	if domain, rule := hasPwRuleForSecret(name); domain != "" && !c.Bool("force") {
 		return s.generatePasswordForRule(ctx, c, length, name, domain, rule)
 	}
 
@@ -220,7 +220,7 @@ func clamp(min, max, value int) int {
 }
 
 func (s *Action) generatePasswordForRule(ctx context.Context, c *cli.Context, length, name, domain string, rule pwrules.Rule) (string, error) {
-	out.Printf(ctx, "Using password rules for %s ...", domain)
+	out.Noticef(ctx, "Using password rules for %s ...", domain)
 	wl := 16
 	if iv, err := strconv.Atoi(length); err == nil {
 		wl = clamp(rule.Minlen, rule.Maxlen, iv)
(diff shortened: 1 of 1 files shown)


internal/action/generate.go
```diff
@@ -111,7 +111,7 @@ func keyAndLength(args argList) (string, string) {
 func (s *Action) generateCopyOrPrint(ctx context.Context, c *cli.Context, name, key, password string) error {
 	entry := name
 	if key != "" {
-		entry += ":" + key
+		entry += " " + key
```
AnomalRoil, line 114: This should have broken some test 😢
Why not keep the entry:key format?

internal/action/generate.go
```diff
@@ -161,7 +161,7 @@ func hasPwRuleForSecret(name string) (string, pwrules.Rule) {
 
 // generatePassword will run through the password generation steps.
 func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, name string) (string, error) {
-	if domain, rule := hasPwRuleForSecret(name); domain != "" {
+	if domain, rule := hasPwRuleForSecret(name); domain != "" && !c.Bool("force") {
```
AnomalRoil, line 164: Well, I'm not sure that's really the behavior we want for `force` in generate, is it?

As I understand it, `force` is supposed to overwrite the existing entry even if it already exists, no?
So, if for example I need to generate a new password for a website because it was recently compromised, I might want to use the force flag, but I would still like it to respect the password rules for that website, no?

internal/action/generate.go
```diff
@@ -220,7 +220,7 @@ func clamp(min, max, value int) int {
 }
 
 func (s *Action) generatePasswordForRule(ctx context.Context, c *cli.Context, length, name, domain string, rule pwrules.Rule) (string, error) {
-	out.Printf(ctx, "Using password rules for %s ...", domain)
+	out.Noticef(ctx, "Using password rules for %s ...", domain)
```
AnomalRoil, line 223: Maybe rephrase that to make it more obvious it's due to the domain name?
Something like :
"The domain %s appears to have specific password rules, using them."

internal/action/generate.go
```diff
@@ -231,6 +231,7 @@ func (s *Action) generatePasswordForRule(ctx context.Context, c *cli.Context, le
 	if err != nil {
```
AnomalRoil, line 231: Shouldn't we reject invalid lengths instead of just clamping them below?

internal/action/generate.go
dominikschulz: Because it does not work. IIRC we did never support that format.

internal/action/generate.go
dominikschulz: Yes. We could have another flag for that. More flags, more documentation, more maintenance but also more control.
I don't have a strong preference.

internal/action/generate.go
dominikschulz: Maybe. But either we make the implementation much more convoluted (loops etc.) or we abort the execution and possibly disrupt the user. Not sure either. But not clamping or checking them also sounds wrong.

internal/action/generate.go
AnomalRoil: I just double checked and the force flag isn't really needed to overwrite an existing password: we get a prompt instead:
https://github.com/gopasspw/gopass/blob/55ccbbb197e4c9dae2f8f8de6b8a8586353e014c/internal/action/generate.go#L60

AnomalRoil (approved):

PR #2154, opened 2022-03-12
Improve gopass age identities

### LINKED ISSUE #2099: Add `gopass identities` to manipulate identities (private keys)
`gopass` has some automatic helper to create new private keys when needed, but exposes no way to inspect or maniuplate them.

For `gpg` this might be reasonable, but for `age` this doesn't work well. And when mixing different encryption backends it becomes confusing. Some kind of helpers would be good.

dominikschulz:
RELEASE_NOTES=n/a

Fixes #2099

### DIFF
--- a/internal/backend/crypto/age/identities.go
+++ b/internal/backend/crypto/age/identities.go
@@ -3,6 +3,7 @@ package age
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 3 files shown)


internal/backend/crypto/age/identities.go
```diff
@@ -34,8 +35,9 @@ func (a *Age) Identities(ctx context.Context) ([]age.Identity, error) {
 	if err != nil {
 		debug.Log("failed to decrypt existing identities from %s: %s", a.identity, err)
 		if !os.IsNotExist(err) {
-			return nil, fmt.Errorf("failed to decrypt %s: %s", a.identity, err)
+			return nil, fmt.Errorf("failed to decrypt %s: %w", a.identity, err)
```
AnomalRoil, line 38: TIL `%w`, thanks.

AnomalRoil (approved):
Not sure why the tests are failing now. Lgtm overall

PR #2155, opened 2022-03-12
Add --chars option to show to allow printing a subset of the secret only

### LINKED ISSUE #2068: [Feature request] Add ability to extract specific characters of secrets
### Summary

Many websites persist in requesting specific characters of passwords/memorable phrases to be entered

image

image

It would be nice to be able to invoke a sub-command to just retrieve a set of characters based on their index from a password.

I'm sure you could do this by e.g. using a bash string substitution or e.g. `head`/`tail` but something like `gopass show bank/memorable :1,5,7` would be super convenient to get the necessary characters out in one go.

dominikschulz:
Fixes #2068

RELEASE_NOTES=Add --chars option to print subset of secrets


AnomalRoil (approved):
Lgtm, but maybe we should add it to the documentation for show too?

PR #2158, opened 2022-03-12
Enable more lint checks

dominikschulz:
Be more strict and bring back codequality checks for the repo.

Depends on https://github.com/golangci/golangci-lint/pull/2438 being fixed and released first.

RELEASE_NOTES=n/a


AnomalRoil (dismissed):
LGTM

PR #2159, opened 2022-03-15
Automatically export creators key to the store

### LINKED ISSUE #1919: Automatically add creators gpg key to recipients
### Summary
I just created a mount/repo, added another recipient, synced, and had the other person run `gopass recipients`.
It says that my gpg key is missing and they can not create secrets without it.

Manually running `gopass recipients add --store foo my_gpg` worked around this issue, adding my key as a recipient.

### Expected behavior
My gpg key should be automatically included in the recipients upon initialization of the store.

### Environment

- OS: Arch and OSX
- gopass Version: gopass 1.12.5 (2021-03-29 17:27:06) go1.16.2 linux amd64
- Installation method: from the respective package managers

dominikschulz:
Fixes #1919

This will also automatically try to import a public key from the store if none is found. This makes it very convenient to add new recipients, but on the other hand this is might open a small attack vector if someone manages to inject their public key into the git repo and getting an existing member to add it to the recipients. But then on the other hand I wonder how carefully people examine imported GPG keys anyway.

RELEASE_NOTES=[ENHANCEMENT] Automatically export creators key to the
store. Change default for AutoImport to false.

### DIFF
--- a/internal/action/clone.go
+++ b/internal/action/clone.go
@@ -120,7 +120,7 @@ func (s *Action) clone(ctx context.Context, repo, mount, path string) error {
 
 	// make sure the parent directory exists.
 	if parentPath := filepath.Dir(path); !fsutil.IsDir(parentPath) {
-		if err := os.MkdirAll(parentPath, 0700); err != nil {
+		if err := os.MkdirAll(parentPath, 0o700); err != nil {
 			return exit.Error(exit.Unknown, err, "Failed to create parent directory for clone: %s", err)
 		}
 	}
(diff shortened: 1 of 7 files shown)


internal/action/clone.go
```diff
@@ -120,7 +120,7 @@ func (s *Action) clone(ctx context.Context, repo, mount, path string) error {
 
 	// make sure the parent directory exists.
 	if parentPath := filepath.Dir(path); !fsutil.IsDir(parentPath) {
-		if err := os.MkdirAll(parentPath, 0700); err != nil {
+		if err := os.MkdirAll(parentPath, 0o700); err != nil {
```
AnomalRoil, line 123: I'd argue that `0700` is already octal and the way we are used to see permission in the unix world.

internal/config/config.go
```diff
@@ -37,7 +37,7 @@ type Config struct {
 // New creates a new config with sane default values.
 func New() *Config {
 	return &Config{
-		AutoImport:    true,
+		AutoImport:    false,
```
AnomalRoil, line 40: This might warrant a warning in the RELEASE_NOTES, no?

AnomalRoil (approved):

internal/action/clone.go
```diff
@@ -200,7 +200,21 @@ func (s *Action) cloneCheckDecryptionKeys(ctx context.Context, mount string) err
...
+			debug.Log("failed to export missing public keys: %w", err)
+		}
+	} else {
+		debug.Log("failed to get sub store: %s", err)
+	}
+
 	out.Noticef(ctx, "Please ask the owner of the password store to add one of your keys: %s", strings.Join(idSet.Elements(), ", "))
+	if exported {
+		out.Noticef(ctx, "The missing keys were exported to the password store. Run `gopass sync` to push them.")
+	}
```
AnomalRoil, line 217: That's a nice addition, especially for usage with age.

internal/action/clone.go
dominikschulz: This is a [gofumpt](https://github.com/mvdan/gofumpt) change. I don't have a strong opinion on this, but since we don't have strong community of "style reviewers" here I'd prefer to rely on automation as much as possible, i.e. follow the conventions that the tools suggest.

internal/config/config.go
dominikschulz: Done

PR #2165, opened 2022-03-16
Small formatting fix

dominikschulz:
RELEASE_NOTES=n/a

### DIFF
--- a/internal/action/sync.go
+++ b/internal/action/sync.go
@@ -100,7 +100,7 @@ func (s *Action) syncMount(ctx context.Context, mp string) error {
 		debug.Log("Push succeeded")
 		out.Printf(ctxno, color.GreenString("OK"))
 	case errors.Is(err, store.ErrGitNoRemote):
-		out.Printf(ctxno, "Skipped (no remote)")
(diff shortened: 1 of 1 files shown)


internal/action/sync.go
```diff
@@ -100,15 +100,15 @@ func (s *Action) syncMount(ctx context.Context, mp string) error {
 		debug.Log("Push succeeded")
 		out.Printf(ctxno, color.GreenString("OK"))
 	case errors.Is(err, store.ErrGitNoRemote):
-		out.Printf(ctxno, "Skipped (no remote)")
+		out.Printf(ctx, "Skipped (no remote)")
```
AnomalRoil, line 103: Shouldn't this be printed on stderr instead?
Not sure tho, because I guess we might want it to break people's scripts in case they forget to add a remote?

AnomalRoil (approved):

internal/action/sync.go
dominikschulz: `gopass sync` is an interactive command. Skipping because there is no remote is not an error, just a notice.
I have stores without any remotes. Perfectly legit.

PR #2166, opened 2022-03-16
Do not display (missing public key) for age identities

dominikschulz:
They are fully defined by the public key.

RELEASE_NOTES=[BUGFIX] Do not print missing public key for age.

### DIFF
--- a/internal/backend/crypto/gpg/cli/keyring.go
+++ b/internal/backend/crypto/gpg/cli/keyring.go
@@ -27,7 +27,7 @@ func (g *GPG) listKeys(ctx context.Context, typ string, search ...string) (gpg.K
 		}
 	}
 	cmd := exec.CommandContext(ctx, g.binary, args...)
-	var errBuf = bytes.Buffer{}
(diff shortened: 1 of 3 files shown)


internal/backend/crypto/gpg/cli/keyring.go
```diff
@@ -55,16 +59,26 @@ func (g *GPG) Fingerprint(ctx context.Context, id string) string {
...
 	}
 
 	tmpl, err := template.New(tpl).Parse(tpl)
 	if err != nil {
 		return ""
 	}
 
+	var gid gpg.Identity
+	k, found := g.findKey(ctx, id)
+	if found {
```
AnomalRoil, line 76: Else what?

AnomalRoil (approved):

internal/backend/crypto/gpg/cli/keyring.go
dominikschulz: gid will be left at the zero value of `gpg.Identity`.

PR #2176, opened 2022-03-25
Trying to patch our CI

AnomalRoil:
I suspect Go 1.18 is not the default for GHA runners, so we need to specify it.


AnomalRoil:
superseded by #2177

PR #2183, opened 2022-04-08
Do not decrypt secrets when moving or copying them

### LINKED ISSUE #2181: `gopass mv` decrypts the entry
### Summary
When running `gopass mv a/path/to/entry b/new/path/name`, Gopass wants to decrypt the entry, which doesn't really make sense since it doesn't need to know its content to move it around on the filesystem. 

### Steps To Reproduce
Run gopass without the private key for that store. 
(I do it by unplugging my Yubikey, but there are other ways to achieve it by setting `GNUPGHOME=some/path` e.g.)

```
$ gopass generate test/to/move 20 
✅ Password for entry "test/to/move" generated

$ gopass mv test/to/move test/new/path
gpg: decryption failed: No secret key
❌ Decryption failed: exit status 2

Error: source test/to/move does not exist in source store : failed to decrypt
```

Plus notice the error message is quite cryptic and factually wrong since it says "source test/to/move does not exist in source store".

### Expected behavior

This should work without decrypting the entry first.

### Environment

- OS version: based on Ubuntu 21.10, kernel  `5.16.15-76051615-generic x86_64 GNU/Linux`
- gopass Version: `gopass 1.14.0-git+HEAD`
- Installation method: `go install`

dominikschulz:
This PR attempts to avoid decrypting secrets needlessly when they are
only copied or moved. All currently supported backends have a filesystem
backing them so the old workaround of Get and Set shouldn't be necessary
anymore. In case we re-introduce such backends or a direct move fails
for some reason we still fall back to Get and Set if necessary.

Fixes #2181

RELEASE_NOTES=[ENHANCEMENT] Avoid decryption on move or copy

### DIFF
--- a/internal/store/leaf/move.go
+++ b/internal/store/leaf/move.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/gopasspw/gopass/internal/queue"
 	"github.com/gopasspw/gopass/internal/store"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
@@ -19,6 +20,11 @@ func (s *Store) Copy(ctx context.Context, from, to string) error {
 	// recursive copy?
 	if s.IsDir(ctx, from) {
 		return fmt.Errorf("recursive operations are not supported")
+	}
+
+	// try direct copy first
+	if err := s.directMove(ctx, from, to, false); err == nil {
+		return nil
 	}
 
 	content, err := s.Get(ctx, from)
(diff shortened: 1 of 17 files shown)


internal/store/leaf/move.go
```diff
@@ -21,6 +22,11 @@ func (s *Store) Copy(ctx context.Context, from, to string) error {
 		return fmt.Errorf("recursive operations are not supported")
 	}
 
+	// try direct copy first
+	if err := s.directMove(ctx, from, to, false); err == nil {
+		return nil
+	}
+
```
AnomalRoil, line 29: This might benefit from a debug log like below on line 58, no?

AnomalRoil (approved):

PR #2189, opened 2022-04-12
Set permissions for GitHub actions

nathannaveen:
- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

 Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

Signed-off-by: nathannaveen <42319948+nathannaveen@users.noreply.github.com>


AnomalRoil:
LGTM, could you add a `RELEASE_NOTES=n/a ` to your commit message for our changelog system?

AnomalRoil (approved):

.github/workflows/autorelease.yml
```diff
@@ -9,6 +9,9 @@ on:
     tags:
       - 'v*'
           
+permissions:
+  contents: read
```
dominikschulz, line 13: Is that really sufficient? I think our release process also needs to write. But I'm not 100% sure.

dominikschulz:
Let's try it. If it fails we can undo the change easily.

dominikschulz (approved):
LGTM

PR #2191, opened 2022-04-16
Add autosync once a week

dominikschulz:
RELEASE_NOTES=[ENHANCEMENT] Automatically sync once a week


internal/action/sync.go
```diff
@@ -23,6 +25,30 @@ func (s *Action) Sync(c *cli.Context) error {
...
+func (s *Action) autoSync(ctx context.Context) error {
+	if !ctxutil.IsInteractive(ctx) {
+		return nil
+	}
+
+	if !ctxutil.IsTerminal(ctx) {
+		return nil
+	}
+
+	if sv := os.Getenv("GOPASS_NO_AUTOSYNC"); sv != "" {
```
AnomalRoil, line 37: We might want to document this new variable in https://github.com/gopasspw/gopass/blob/master/docs/config.md

AnomalRoil (approved):
This seems nice. It's adding a minimal amount of code to achieve it, so I'm in favor of adding it.
I didn't check why the tests are failing, tho.

Why 1 week and not 2 or 3 days instead? We might want to make it tweakable through the env variable, no?

internal/action/sync.go
dominikschulz: Done.

dominikschulz:
I don't like to keep adding config options.
At least not until we can revamp the config handling.

We need some way to override some settings per store.
But the config also needs to be backward compatible and work with partial configs.

I really like the git config format more and more.

PR #2197, opened 2022-04-25
Add Fedora installation instructions

Fale:
gopass is now avaialbe in official Fedora repositories.

See [blog post](https://fale.io/blog/2022/04/25/gopass-in-fedora), [src.fpo](https://src.fedoraproject.org/rpms/gopass/)

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -198,7 +198,13 @@ layman -a go-overlay
 emerge -av gopass
 ```
 
-### Fedora / Red Hat / CentOS
+### Fedora
+
+```bash
+dnf install gopass
+```
+
+### Red Hat / CentOS
 
 There is an unofficial RPM build maintained by a contributor.
 


dominikschulz:
Wow, that's amazing. Thanks a lot. Also for the blog post!

AnomalRoil:
Great, thanks.

It's true we never thought about how moving to Go 1.18 could be hurting mainstream adoption by official repositories that do not package Go in its latest version.

Fale:
Happy to help :-).

On the 1.18 topic, yes, some older versions could have the issue, but soon major distros will have 1.18+, I hope.
In a couple of weeks the Fedora stable version will be F36, which has go 1.18. F35 will go EOL in 7 months, after which all Fedora supported releases will be ok.
Other distros (I think for instance Debian) might have a longer lag, though!

PR #2206, opened 2022-05-01
Upload released deb packages to our apt host

dominikschulz:
I'm working on a new approach to hosting packages for users of deb based distributions.
It's not ready for public consumption, yet. But I do want to test if it works with our release pipeline at the next release.

RELEASE_NOTES=n/a

### DIFF
--- a/.github/workflows/autorelease.yml
+++ b/.github/workflows/autorelease.yml
@@ -74,3 +74,12 @@ jobs:
           echo gh release upload "${tag}" "${msi}"
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+      -
+        name: "Upload deb files to apt hosting"
+        run: |
+          for D in dist/*.deb; do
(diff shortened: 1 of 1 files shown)


.github/workflows/autorelease.yml
```diff
@@ -74,3 +74,12 @@ jobs:
           echo gh release upload "${tag}" "${msi}"
         env:
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
+      -
+        name: "Upload deb files to apt hosting"
+        run: |
+          for D in dist/*.deb; do
+            curl -H"X-Filename: ${D}" -H'X-Apikey: ${APIKEY}' -XPOST --data-binary @$D http://packages.gauner.org/repos/gopass/upload
```
AnomalRoil, line 81: The only thing I would complain about is that it's going over a plain HTTP line instead of HTTPS.

AnomalRoil (approved):

.github/workflows/autorelease.yml
dominikschulz: Yes. But this will take some time to resolve. That's why I said they it's not ready for public consumption.

But for a test it should be alright.

Once I have the infrastructure in place I'll wipe the repo and regenerate the signing key.

PR #2219, opened 2022-05-11
Determine password length from env variable, if set

### LINKED ISSUE #2190: Feature request: configure default length
Since [the `ji32k7au4a83` incident](https://duckduckgo.com/?q=ji32k7au4a83) I tend o use passwords with an odd number of characters, because I assume this can not happen then. gopass has a default length value set to `24` though:

    mcn@tx ~ gopass generate -c thismount/mysite/mylogin
    How long should the password be? [24]: 25

It would be cool to be able to set the default value to something else. Since I never found a setting therefore in the docs, I herewith kindly request this feature :)

PS: this is a copy of #1198 which sadly got auto-closed without any feedback.

dotcs:
With this change, the variable `GOPASS_PW_DEFAULT_LENGTH` can be used to set a default length for the `generate` cmd.

This PR fixes #2190.

### DIFF
--- a/internal/action/generate_test.go
+++ b/internal/action/generate_test.go
@@ -178,6 +178,35 @@ func TestGenerate(t *testing.T) { //nolint:paralleltest
 		assert.Contains(t, buf.String(), "Copied to clipboard")
 		buf.Reset()
 	})
+
+	// generate --force foobar w/ pw length set via env variable (42 chars)
+	t.Run("generate --force foobar", func(t *testing.T) { //nolint:paralleltest
+		t.Setenv("GOPASS_PW_DEFAULT_LENGTH", "42")
+
+		if testing.Short() {
+			t.Skip("skipping test in short mode.")
+		}
+
+		assert.NoError(t, act.Generate(gptest.CliCtxWithFlags(ctx, t, map[string]string{"force": "true", "print": "true", "symbols": "false"}, "foobar")))
+		lines := strings.Split(strings.TrimSpace(buf.String()), "\n")
+		assert.Len(t, lines[3], 42)
+		buf.Reset()
+	})
+
+	// generate --force foobar w/ pw length set via env variable to invalid value, fallback mechanism
+	t.Run("generate --force foobar", func(t *testing.T) { //nolint:paralleltest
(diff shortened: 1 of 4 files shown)


internal/action/generate_test.go
```diff
@@ -178,6 +178,35 @@ func TestGenerate(t *testing.T) { //nolint:paralleltest
 		assert.Contains(t, buf.String(), "Copied to clipboard")
 		buf.Reset()
 	})
+
+	// generate --force foobar w/ pw length set via env variable (42 chars)
+	t.Run("generate --force foobar", func(t *testing.T) { //nolint:paralleltest
+		t.Setenv("GOPASS_PW_DEFAULT_LENGTH", "42")
+
+		if testing.Short() {
```
dominikschulz, line 186: I don't think you need to skip this test in short mode.

dominikschulz (dismissed):
LGTM. Thanks a lot!

internal/action/generate_test.go
dotcs: Thanks for the hint. I've removed the if-condition. 
I'm not exactly sure when to skip tests in short mode and when not. I was just copying the lines from another test. Sorry, should have mentioned that. Is it okay to leave it for the other test in line 193-204, or should I remove it there as well.

dotcs:
Thanks @dominikschulz for the review! :+1:
I have incorporated some feedback that I received in the linked issue, fixed a few lint bugs and added some more tests. I guess you need to take another look. Thanks in advance.

AnomalRoil (approved):
Awesome, that looks good.

dotcs:
Thanks for the review, @AnomalRoil. I've fixed the remaining lint issue in my latest commit. There are a couple of other lint issues which have nothing to do with my changes in this PR. `make codequality` also fails for me on the `master` branch. Is there an easy way to locally test this and only get those lint errors that are because of the code changes in the PR? And is there any chance that existing lint errors will be fixed in `master` to have a clean result? That would imo make writing PRs much simpler. I hope that this time everything is working out. 🤞

dominikschulz (approved):
LGTM. Thanks a lot!

PR #2226, opened 2022-05-19
Add flag '--keep-case' to env subcommend

### LINKED ISSUE #2225: env subcommand converts entry name to all upper case when setting them as environment variables
### Summary

Env subcommand converts the entry name to all upper case and sets them as environment variable. This doesn't work with CLI tools that are case-sensitive.

Terraform for example lets you pass values for variables as environment variables and requires them to be prefixed with `TF_VAR_` but rest of the variable name needs to be lower case, i.e. TF_VAR_example_var. More info https://www.terraform.io/cli/config/environment-variables#tf_var_name. As the subcommand sets them all upper case, Terraform doesn't read it as it is case sensitive and instead prompts the user to enter a value for the variable.

### Steps To Reproduce

1. Create a new entry in gopass by running 
```shell
echo "text-goes-here" | gopass insert test/env/TF_VAR_example_var
```
2. Run the following command to view the entry you just created set as environment variable. 
```shell
gopass env test/env printenv | grep TF_VAR
```

### Expected behavior

env subcommand retains the original case when setting the entry name as environment variable. Something like below

```shell
❯ gopass env test/env printenv | grep TF_VAR
TF_VAR_example_var=text-goes-here
```

### Environment

- OS: macOS Monterey
- OS version: Darwin MBP 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64
- gopass Version: 1.14.1
- Installation method: brew

### Additional context

Using gopass with age crypto & git backend if that makes a difference. 

Maybe this could be implemented as a flag to env subcommand to keep the case as it is from entry name and leave the current behaviour, so it doesn't break other people's work flows.

dotcs:
By adding this boolean flag to the `gopass env` cmd, the original capitalization of the secret is maintained. This fixes #2225.

RELEASE_NOTES=[ENHANCEMENT] Add flag to keep env variable capitalization in `gopass env` subcmd

Signed-off-by: dotcs <git@dotcs.me>

### DIFF
--- a/internal/action/commands.go
+++ b/internal/action/commands.go
@@ -300,6 +300,13 @@ func (s *Action) GetCommands() []*cli.Command {
 			Action:       s.Env,
 			BashComplete: s.Complete,
 			Hidden:       true,
+			Flags: []cli.Flag{
+				&cli.BoolFlag{
+					Name:  "keep-capitalization",
+					Value: false,
+					Usage: "Do not capitalize the environment variable and instead retain the original capitalization",
+				},
+			},
 		},
 		{
 			Name:      "find",
--- a/internal/action/commands.go
+++ b/internal/action/commands.go
@@ -302,9 +302,10 @@ func (s *Action) GetCommands() []*cli.Command {
 			Hidden:       true,
 			Flags: []cli.Flag{
 				&cli.BoolFlag{
-					Name:  "keep-capitalization",
-					Value: false,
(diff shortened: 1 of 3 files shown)


dominikschulz (approved):
LGTM. Thanks a lot.

PR #2239, opened 2022-05-27
Allow removing multiple secrets at one

dominikschulz:
RELEASE_NOTES=[ENHANCEMENT] Allow deleting multiple secrets

### DIFF
--- a/internal/action/delete.go
+++ b/internal/action/delete.go
@@ -21,35 +21,44 @@ func (s *Action) Delete(c *cli.Context) error {
 		return exit.Error(exit.Usage, nil, "Usage: %s rm name", s.Name)
 	}
 
-	if !recursive && s.Store.IsDir(ctx, name) && !s.Store.Exists(ctx, name) {
+	if recursive {
+		if len(c.Args().Tail()) > 1 {
+			return exit.Error(exit.Usage, nil, "Deleting multiple keys is not supported in recursive mode")
+		}
+
+		return s.deleteRecursive(ctx, name, c.Bool("force"))
+	}
+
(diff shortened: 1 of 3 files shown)


internal/action/delete.go
```diff
@@ -21,48 +21,75 @@ func (s *Action) Delete(c *cli.Context) error {
...
-		return exit.Error(exit.Usage, nil, "Can not use -r with a key. Invoke delete either with a key or with -r")
+	// multiple secrets, so not a key
+	if len(c.Args().Tail()) > 1 {
+		key = ""
+	}
+
+	names := append([]string{name}, c.Args().Tail()...)
+
+	if key != "" && s.Store.Exists(ctx, key) {
+		return exit.Error(exit.Unsupported, nil, "Ambiguous key %q, use 'gopass edit %s' to delete", key, name)
```
AnomalRoil, line 47: This could be more explicit, saying that the key name clashes with an entry name.

internal/action/delete.go
```diff
@@ -21,48 +21,75 @@ func (s *Action) Delete(c *cli.Context) error {
...
+	}
+
+	if !s.Store.Exists(ctx, name) {
```
AnomalRoil, line 50: Shouldn't this be checking each item in `names` instead of just `name`?

AnomalRoil (approved):

internal/action/delete.go
dominikschulz: Done.

internal/action/delete.go
dominikschulz: This will clash with the key deletion logic below. But the delete call will check that again for every secret.

PR #2248, opened 2022-05-31
Use IsTerminal instead of OutputIsRedirected

### LINKED ISSUE #2025: Consider replacing `out.OutputIsRedirected`  by `ctxutil.IsTerminal(ctx)`
It seems we might have redundant handling of the "Are we in a terminal?" and "is output being redirected?"

`ctxutil.IsTerminal` relies on https://github.com/mattn/go-isatty in order to tell if we are in a terminal or not, which is great because it is very cross-platform, even supporting Plan9 and Windows.

I recently added in #2019 a method called `OutputIsRedirected` in our `out` package, which is meant to say if we are being redirected or not:
https://github.com/gopasspw/gopass/blob/b490c57e97689c4bb97933031b0369afa1863928/internal/out/print.go#L30-L34

It is not clear to me if this is equivalent to saying whether or not we are in a terminal, but it's pretty clear it probably isn't very cross-platform as is.

If "not being a tty" is including "being redirected", then we might want to change this new function to rely on go-isatty/`ctxutil.IsTerminal(ctx)` instead...

_Originally posted by in https://github.com/gopasspw/gopass/pull/2024#discussion_r745122745_

AnomalRoil:
Deprecate `out.OutputIsRedirected` in favour of `!ctxutil.IsTerminal`.

Fixes #2025.

### DIFF
--- a/internal/action/otp.go
+++ b/internal/action/otp.go
@@ -95,7 +95,7 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	skip := ctxutil.IsHidden(ctx) || pw || qrf != "" || !ctxutil.IsTerminal() || !ctxutil.IsInteractive(ctx) || clip
+	skip := ctxutil.IsHidden(ctx) || pw || qrf != "" || !ctxutil.IsTerminal(ctx) || !ctxutil.IsInteractive(ctx) || clip
 	if !skip {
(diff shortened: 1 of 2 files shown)


dominikschulz (changes requested):
The GHA comments are correct. You need to pass in ctx.

AnomalRoil:
Yup, sorry for that, I had Go 1.17.9 installed locally with golangci-lint 1.45 so couldn't run it locally, and then really wanted to go to bed yesterday evening 😅

dominikschulz (approved):
LGTM

PR #2251, opened 2022-05-31
Fixing Windows releaser

AnomalRoil:
This should correct our issue:
```
✅ Current version is: 1.14.3
✅ Wrote Wix XML config to: /tmp/gopass-880485264/wix.xml
✅ Created MSI package at: dist/gopass-x64-windows-1.14.3.msi

stat dist/gopass-1.14.3-windows-x64.msi: no such file or directory
```

So we need to change the path to the correct `gopass-x64-windows-1.14.3.msi` instead.


dominikschulz (approved):
LGTM

PR #2278, opened 2022-07-18
Use github.com/pquerna/otp to allow using the key period

### LINKED ISSUE #2276: TOTP `period` parameter is ignored
### Summary

I configure a TOTP token with a custom `period` (other than 30).

Unfortunately, gopass currently uses a hardcoded default value of 30:
https://github.com/gopasspw/gopass/blob/master/internal/action/otp.go#L23

### Steps To Reproduce

Add a secret with a TOTP and `period=60` like the following:

```
% gopass show testForGopassIssue 
Secret: testForGopassIssue

someSecretPassword
otpauth: //totp/demoTest?secret=bcde23456&algorithm=SHA256&digits=6&period=60
```
Then retrieve the current token:
```
% gopass otp testForGopassIssue
245371
This OTP password still lasts for:
]  3 / 29 [Goooooopass                                                                                              ]  10.34%
```

with a period of 30, instead of the period length of 60 given via the `otpauth://` URI.

### Expected behavior

Use the `period`-value as given via the `otpauth://` URI.

### Environment

- OS: Ubuntu 21.10
- OS version: Linux t480ssm 5.13.0-52-generic # 59-Ubuntu SMP Wed Jun 15 20:17:13 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: 1.14.0 go1.18 linux amd64
- Installation method: deb package diretly

dominikschulz:
RELEASE_NOTES=[BUGFIX] Use OTP key period

Fixes #2276

### DIFF
--- a/internal/action/otp.go
+++ b/internal/action/otp.go
@@ -143,10 +143,11 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 				return exit.Error(exit.Unknown, err, "Failed to compute OTP token for %s: %s", name, err)
 			}
 			counter++
-			sec.Set("counter", counter)
+			sec.Set("counter", strconv.Itoa(int(counter)))
 			if err := s.Store.Set(ctx, name, sec); err != nil {
 				out.Errorf(ctx, "Failed to persist counter value: %s", err)
 			}
+			debug.Log("Saved counter as %d", counter)
 		}
 
 		now := time.Now()
--- a/internal/action/otp.go
+++ b/internal/action/otp.go
@@ -16,6 +16,7 @@ import (
 	"github.com/gopasspw/gopass/pkg/termio"
 	"github.com/mattn/go-tty"
 	potp "github.com/pquerna/otp"
+	"github.com/pquerna/otp/hotp"
 	"github.com/pquerna/otp/totp"
 	"github.com/urfave/cli/v2"
 )
@@ -111,14 +112,27 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 			return exit.Error(exit.Unknown, err, "No OTP entry found for %s: %s", name, err)
 		}
 
-		token, err := totp.GenerateCodeCustom(two.Secret(), time.Now(), totp.ValidateOpts{
-			Period:    uint(two.Period()),
-			Skew:      1,
-			Digits:    potp.DigitsSix,
-			Algorithm: potp.AlgorithmSHA1,
-		})
-		if err != nil {
-			return exit.Error(exit.Unknown, err, "Failed to compute OTP token for %s: %s", name, err)
+		var token string
+		switch two.Type() {
+		case "totp":
+			token, err = totp.GenerateCodeCustom(two.Secret(), time.Now(), totp.ValidateOpts{
+				Period:    uint(two.Period()),
+				Skew:      1,
+				Digits:    potp.DigitsSix,
+				Algorithm: potp.AlgorithmSHA1,
+			})
+			if err != nil {
+				return exit.Error(exit.Unknown, err, "Failed to compute OTP token for %s: %s", name, err)
+			}
+		case "hotp":
+			// TODO: Counter shouldn't be fixed.
+			token, err = hotp.GenerateCodeCustom(two.Secret(), 1, hotp.ValidateOpts{
+				Digits:    potp.DigitsSix,
+				Algorithm: potp.AlgorithmSHA1,
+			})
+			if err != nil {
+				return exit.Error(exit.Unknown, err, "Failed to compute OTP token for %s: %s", name, err)
+			}
 		}
 
 		now := time.Now()
--- a/internal/action/otp.go
+++ b/internal/action/otp.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net/url"
+	"strconv"
+	"strings"
 	"time"
 
 	"github.com/gopasspw/gopass/internal/action/exit"
@@ -100,6 +103,13 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 		go waitForKeyPress(ctx, cancel)
 	}
 
+	// only used for the HOTP case as a fallback
+	var counter uint64 = 1
+	if sv, found := sec.Get("counter"); found && sv != "" {
+		if iv, err := strconv.ParseUint(sv, 10, 64); iv != 0 && err == nil {
+			counter = iv
+		}
+	}
 	for {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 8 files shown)


pkg/otp/otp.go
```diff
@@ -31,12 +33,10 @@ func Calculate(name string, sec gopass.Secret) (twofactor.OTP, string, error) {
 	if otpURL != "" {
 		debug.Log("found otpauth url: %s", out.Secret(otpURL))
 
-		return twofactor.FromURL(otpURL) //nolint:wrapcheck
+		return otp.NewKeyFromURL(otpURL) //nolint:wrapcheck
```
AnomalRoil, line 36: My understanding it that this supports both TOTP and HOTP:
 https://github.com/pquerna/otp/blob/60112ee2a95553a491c22956619ef30260ec93e8/otp.go#L55

pkg/otp/otp.go
```diff
@@ -45,47 +45,43 @@ func Calculate(name string, sec gopass.Secret) (twofactor.OTP, string, error) {
...
+		k, err := otp.NewKeyFromURL(secKey)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse otpauth URL: %w", err)
+		}
+
+		return k, nil
 	}
 
-	otp, err := twofactor.NewGoogleTOTP(twofactor.Pad(secKey))
+	debug.Log("assembling otpauth URL from secret only (%q), using defaults", out.Secret(secKey))
```
AnomalRoil, line 58: This is logging a secret, not great.

pkg/otp/otp.go
```diff
@@ -45,47 +45,43 @@ func Calculate(name string, sec gopass.Secret) (twofactor.OTP, string, error) {
...
+
+	// otpauth://totp/Example:alice@google.com?secret=JBSWY3DPEHPK3PXP&issuer=Example
+	key, err := otp.NewKeyFromURL(fmt.Sprintf("otpauth://totp/new?secret=%s&issuer=gopass", secKey))
```
AnomalRoil, line 61: This is not supporting HOTP, we might want to errror out if HOTP was requested, no?

internal/action/otp.go
```diff
@@ -108,18 +105,30 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 			return nil
 		default:
 		}
-		two, label, err := otp.Calculate(name, sec)
+
+		two, err := otp.Calculate(name, sec)
```
AnomalRoil, line 109: `two` supports both HOTP and TOTP as per the doc of `pquerna/otp` if it took the happy path where the URL is parsed easily in `otp.Calculate`.

internal/action/otp.go
```diff
@@ -108,18 +105,30 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
 		if err != nil {
 			return exit.Error(exit.Unknown, err, "No OTP entry found for %s: %s", name, err)
 		}
-		token := two.OTP()
+
+		token, err := totp.GenerateCodeCustom(two.Secret(), time.Now(), totp.ValidateOpts{
```
AnomalRoil, line 114: I guess we might want to switch on:
https://github.com/pquerna/otp/blob/60112ee2a95553a491c22956619ef30260ec93e8/otp.go#L99

if we really want to support both HOTP and TOTP.

AnomalRoil:
So, my tests show that the old binary and the new binary both work fine with TOTP and give the same value as Google Authenticator.

But the HOTP part isn't working with the new binary.

pkg/otp/otp.go
dominikschulz: Yes, it's logging a secret. But it's doing so responsibly. See https://github.com/gopasspw/gopass/blob/master/pkg/debug/debug.go#L210 and https://github.com/gopasspw/gopass/blob/master/internal/out/print.go#L27.

internal/action/otp.go
dominikschulz: Done.

pkg/otp/otp.go
dominikschulz: Both should work, yes.

pkg/otp/otp.go
dominikschulz: How would we know that HOTP was requested?

dominikschulz:
I'm not sure if I ever used HTOP. Do you have an example that I could use to reproduce the issue?

pkg/otp/otp.go
AnomalRoil: The URL changes, it says "otpauth://hotp" instead.
Or we use the KV `hotp` instead of `otpauth` or `totp`.

Also I guess just relying on the library's `Type` method should be enough.

AnomalRoil:
I guess
`otpauth://hotp/hotp@example.com?secret=N6CIHPYDGHI6QC4ZV7Q3S3FXA22BZOCY&issuer=example.com&digits=6&algorithm=SHA1&counter=1` should work.

Old binary gives `683787` with `&counter=1` and `044228` with `&counter=2`

internal/action/otp.go
```diff
@@ -111,14 +112,27 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
+				Period:    uint(two.Period()),
+				Skew:      1,
+				Digits:    potp.DigitsSix,
+				Algorithm: potp.AlgorithmSHA1,
+			})
+			if err != nil {
+				return exit.Error(exit.Unknown, err, "Failed to compute OTP token for %s: %s", name, err)
+			}
+		case "hotp":
+			// TODO: Counter shouldn't be fixed.
```
AnomalRoil, line 128: Maybe we should just prompt the user for their input for the counter?

internal/action/otp.go
```diff
@@ -111,14 +112,27 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
...
+			token, err = hotp.GenerateCodeCustom(two.Secret(), 1, hotp.ValidateOpts{
+				Digits:    potp.DigitsSix,
+				Algorithm: potp.AlgorithmSHA1,
```
AnomalRoil, line 131: Actually this is sometimes contained in the URL sometimes. But I guess we could just say "not supported but for SHA1"?

AnomalRoil (dismissed):

pkg/otp/otp.go
dominikschulz: Actually this is only the fallback. If we end up having only the secret we can only assume that it's TOTP. Everything else would require information we just don't have.

Maybe it's not worth keeping this fallback after all. But I guess somebody would complain if we drop it.

internal/action/otp.go
dominikschulz: Done.

internal/action/otp.go
dominikschulz: I think we actually need to keep track of the counter, don't we?

AnomalRoil (approved):

PR #2284, opened 2022-07-22
Add more info in case when gpg fails to decode passwords with no error message. Ie (❌  empty secret) case.

PiotrCzapla:
Reformatted FAQ to. allow for linking to particular question.  I will add the "empty secret" section in the next commit


dominikschulz (approved):
LGTM

PR #2285, opened 2022-07-22
Add explanation for empty secret warning. Fixes #2283 

### LINKED ISSUE #2283: gopass  thinks my secrets are empty on ubuntu 18.04 (gpg 2.2.4)
### Summary
I'm setting up gopass on google colab, I've managed to configure gpg along with preset-password so that I'm able to fetch data from private github repo. So gpg works. 

But gopass thinks it isn't. gopass clone fetches private repo using gpg-agent and then complains that there is no usable keys. If I skip key generation, the decryption fails with message '❌  empty secret'.

Unfortunately debugging is problematic as gopass suppress information in this line:
https://github.com/gopasspw/gopass/blob/d122a9c24b698acf1190f61ce5653d0c04741539/internal/action/setup.go#L231-L234

### Steps To Reproduce
it is obvious from code.

### Expected behavior
gopass informs user why it can't access any keys or there is a way to check that
### Environment
google colab, installed today via direct download from github

PiotrCzapla:
I was too slow adding the commit to previous PR #2284. Here is the propose paragraph for the 'empty  secret warning.'

I would then add a link to the empty secret message as follows:
 https://github.com/gopasspw/gopass/blob/master/docs/faq.md#empty-secret

But It would be even better if you could expose a direct redirection on https://gopass.pw/ so that we could write https://www.gopass.pw/faq#empty-secret

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -36,6 +36,9 @@ Set the `auto-expand-secmem` option in your gpg-agent.conf, if your version of G
 ## I'm getting `Path too long for Unix domain socket` errors, usually on MacOS.
 This can be fixed by setting `export TMPDIR=/tmp` (or any other suiteable location with a path shorter than 80 characters).
 
+## Empty secret?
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
LGTM

dominikschulz:
The plan sounds good. Thanks a lot.

At first we need to use the long path but I will look into setting up a redirection or mirror the faq on the website.

PR #2293, opened 2022-07-29
Enforce min and max length for string and password types in the wizard

### LINKED ISSUE #2292: `gopass create` wizard template `min` & `max` attributes are not enforced
### Summary

The `min` & `max` options for the wizard attributes of `gopass create` appear to be unimplemented and seem to have no effect.
https://github.com/gopasspw/gopass/blob/16c071a78070562a4c26fdef8327d79b2bc587c7/internal/create/wizard.go#L35-L42

### Steps To Reproduce

1. Invoke `gopass create`
2. Select the stock "PIN Code (numerical)" template when prompted
3. When prompted for "Authority (Issuer)" press enter without providing any input
4. Observe that the template attribute's minimum length requirement was not respected

https://github.com/gopasspw/gopass/blob/16c071a78070562a4c26fdef8327d79b2bc587c7/internal/create/wizard.go#L99-L106

### Expected behavior

Because the authority attribute specifies a minimum length of 1, the wizard should immediately abort when given an input value that is too short (and vice versa for the max attribute).

The ergonomics and implementation should essentially be the same as the current wizard behavior for when a hostname cannot be extracted on an attribute of the `hostname` type

https://github.com/gopasspw/gopass/blob/16c071a78070562a4c26fdef8327d79b2bc587c7/internal/create/wizard.go#L234-L237

### Environment

- OS: Manjaro Linux (Stable)
- OS version: `Linux chao-desktop 5.15.55-1-MANJARO #1 SMP PREEMPT Fri Jul 15 09:45:39 UTC 2022 x86_64 GNU/Linux`
- gopass Version: `gopass 1.14.3 go1.18.3 linux amd64`
- Installation method: Arch community repository via pacman

dominikschulz:
RELEASE_NOTES=[BUGFIX] Wizard: Enforce min and max length.

Fixes #2292

### DIFF
--- a/internal/create/wizard.go
+++ b/internal/create/wizard.go
@@ -222,6 +222,12 @@ func mkActFunc(tpl Template, s *root.Store, cb ActionCallback) func(context.Cont
 				if err != nil {
 					return err
 				}
+				if v.Min > 0 && len(sv) < v.Min {
+					return fmt.Errorf("%s is too short (needs %d)", v.Name, v.Min)
+				}
+				if v.Max > 0 && len(sv) > v.Min {
+					return fmt.Errorf("%s is too long (at most %d)", v.Name, v.Max)
+				}
 				if wantForName[k] {
 					nameParts = append(nameParts, sv)
 				}
@@ -258,6 +264,12 @@ func mkActFunc(tpl Template, s *root.Store, cb ActionCallback) func(context.Cont
 					password, err = termio.AskForPassword(ctx, v.Prompt, true)
 					if err != nil {
 						return err
+					}
+					if v.Min > 0 && len(password) < v.Min {
+						return fmt.Errorf("%s is too short (needs %d)", v.Name, v.Min)
+					}
+					if v.Max > 0 && len(password) > v.Min {
+						return fmt.Errorf("%s is too long (at most %d)", v.Name, v.Max)
 					}
 				}
 


chaorace:
I believe that this actually fixes #2292 :wink:

internal/create/wizard.go
```diff
@@ -222,6 +222,12 @@ func mkActFunc(tpl Template, s *root.Store, cb ActionCallback) func(context.Cont
 				if err != nil {
 					return err
 				}
+				if v.Min > 0 && len(sv) < v.Min {
+					return fmt.Errorf("%s is too short (needs %d)", v.Name, v.Min)
+				}
+				if v.Max > 0 && len(sv) > v.Min {
```
AnomalRoil, line 228:
```suggestion
				if v.Max > 0 && len(sv) > v.Max {
```

internal/create/wizard.go
```diff
@@ -259,6 +265,12 @@ func mkActFunc(tpl Template, s *root.Store, cb ActionCallback) func(context.Cont
 					if err != nil {
 						return err
 					}
+					if v.Min > 0 && len(password) < v.Min {
+						return fmt.Errorf("%s is too short (needs %d)", v.Name, v.Min)
+					}
+					if v.Max > 0 && len(password) > v.Min {
```
AnomalRoil, line 271:
```suggestion
					if v.Max > 0 && len(password) > v.Max {
```

AnomalRoil (changes requested):

Issue #2282, opened 2022-07-21
Pop!_OS NO_PUBKEY


gray380:
### Summary

The following message has appeared: 
```
W: GPG error: https://packages.gopass.pw/repos/gopass stable InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 686E9DE6E1AEFFBE
E: The repository 'https://packages.gopass.pw/repos/gopass stable InRelease' is not signed.
N: Updating from such a repository can't be done securely, and is therefore disabled by default
```
### Steps To Reproduce

```
$ curl http://packages.gopass.pw/repos/gopass/gopass-archive-keyring.gpg | sudo tee /usr/share/keyrings/gopass-archive-keyring.gpg
$ cat << EOF | sudo tee /etc/apt/sources.list.d/gopass.sources
Types: deb
URIs: https://packages.gopass.pw/repos/gopass
Suites: stable
Architectures: all amd64 arm64 armhf
Components: main
Signed-By: /usr/share/keyrings/gopass-archive-keyring.gpg
EOF
$ sudo apt update
```

### Environment

- OS: [Pop!_OS 22.04 LTS]
- OS version: [5.17.15-76051715-generic #202206141358~1655919116~22.04~1db9e34 SMP PREEMPT Wed Jun 22 19 x86_64 x86_64 x86_64 GNU/Linux]
- gopass Version: [gopass 1.14.3 go1.18.2 linux amd64]
- Installation method: [gopass repo]

dominikschulz:
I've never used Pop OS. Maybe they don't support the `Signed-By`?

I'd like to give it a try some day. But until then you could try adding the the key to the global apt keyring with apt-key add, if you want.

gray380:
This command works for me:
```bash
wget -O- http://packages.gopass.pw/repos/gopass/gopass-archive-keyring.gpg | gpg --dearmor | sudo tee /usr/share/keyrings/gopass-archive-keyring.gpg
```

dominikschulz:
Hmm, the gpg file should be binary and not need dearmor.

Let me verify this.

AnomalRoil:
I'm currently running Pop!_OS on a laptop in case it needs some testing.

I can confirm the same `wget` is giving out binary data that's not armored, btw.

AnomalRoil:
`curl http://packages.gopass.pw/repos/gopass/gopass-archive-keyring.gpg` on the other hand seems to return nothing.

[...]

PR #2298, opened 2022-08-02
Correcting instructions for apt install

AnomalRoil:
Fixing #2282 

I think the culprit is cloudflare, but not sure.

curl is getting a HTTP/1.1 301 Moved Permanently, but since wget works 🤷🏻‍♂️

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -181,7 +181,7 @@ When installing on Ubuntu or Debian you can either download the `deb` package,
 [install manually or build from source](#installing-from-source) or use our APT repository.
 
 ```bash
-$ curl http://packages.gopass.pw/repos/gopass/gopass-archive-keyring.gpg | sudo tee /usr/share/keyrings/gopass-archive-keyring.gpg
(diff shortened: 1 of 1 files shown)


docs/setup.md
```diff
@@ -181,7 +181,7 @@ When installing on Ubuntu or Debian you can either download the `deb` package,
 [install manually or build from source](#installing-from-source) or use our APT repository.
 
 ```bash
-$ curl http://packages.gopass.pw/repos/gopass/gopass-archive-keyring.gpg | sudo tee /usr/share/keyrings/gopass-archive-keyring.gpg
```
dominikschulz, line 184: Actually the problem is that the keyring should be fetched over https, not http. Then curl wouldn't get a redirect and would succeed.

I think it's correct to fail in this case.

AnomalRoil: Ah, that's why. Good catch.

dominikschulz (approved):
LGTM

PR #2314, opened 2022-08-19
Purge incorrect passwords from age's askpass cache

duxovni:
Currently, if I mistype my store password in the gopass cli using the age backend, I need to close and restart the cli, because it keeps trying to use the cached incorrect password.

Possible other approaches:
- Since the password callback and the password purge callback are closely interrelated, it might make sense to have a single `ctxutil` method that sets both of them together.
- Instead of having a callback that returns a password at all, there could be a callback that attempts to decrypt data using the password, and that callback is responsible for clearing its own cache if it fails to decrypt.


dominikschulz (approved):
LGTM

Issue #2317, opened 2022-08-28
Suggested method for "Securing Your Editor" still leaks passwords.


BoostCookie:
The suggested method here https://github.com/gopasspw/gopass/blob/master/docs/setup.md#securing-your-editor, which is adding
```
au BufNewFile,BufRead /dev/shm/gopass.* setlocal noswapfile nobackup noundofile
```

to the vimrc, is not enough. The viminfo-file still gets written with content of the decrypted file. What should be used instead is
```
autocmd BufNewFile,BufRead /dev/shm/gopass* setlocal noswapfile nobackup noundofile viminfo=""
```
for vim and
```
autocmd BufNewFile,BufRead /dev/shm/gopass* setlocal noswapfile nobackup noundofile shada=""
```
for neovim.
So the path is wrong (the dot at the end) and it is still writing viminfo/shada.
I do not know if the MacOS path is correct, but the viminfo/shada-stuff is of course similar.

micxer:
Using this instead of the info in the setup.md still gives me the warning message. The path of the temporary file looks different as well: `/var/folders/s_/84wxnwrs0js3bvcq08wcpr1h0000gn/T/gopass-edit1483846318`.

BoostCookie:
@micxer, what OS are you on? I'm on Artix Linux and my path always is `/dev/shm/gopass*`.

> still gives me the warning message.

Are you actually using a version of gopass that has this commit already included?

PR #2318, opened 2022-08-28
Correct (neo)vim settings

BoostCookie:
The previous recommended vim settings still allowed data to leak into
the .viminfo file
For Linux even the matching path was wrong which means that the
(insufficient) rules were not even used.

RELEASE_NOTES=[DOCUMENTATION] fix in recommended vim setting

Signed-off-by: Stefan Gehr <stefangehr@protonmail.com>

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -271,18 +271,20 @@ If `$GOPATH/bin` is in your `$PATH`, you can now run `gopass` from anywhere on y
 
 ### Securing Your Editor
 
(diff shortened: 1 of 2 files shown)


BoostCookie:
This fixes https://github.com/gopasspw/gopass/issues/2317.

dominikschulz (approved):
LGTM. Thanks a lot!

dominikschulz:
I'll keep the issue open after this is merged. I'd like to see if we can remove this config sniffing and just invoke known editors with the proper options automatically.

PR #2341, opened 2022-09-13
Do not omit shadowed entries in ls output

### LINKED ISSUE #2338: `gopass list --flat` doesn't show secrets with the same names as folders
### Summary
If a secret has the same name as a folder, it won't be listed by `gopass list --flat`.

### Steps To Reproduce

```
gopass add foo/bar
gopass add foo/bar/baz
gopass list --flat
```

### Expected behavior

Should return:
```
foo/bar
foo/bar/baz
```

### Environment

- OS: NixOS
- OS version: 22.11
- gopass Version: 1.14.5

dominikschulz:
Fixes #2338

RELEASE_NOTES=[BUGFIX] Do not shadow entries behind folders.

### DIFF
--- a/internal/tree/node.go
+++ b/internal/tree/node.go
@@ -2,12 +2,14 @@ package tree
 
 import (
 	"bytes"
+
+	"github.com/gopasspw/gopass/pkg/debug"
 )
 
 // Node is a tree node.
 type Node struct {
 	Name     string
-	Type     string
+	Leaf     bool
 	Template bool
 	Mount    bool
 	Path     string
@@ -40,7 +42,7 @@ func (n Node) Equals(other Node) bool {
 		return false
 	}
 
-	if n.Type != other.Type {
+	if n.Leaf != other.Leaf {
 		return false
 	}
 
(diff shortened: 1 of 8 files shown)


internal/tree/node.go
```diff
@@ -59,6 +61,40 @@ func (n Node) Equals(other Node) bool {
 	return true
 }
 
+func (n Node) Merge(other Node) *Node {
```
dominikschulz, line 64: This method needs some more explanation.

internal/tree/node.go
```diff
@@ -113,7 +153,7 @@ func (n *Node) format(prefix string, last bool, maxDepth, curDepth int) string {
 
 // Len returns the length of this subtree.
 func (n *Node) Len() int {
-	if n.Type == "file" {
+	if n.Subtree == nil {
 		return 1
 	}
```
dominikschulz, line 159: We need to add 1 to l if n.Leaf is true.

internal/tree/root.go
```diff
@@ -80,11 +82,15 @@ func (r *Root) insert(path string, template bool, mountPath string) error {
...
+		debug.Log("[%d] %s -> Node: %+v", i, e, n)
+
+		node := t.Insert(n)
+		debug.Log("node after insert: %+v", node)
+
 		// do we need to extend an existing subtree?
 		if i < len(p)-1 && node.Subtree == nil {
 			node.Subtree = NewTree()
-			node.Type = "dir"
+			// node.Type = "dir"
```
dominikschulz, line 93: Remove this.

internal/tree/tree.go
```diff
@@ -44,28 +46,27 @@ func (t *Tree) Equals(other *Tree) bool {
 }
 
 // Insert adds a new node at the right position.
-func (t *Tree) Insert(node *Node) (*Node, error) {
-	pos, found := t.find(node.Name)
+func (t *Tree) Insert(node *Node) *Node {
+	pos, found := t.findPositionFor(node.Name)
```
dominikschulz, line 50: `found` implies this to be a bool. Maybe we call it `existing` or something.

PR #2342, opened 2022-09-16
Restore TTY when aborting TOTP ticker with Ctrl+C

### LINKED ISSUE #2320: gopass totp disabled terminal echo
### Summary
After running `gopass totp ...`, and ctrl-c to exit, terminal input is hidden, because echo was disabled.

### Steps To Reproduce
- Run `gopass totp ...`
- Use Ctrl-C to exit
- Try to type something
- Input is not displayed on the terminal

Note:
- If `q` is used to exit, then the terminal input IS correctly displayed.

### Expected behavior
- Terminal input should be displayed in both cases. 
- Terminal echo should be re-enabled as soon as gopass has decrypted the secret, long before doing anything else.

### Environment
- OS: Gentoo Linux
- OS version: `Linux REDACTED 5.18.15-gentoo-dist #1 SMP PREEMPT_DYNAMIC Fri Jul 29 22:03:23 -00 2022 x86_64 Intel(R) Core(TM) REDACTED GenuineIntel GNU/Linux`
- gopass Version: 1.14.4
- Installation method: distro package manager

dominikschulz:
Fixes #2320

RELEASE_NOTES=[BUGFIX] Handle Ctrl+C in TOTP

### DIFF
--- a/internal/action/otp.go
+++ b/internal/action/otp.go
@@ -53,35 +53,35 @@ func tickingBar(ctx context.Context, expiresAt time.Time, bar *termio.ProgressBa
 	}
 }
 
-func waitForKeyPress(ctx context.Context, cancel context.CancelFunc) {
-	tty, err := tty.Open()
+func waitForKeyPress(ctx context.Context, cancel context.CancelFunc) (func(), func()) {
+	tty1, err := tty.Open()
 	if err != nil {
 		out.Errorf(ctx, "Unexpected error opening tty: %v", err)
 		cancel()
 	}
 
-	defer func() {
-		_ = tty.Close()
-	}()
+	return func() {
+			for {
+				select {
+				case <-ctx.Done():
+					return // returning not to leak the goroutine.
+				default:
+				}
 
-	for {
-		select {
-		case <-ctx.Done():
-			return // returning not to leak the goroutine.
-		default:
+				r, err := tty1.ReadRune()
+				if err != nil {
+					out.Errorf(ctx, "Unexpected error opening tty: %v", err)
+				}
+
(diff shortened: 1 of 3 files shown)


dominikschulz:
This makes an tricky code path even more tricky :/

dominikschulz:
The failing test looks like a flake to me.

dominikschulz:
No, it's an actual concurrency issue.

main.go
```diff
@@ -70,6 +70,7 @@ func main() {
 	go func() {
 		select {
 		case <-sigChan:
+			fmt.Println("SIGCHAN")
```
AnomalRoil, line 73: Is this really meant to be left in prod code? Maybe make it a debug level log?

internal/action/otp.go
```diff
@@ -97,7 +97,9 @@ func (s *Action) otp(ctx context.Context, name, qrf string, clip, pw, recurse bo
 	skip := ctxutil.IsHidden(ctx) || pw || qrf != "" || !ctxutil.IsTerminal(ctx) || !ctxutil.IsInteractive(ctx) || clip
 	if !skip {
 		// let us monitor key presses for cancellation:.
-		go waitForKeyPress(ctx, cancel)
+		runFn, cleanupFn := waitForKeyPress(ctx, cancel)
+		go runFn()
+		defer cleanupFn()
```
AnomalRoil, line 102: Good catch!

AnomalRoil (dismissed):

main.go
dominikschulz: Ohmy. No, this should go away of course. Thanks :)

PR #2343, opened 2022-09-16
Set vim options instead of sniffing the config

### LINKED ISSUE #2317: Suggested method for "Securing Your Editor" still leaks passwords.

dominikschulz:
Fixes #2317

RELEASE_NOTES=[ENHANCEMENT] Set vim options instead of sniffing

### DIFF
--- a/internal/editor/editor.go
+++ b/internal/editor/editor.go
@@ -7,16 +7,11 @@ import (
 	"io"
 	"os"
 	"os/exec"
-	"path/filepath"
-	"regexp"
 	"runtime"
-	"strings"
 
 	"github.com/fatih/color"
-	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
-	"github.com/gopasspw/gopass/pkg/fsutil"
 	"github.com/gopasspw/gopass/pkg/tempfile"
 	shellquote "github.com/kballard/go-shellquote"
 )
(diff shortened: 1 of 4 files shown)


internal/editor/editor.go
```diff
@@ -128,3 +86,25 @@ func Invoke(ctx context.Context, editor string, content []byte) ([]byte, error)
...
+	viminfo := `viminfo=""`
+	if editor == "neovim" {
+		viminfo = `shada=""`
+	}
+
+	return []string{
+		"-i", "NONE", // disable viminfo
+		"-n", // disable swap
+		"-c",
+		fmt.Sprintf("autocmd BufNewFile,BufRead %s setlocal noswapfile nobackup noundofile %s", path, viminfo),
```
AnomalRoil, line 108: I guess we might want to document the fact that we're setting these somewhere in the documentation, no?

internal/editor/editor.go
```diff
@@ -128,3 +86,25 @@ func Invoke(ctx context.Context, editor string, content []byte) ([]byte, error)
...
 	return nContent, nil
 }
+
+func vimOptions(editor string) []string {
+	if editor != "vi" && editor != "vim" && editor != "neovim" {
+		return []string{}
+	}
+
+	path := "/dev/shm/gopass*"
+	if runtime.GOOS == "darwin" {
```
AnomalRoil, line 96: How does that work on Windows? (Are we even using vim on windows?)

internal/editor/editor.go
```diff
@@ -27,47 +22,9 @@ var (
...
-
-	if vimOptsRe.Match(buf) {
-		debug.Log("Recommended settings found in %s", vrc)
-
-		return nil
-	}
-
-	debug.Log("%s did not match %s", string(buf), vimOptsRe)
-	out.Warningf(ctx, "Vim might leak credentials. Check your setup.\nhttps://go.gopass.pw/setup#securing-your-editor")
-
```
AnomalRoil, line 67: Shouldn't we also change this documentation entry to reflect the new behaviour?

AnomalRoil (dismissed):

internal/editor/editor.go
dominikschulz: I have no idea. Someone else would need to test that.

PR #2344, opened 2022-09-17
Do not ignore symlinks when listing store contents

### LINKED ISSUE #2173: gopass ignoring symlink to other directory
### Summary
I have a symlink in `~/.pass/` that points to a separate directory outside the `~/.pass/` directory tree,  and is not shown by `gopass list`. `gopass show ...` does correctly load secrets from it.

The original `pass` does correctly list the secrets that are contained therein.

### Steps To Reproduce
1. Initialize `~/.pass/` & add a secret
2. Initialize `/tmp/demo-pass/` ` & add a secret
3. Symlink `ln -sf /tmp/demo-pass ~/.pass/demo-pass`
4. Try to list everything `gopass list --flat`

### Expected behavior
All of the secrets in both directories should be included.

### Environment
- OS: Gentoo Linux
- OS version: `Linux REDACTED 5.17.0-rc2-x86_64-REDACTED #92 SMP PREEMPT Sun Feb 6 11:39:50 PST 2022 x86_64 AMD Ryzen 7 3700X 8-Core Processor AuthenticAMD GNU/Linux`
- gopass Version: `gopass 1.13.0 (2022-01-18 23:25:13) go1.17.6 linux amd64`
- Installation method: distro package

dominikschulz:
Fixes #2173

RELEASE_NOTES=[BUGFIX] Do not ignore symlinks when listing

### DIFF
--- a/internal/backend/storage/fs/walk.go
+++ b/internal/backend/storage/fs/walk.go
@@ -1,0 +1,59 @@
+package fs
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/gopasspw/gopass/pkg/debug"
+)
+
+func walkSymlinks(path string, walkFn filepath.WalkFunc) error {
+	w := &walker{
+		seen: make(map[string]bool, 100),
+	}
+
+	return w.walk(path, path, walkFn)
+}
+
+type walker struct {
+	seen map[string]bool
+}
+
+func (w *walker) walk(filename, linkDir string, walkFn filepath.WalkFunc) error {
+	sWalkFn := func(path string, info fs.FileInfo, err error) error {
+		fname, err := filepath.Rel(filename, path)
+		if err != nil {
+			return err
+		}
+		path = filepath.Join(linkDir, fname)
+
+		if info.Mode()&fs.ModeSymlink == fs.ModeSymlink {
+			destPath, err := filepath.EvalSymlinks(path)
+			if err != nil {
+				return err
+			}
+
+			// avoid loops
+			if w.seen[destPath] {
+				debug.Log("Symlink loop detected at %s!", destPath)
+
+				return nil
+			}
+			w.seen[destPath] = true
+
+			destInfo, err := os.Lstat(destPath)
+			if err != nil {
+				return walkFn(path, destInfo, err)
+			}
+
+			if destInfo.IsDir() {
+				return w.walk(destPath, path, walkFn)
+			}
+		}
+
+		return walkFn(path, info, err)
+	}
+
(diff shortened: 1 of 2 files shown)


internal/backend/storage/fs/walk.go
```diff
@@ -0,0 +1,59 @@
...
+	"io/fs"
+	"os"
+	"path/filepath"
+
+	"github.com/gopasspw/gopass/pkg/debug"
+)
+
+func walkSymlinks(path string, walkFn filepath.WalkFunc) error {
+	w := &walker{
+		seen: make(map[string]bool, 100),
```
AnomalRoil, line 13: Why 100 and not 400 or 1000?
I feel like 100 might be a bit small, meaning we're going to be doing hundreds of new allocation later on, but 1000 is probably overkill. I'd lean on a 250 or 300 value being more typical, but that's out of my own workflow, so not necessarily representative

internal/backend/storage/fs/walk.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (w *walker) walk(filename, linkDir string, walkFn filepath.WalkFunc) error {
+	sWalkFn := func(path string, info fs.FileInfo, err error) error {
+		fname, err := filepath.Rel(filename, path)
+		if err != nil {
+			return err
+		}
+		path = filepath.Join(linkDir, fname)
+
+		if info.Mode()&fs.ModeSymlink == fs.ModeSymlink {
```
AnomalRoil, line 31: Nice one

internal/backend/storage/fs/walk.go
```diff
@@ -0,0 +1,59 @@
...
+				return nil
+			}
+			w.seen[destPath] = true
+
+			destInfo, err := os.Lstat(destPath)
+			if err != nil {
+				return walkFn(path, destInfo, err)
+			}
+
+			if destInfo.IsDir() {
```
AnomalRoil, line 50: Do we want to have a max recursion depth as well?

AnomalRoil (approved):

internal/backend/storage/fs/walk.go
dominikschulz: For an application like gopass I'd argue that this doesn't matter at all. We could also just start at zero. I agree that using a magic constant is not good style. But using a different guessed constant isn't good style either. Maybe I'll just start this at zero.

internal/backend/storage/fs/walk.go
dominikschulz: Oh, Go will fail with a nice error message if recurse too deep. I tried that.

PR #2349, opened 2022-09-24
Allow modifying and removing default create templates

### LINKED ISSUE #2291: Allow disabling the default `gopass create` wizard templates
Currently, the two default wizard templates for `gopass create` ("Website login" & "PIN Code (numerical)") seem to always load in as options for the wizard.

It would be nice to have a config option to turn this off so that only custom templates appear as options in that wizard menu. I particularly want this because both stock templates are marked "priority: 0", which forces you to use negative priority values in custom templates if you want them to appear first in the list (important if you want to change the default option).

dominikschulz:
Fixes #2291

RELEASE_NOTES=[ENHANCEMENT] Allow modifying default create templates

### DIFF
--- a/internal/create/wizard.go
+++ b/internal/create/wizard.go
@@ -28,6 +28,7 @@ import (
 const (
 	defaultLength     = 24
 	defaultXKCDLength = 4
+	tplPath           = ".gopass/create/"
 )
 
 // Attribute is a credential attribute that is being asked for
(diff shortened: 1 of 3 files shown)


internal/create/wizard.go
```diff
@@ -28,6 +28,7 @@ import (
 const (
 	defaultLength     = 24
 	defaultXKCDLength = 4
+	tplPath           = ".gopass/create/"
```
AnomalRoil, line 31: Maybe it would be good to add the related documentation to https://github.com/gopasspw/gopass/blob/master/docs/commands/templates.md ?

dominikschulz: It's already documented in docs/commands/create.md.

PR #2365, opened 2022-09-28
Use different concurrency groups for different workflows

dominikschulz:
See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency

RELEASE_NOTES=n/a

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,13 +2,18 @@ name: Build gopass
 
 on:
   push:
-    branches: [ master ]
+    branches:
+      - master
   pull_request:
-    branches: [ master ]
+    branches:
+      - master
 
 concurrency:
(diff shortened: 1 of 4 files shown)


AnomalRoil:
TIL, nice I should do that on my own projects too 😄

dominikschulz:
Not sure if that was clear, but this did lead to workflows "randomly" being canceled for no obvious reason.
That was very confusing until I found this.

Now they seem to work as expected.

AnomalRoil:
Yeah, I guess the issue was that both the Grype and the build one had `-build` at the end.

You need unique concurrency names across workflows. Another option could have been to rename the Grype one with `-grype`.

dominikschulz:
That was my first attempt. Until I saw this here mentioned in the docs. So I went with the best practice.

PR #2373, opened 2022-10-02
re-encrypt entire directory when same key chosen with recipients add --force

anjandev:
I added a new subkey via gpg to my public key. I wanted re-encrypt my entire password store to accessible via this new subkey and I could not cause of the "recipient already in store" error. This change allows you to reencrypt the entire directory with the public key again provided you use the force flag with "gopass recipients add".

RELEASE_NOTES=[ENHANCEMENT] If a key id is already a recipient, allow users to re-encrypt entire password store with the new public key via "gopass recipients add --force"

Signed-off-by: Anjandev Momi <anjan@momi.ca>

### DIFF
--- a/internal/store/leaf/recipients.go
+++ b/internal/store/leaf/recipients.go
@@ -13,6 +13,8 @@ import (
 	"github.com/gopasspw/gopass/internal/store"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
+
+	"github.com/gopasspw/gopass/pkg/termio"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


dominikschulz:
This looks good, thanks. I just want to give it a try first.

internal/store/leaf/recipients.go
```diff
@@ -71,7 +71,7 @@ func (s *Store) AddRecipient(ctx context.Context, id string) error {
 	debug.Log("new recipient: %q - existing: %+v", id, rs)
 
 	for _, k := range rs {
-		if k == id {
```
dominikschulz, line 74: This has the side effect of adding the same key ID to the recipients list twice. I don't like that.

What about moving this loop and the append to a small helper function that only returns an error when force is not set but only appends if the key is indeed missing? That way we'd still be able to re-encrypt although the recipients didn't change.

Or we just drop the error and always re-encrypt, even w/o the force flag?
That would make the implementation more lightweight and easier to use. At the expense that users might accidentially re-encrypt the store (I don't think that's a problem.)

dominikschulz (changes requested):
I'm inclined to always re-encrypt when asked to add a recipient. Even without the force flag.

WDYT?

But at least you'll need to fix the failing tests.

internal/store/leaf/recipients.go
anjandev: So when gopass does "Starting reencrypt" and you ctrl+c out of it, it leaves git in a weird state. Since users might accidentally select the wrong option and try to ctrl+c out of it, I dont think we should just automatically re-encrypt when the same key is selected. Ive implemented a confirm prompt in my latest commit that explains why someone might want to re-encrypt their entire directory. Let me know what you think.

dominikschulz (approved):
This looks nice. Thank you!

I wonder if we could do the nesting a little bit nicer, but we don't need to do another round trip for that right now.

AnomalRoil:
Also notice this is also possible by using the `gopass fsck --decrypt`, no? It will re-encrypt your whole store IIRC, so it's enough to add the subkey on gpg side and then do a `gopass fsck --decrypt`, same in case of key expiration.

PR #2375, opened 2022-10-03
Do not attempt to access OS keychain at all unless the `keychain` option is enabled

### LINKED ISSUE #2374: gopass always prompts me to unlock my (nonexistent) login keyring
### Summary

Every time I run gopass, regardless of what operation I'm performing, immediately on startup it opens a prompt: "Authentication required - The login keyring did not get unlocked when you logged into your computer." I need to hit Cancel twice before it gives up and proceeds with what I wanted to do.

### Steps To Reproduce

Run any gopass subcommand, using the age backend, in an environment that does not have a login keyring.

### Expected behavior

I expect gopass to not prompt me for a login keyring password, since I have a default config with `keychain` set to `false`.

### Environment

- OS: Nixos 22.11
- Desktop environment: greetd and swaywm
- gopass Version: 1.14.9

duxovni:
Fixes #2374.

### DIFF
--- a/internal/backend/crypto/age/askpass.go
+++ b/internal/backend/crypto/age/askpass.go
@@ -82,9 +82,11 @@ func newAskPass(ctx context.Context) *askPass {
 		cache: cache.NewInMemTTL[string, string](time.Hour, 24*time.Hour),
 	}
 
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
LGTM. Thanks for catching this. This keychain access was not intentional.

PR #2376, opened 2022-10-03
Include shadowed entries in flat list output

### LINKED ISSUE #2338: `gopass list --flat` doesn't show secrets with the same names as folders

duxovni:
Fixes #2338.

### DIFF
--- a/internal/action/list_test.go
+++ b/internal/action/list_test.go
@@ -84,6 +84,32 @@ foo2/
 	assert.Equal(t, want, buf.String())
 	buf.Reset()
 
+	// add shadowed entry
+	sec = &secrets.Plain{}
+	sec.SetPassword("123")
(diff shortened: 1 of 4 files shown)


dominikschulz:
The changes look correct, but I want to give them a try first.

Thanks a lot for this contribution!

dominikschulz:
It looks like this was indeed missing from #2341. Thanks a lot!

dominikschulz (approved):
LGTM

PR #2379, opened 2022-10-05
Ignore SIGURG signals

donob4n:
workaround for https://github.com/golang/go/issues/37942

I noticed it when running gopass with strace. It seems working fine ignoring that and strace is more clean and fast.

### DIFF
--- a/main.go
+++ b/main.go
@@ -16,6 +16,7 @@ import (
 	rdebug "runtime/debug"
 	"runtime/pprof"
 	"sort"
+	"syscall"
 	"time"
 
 	"github.com/blang/semver/v4"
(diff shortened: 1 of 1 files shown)


dominikschulz:
I'm generally fine with this PR but I want to test it with and without the patch first.

donob4n:
I don't know what this happens:
```
Check failure on line 52 in main.go

GitHub Actions
/ linux

undefined: syscall.SIGURG
```
It builded fine locally (Alpinelinux-edge) after importing "syscall". Also it builds fine on macOS. I could try to reproduce on an ubuntu MV or container.

dominikschulz:
Maybe you want to try [this](https://stackoverflow.com/questions/63396105/undefined-syscall-sigurg-on-a-working-code) suggestion.

donob4n:
Ah, ok I will try to change it.

donob4n:
Let's try this way or better avoid it if running 'OS == windows' ?

donob4n:
Last also worked for me locally but failed on CI, I'm now trying directly getting the signal from syscall.

donob4n:
Yeah, it worked even for windows hehe

dominikschulz:
Closing in favor of #2440

PR #2385, opened 2022-10-15
Add option for OTP to output password combined with OTP token

lafriks:
There are places where authorization requires adding OTP token at the end of the password. Added option for totp/hotp command to output password combined with OTP token added to end.

Currently it required two commands show + totp to copy password and OTP token to password field separately.


AnomalRoil:
Wow, I have never encountered that kind of authentication requests. Do you have a few examples of projects using this?

lafriks:
For example FreeIPA uses this: https://www.freeipa.org/page/V4/OTP

lafriks:
Code wise it does not add much complexity but of course I can still use my own binary :)

Issue #1798, opened 2021-02-14
Cleanup: Migrate from `ctxutil` to option structs where feasible


dominikschulz:
The `ctxutil` package was introduced to clean up the signatures and make it (much) easier to channel through UI specific settings and callbacks to the place where they are needed. However effective this approach is, it's something of an anti-pattern and most APIs use some kind of optional option structs (e.g. `func (s* Store) Get(key string, opts *GetOptions)`).

This makes is more tedious but also a lot more explicit to channel through settings and callbacks.
We should consider if the added clarity is worth the effort.

dominikschulz:
Much of what should been done has been addressed by #2395. We can always do more cleanups, but for the sake of this issue I'm happy.

PR #2395, opened 2022-10-24
.gitconfig parser

### LINKED ISSUE #1567: Setting config options for substores not working
### Summary
The `gopass config --store <store> key value` command sets the config option globally and not on the mount.

And why is the `autosync` option not available anymore? How can I disable the automatic sync?

### Environment

- OS: Arch
- gopass Version: 1.10.1
- Installation method: repo

### LINKED ISSUE #1764: autoclip for `gopass show`
The current usage for `gopass show` is:
* default: show the password
* if the `-c` option is used: copy the password to the clipboard

However, 9/10 times I want to clip the password and not show it. So it would be nice if there was a config option to clip by default.

It looks like this option was removed in #1502, because it shared the same configuration option as was used for `gopass generate`. I'm not sure why this was a problem, but if it is, we could maybe split it into two options? `autoclip-show` and `autoclip-generate` (alias `autoclip` for backwards compatibility).

### LINKED ISSUE #1819: RFC: Simplify config handling

### LINKED ISSUE #1878: Can auto push be disabled for team sharing store

### LINKED ISSUE #2387: No replacement for `autosync=False`
It looks as though `autosync` has been removed in in the drive towards zero config.

However, it would be great if there was a flag or similar to disable automatic remote interactions. I (and I'm sure others) like to tell my machine when to talk to the internet.

E.g., would something like `gopass --no-sync ...` be possible? Then I could alias `gopass=/usr/bin/gopass --no-sync`.

### LINKED ISSUE #2408: gopass sync with no_push remote and missing GPG public keys
### Summary

I _may_ be doing something wrong here, but when sharing password stores, I often share a password store but essentially in "read-only" mode with a Git push origin of `no_push`. I don't import the GPG keys either as I have no intention of re-encrypting secrets in the "shared" (read-onyl) store.

I _think_ we should detect the `no_push` origin and disable the export/import step and squelt the fatal error that occurs when running `gopass sync`.

### Steps To Reproduce

1. Create two stores
2. Create a pair of GPG keys
3. In the 2nd store, create some secrets encrypted to both GPP keys
4. In the first store, clone and mount the 2nd store with a Git origin of `no_push` (DO NOT import the 2nd GPG key)
5. Run `gopass` sync and observe the behaviour.

### Expected behavior

I expect the export/import step to be skipped and the git pull/push step to be skipped, without error.

### Environment

macOS with latest `goapss`.

### LINKED ISSUE #2418: gitconfig cleanup
We have a number of config options that we either removed or pushed into env vars.
Once we have the gitconfig merged we can start moving these to the config (again).

Adding this to #2395 seems a bit too much, so tracking it here.

- `internal/action/sync.go` - `GOPASS_AUTOSYNC_INTERVAL` and others

dominikschulz:
This PR adds a new config format - again. This time it's based on git's very own "gitconfig" format. Unfortunately it's not entirely trivial and to date there seem to be no git config handlers in pure Go. All I've found were shelling out to the git binary (which does make sense, as most of the options do have default values that only the binary would know about).

But I still think this format is best suited for our use case since it does combine human and machine accessibility with the right amount of flexibility.

WARNING: This will certainly break some things! The current config handling is so deeply wired in and mixed with our (ab-)use of the `context` package that I'm unlikely to catch every possible breakage on my own. Our users will have to help testing this. Either before it's released or afterwards. But the upshot is that this has the potential to allow us to resolve at least half a dozen outstanding issues and remove quite a bit of legacy from the codebase (more so once we drop the config backwards compatability layer). We could even consider adopting the parser for the secrets themselves in the future.

- [x] Describe what this PR is doing here
- [x] and in our docs
- [x] and add some examples how it can be leveraged
- [x] Make unit tests pass
- [x] Make integration tests pass
- [x] Support read-only mounts, see https://github.com/gopasspw/gopass/issues/1878
- [x] Add autosync option, see https://github.com/gopasspw/gopass/issues/2387
- [x] Add config.worktree support
- [x] Move domain aliases to the config
- [x] Review https://github.com/gopasspw/gopass/issues/1819
- [x] Review https://github.com/gopasspw/gopass/issues/1798
- [x] Review https://github.com/gopasspw/gopass/issues/1567
- [x] Support autoclip on show, see https://github.com/gopasspw/gopass/issues/1764

Fixes #1567 
Fixes #1764 
Fixes #1819 
Fixes #1878 
Fixes #2387 
Fixes #2408 
Fixes #2418


internal/action/commands.go
```diff
@@ -981,7 +958,7 @@ func (s *Action) GetCommands() []*cli.Command {
 	for _, be := range backend.CryptoRegistry.Backends() {
 		bc, ok := be.(commander)
 		if !ok {
-			debug.Log("Backend %s does not implement commander interface\n", be)
+			// debug.Log("Backend %s does not implement commander interface\n", be)
```
AnomalRoil, line 961: Why remove this debug log? Feels like it could be useful sometimes, no?

internal/action/commands.go
```diff
@@ -993,7 +970,7 @@ func (s *Action) GetCommands() []*cli.Command {
 	for _, be := range backend.StorageRegistry.Backends() {
 		bc, ok := be.(storeCommander)
 		if !ok {
-			debug.Log("Backend %s does not implement commander interface\n", be)
+			// debug.Log("Backend %s does not implement commander interface\n", be)
```
AnomalRoil, line 973: Why remove this debug log? Feels like it could be useful sometimes, no?

internal/action/config.go
```diff
@@ -30,45 +32,31 @@ func (s *Action) Config(c *cli.Context) error {
...
 	}
 
 	return nil
 }
 
-func (s *Action) printConfigValues(ctx context.Context, needles ...string) {
-	m := s.cfg.ConfigMap()
-	for _, k := range filterMap(m, needles) {
+func (s *Action) printConfigValues(ctx context.Context, store string, needles ...string) {
+	// debug.Log("printing value of %v", needles)
```
AnomalRoil, line 43: Why leave this commented? I'd say either leave it as a debug log or remove it, no?
Overall I'd leave it as a debug log, feels it could be useful in the coming months.

internal/action/config.go
```diff
@@ -30,45 +32,31 @@ func (s *Action) Config(c *cli.Context) error {
...
+
+	for _, k := range set.SortedFiltered(s.cfg.Keys(store), func(e string) bool {
+		// debug.Log("does %+v contain %q?", needles, e)
```
AnomalRoil, line 46: Why leave this commented? I'd say either leave it as a debug log or remove it, no?
Overall I'd leave it as a debug log, feels it could be useful in the coming months.

internal/action/commands.go
```diff
@@ -68,29 +68,6 @@ func (s *Action) GetCommands() []*cli.Command {
...
-					Description: "Remove an alias from a domain",
-				},
-				{
-					Name:        "delete",
-					Action:      s.AliasesDelete,
-					Usage:       "Delete an entire domain",
-					ArgsUsage:   "[alias]",
-					Description: "Delete an entire domain",
-				},
-			},
```
AnomalRoil, line 93: Same question. Why remove it?

AnomalRoil (commented):
Okay, I haven't yet started to dig into the big changes yet. A few comments. I'll keep going later.

Makefile
```diff
@@ -110,7 +110,7 @@ test-win: $(GOPASS_OUTPUT)
 		$(GO) test -test.short -run '(Test|Example)' $(pkg) || exit 1;)
 
 test-integration: $(GOPASS_OUTPUT)
-	cd tests && GOPASS_BINARY=$(PWD)/$(GOPASS_OUTPUT) GOPASS_TEST_DIR=$(PWD)/tests $(GO) test -v
+	cd tests && GOPASS_BINARY=$(PWD)/$(GOPASS_OUTPUT) GOPASS_TEST_DIR=$(PWD)/tests $(GO) test -v $(TESTFLAGS)
```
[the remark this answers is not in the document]
dominikschulz: TESTFLAGS is already present in the Makefile since ... forever. I don't want to mess with it in this PR.

fish.completion
```diff
@@ -51,42 +51,6 @@ complete -c $PROG -f -n '__fish_gopass_uses_command age identities -l chars -d "
...
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l unsafe -d "Display unsafe content (e.g. the password) even if safecontent is enabled"'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l password -d "Display only the password. Takes precedence over all other flags."'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l revision -d "Show a past revision. Does NOT support RCS specific shortcuts. Use exact revision or -&lt;N&gt; to select the Nth oldest revision of this entry."'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l noparsing -d "Do not parse the output."'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l chars -d "Print specific characters from the secret"'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l help -d "show help"'
-complete -c $PROG -f -n '__fish_gopass_uses_command alias delete -l version -d "print the version"'
 complete -c $PROG -f -n '__fish_gopass_needs_command' -a audit -d 'Command: Decrypt all secrets and scan for weak or leaked passwords'
```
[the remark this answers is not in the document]
dominikschulz: I have moved domain aliases to the config. Previously they were in their own file. This saves us a bunch of code since we can just use the normal `gopass config` command. I think it's a nice clean up.

internal/action/aliases.go
```diff
@@ -27,59 +25,3 @@ func (s *Action) AliasesPrint(c *cli.Context) error {
...
-	}
-
-	if err := pwrules.DeleteCustomAlias(domain); err != nil {
-		return err
-	}
-
-	out.Printf(ctx, "Remove aliases for domain %q", domain)
-
-	return nil
-}
```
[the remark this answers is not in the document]
dominikschulz: Because I'm moving domain aliases to the config.
Why in this PR? It's kind of a feasibility test. Sure it might have deserved it's own PR but a this point I'm too lazy to tear it apart again.

internal/action/commands.go
dominikschulz: I think this has been discussed / answered above?!

internal/action/commands.go
dominikschulz: These were left overs. We're already way to verbose in the debug log. I'd rather remove those.

internal/action/show.go
```diff
@@ -49,7 +50,7 @@ func showParseArgs(c *cli.Context) context.Context {
 	}
 
 	if c.IsSet("noparsing") {
-		ctx = ctxutil.WithShowParsing(ctx, !c.Bool("noparsing"))
+		_ = config.FromContext(ctx).SetEnv("core.parsing", fmt.Sprintf("%t", !c.Bool("noparsing")))
```
AnomalRoil, line 53: How long is `SetEnv` persisted?
I feel like running this in the REPL might be persisted for more than just one command, no?

internal/action/show.go
```diff
@@ -255,20 +256,23 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	}
 
 	// everything but the first line.
-	if ctxutil.IsShowSafeContent(ctx) && !ctxutil.IsForce(ctx) {
-		body := showSafeContent(ctx, sec)
+	if config.Bool(ctx, "core.showsafecontent") && !ctxutil.IsForce(ctx) {
+		debug.Log("core.showsafecontent is true")
```
AnomalRoil, line 260: Is this log really meant to be added only for that case and not for all the previous ones?

internal/action/show_test.go
```diff
@@ -54,21 +62,15 @@ func TestShowMulti(t *testing.T) { //nolint:paralleltest
...
-		assert.NoError(t, act.Store.Set(ctx, "bar/baz", sec))
-		buf.Reset()
-
 		c := gptest.CliCtx(ctx, t, "bar")
 		assert.NoError(t, act.Show(c))
 		assert.Equal(t, "bar/\n└── baz\n\n", buf.String())
 		buf.Reset()
 	})
 
+	require.NoError(t, act.cfg.Set("", "core.showsafecontent", "true"))
```
AnomalRoil, line 71: Any reason to use a "global" one here instead of the explicit declaration in each test like it's done later in the same file for parsing?

internal/action/sync.go
```diff
@@ -27,6 +28,7 @@ var (
 )
 
 func init() {
+	// TODO(gitconfig) use proper config setting
```
AnomalRoil, line 31: I'm not sure to understand this todo. Wouldn't it be best to open an issue for it?

internal/action/sync.go
```diff
@@ -128,6 +136,12 @@ func (s *Action) sync(ctx context.Context, store string) error {
 
 // syncMount syncs a single mount.
 func (s *Action) syncMount(ctx context.Context, mp string) error {
+	if as := config.FromContext(ctx).GetM(mp, "core.autosync"); as == "false" {
```
AnomalRoil, line 139: Why not a `config.Bool(ctx, "core.autosync")` here?

internal/action/sync.go
```diff
@@ -181,11 +195,12 @@ func (s *Action) syncMount(ctx context.Context, mp string) error {
 	}
 	syncPrintDiff(ctxno, l, ln)
 
-	debug.Log("Syncing Mount %s. Exportkeys: %t", mp, ctxutil.IsExportKeys(ctx))
+	exportKeys := s.cfg.GetBool("core.exportkeys")
```
AnomalRoil, line 198: I see we're using both `s.cfg` and `config.FromContext` in this file, I'm not sure I grasp fully the difference, if it's on purpose or if we could remove the `cfg` from the `Action` struct since it's now in the contexts?

internal/action/sync_test.go
```diff
@@ -30,6 +30,7 @@ func TestSync(t *testing.T) { //nolint:paralleltest
 	act, err := newMock(ctx, u.StoreDir(""))
 	require.NoError(t, err)
 	require.NotNil(t, act)
+	ctx = act.cfg.WithConfig(ctx)
```
AnomalRoil, line 33: I guess this answers my above question, and so no we cannot remove the cfg from the action since we use it to populate the contexts...

pkg/ctxutil/ctxutil.go
```diff
@@ -15,12 +15,9 @@ const (
 	ctxKeyTerminal contextKey = iota
 	ctxKeyInteractive
 	ctxKeyStdin
-	ctxKeyNoPager
-	ctxKeyShowSafeContent
 	ctxKeyGitCommit
 	ctxKeyAlwaysYes
 	ctxKeyVerbose
-	ctxKeyNotifications
```
AnomalRoil, line 23: I feel like most of the rest of `ctxutil` could be replaced with `SetEnv` configs, no?

AnomalRoil (commented):
70/114 reviewed, but I'm getting to the big chunks. Here's a few initial comments and I'll go through the rest tomorrow hopefully.

internal/action/show.go
dominikschulz: Yes. This is persistet as long as the binary is running. With the REPL this might be tricky to handle properly.

internal/action/show.go
dominikschulz: Which previous ones?

internal/action/sync.go
dominikschulz: Added a comment.

internal/action/sync.go
dominikschulz: No. I think we should use s.cfg when it's available. The FromContext was a mistake.

internal/action/sync_test.go
dominikschulz: Ack.

pkg/ctxutil/ctxutil.go
dominikschulz: No. This is not possible (yet?).

So of these are used for in-process communication.
Maybe some day we can get rid of these. But it's so much work with so little user-visible value that I'm not very motivated to do it now.

internal/action/show.go
AnomalRoil: All the previous `if/else` are not logged, but just this one, so I was wondering it there was a special reason to log this?

dominikschulz: This was a stray debug statement. We could leave it, but we can also remove it. The other clauses are not new, so they don't have it.

internal/config/config.go
```diff
@@ -2,152 +2,223 @@ package config
...
-			f.SetInt(int64(iv))
-
-			return nil
-		default:
-			continue
-		}
+// GetBool returns true if the value of the key evaluates to "true".
+// Otherwise it returns false.
+func (c *Config) GetBool(key string) bool {
+	if strings.ToLower(strings.TrimSpace(c.Get(key))) == "true" {
```
AnomalRoil, line 117: Do we really need the `strings.ToLower(strings.TrimSpace` here? Why not on the `Get` side?

internal/action/sync.go
```diff
@@ -128,6 +136,14 @@ func (s *Action) sync(ctx context.Context, store string) error {
 
 // syncMount syncs a single mount.
 func (s *Action) syncMount(ctx context.Context, mp string) error {
+	// using GetM here to get the value for this mount, it might be different
+	// than the global value
+	if as := s.cfg.GetM(mp, "core.autosync"); as == "false" {
```
AnomalRoil, line 141: Note that the current `GetBool` implementation assumes that `"true"` will mean true and any other value will mean `false` whereas here is assumes `"false"`. 
A slight difference that could bite since `GetBool` hasn't the same behaviour.

internal/config/legacy.go
```diff
@@ -1,329 +1,44 @@
...
-	}
-
-	return cfg
-}
-
-func pathFromURL(u string) (string, error) {
-	if !strings.Contains(u, "://") {
-		return u, nil
+	for alias, path := range cfg.Mounts {
+		if err := c.SetGlobal(mpk(alias), path); err != nil {
```
AnomalRoil, line 36: What if different mounts have the same alias?

internal/config/legacy/legacy.go
```diff
@@ -0,0 +1,329 @@
+package legacy
+
+import (
+	"net/url"
+	"strings"
+)
+
+// Pre1127 is a pre-1.12.7 config.
```
AnomalRoil, line 8: Shouldn't we add here a "pre 1.15" test?

internal/config/config.go
dominikschulz: It's being applied to the value. We only want this here.

internal/config/legacy.go
dominikschulz: That's an invalid setup. I mean ... the mounts map is keyed by this alias. So I'd say that can't happen.

internal/config/legacy/legacy.go
dominikschulz: It's not a test. But a backwards compat handler. The latest legacy config is the pre-1.15 one.

internal/action/sync.go
dominikschulz: This is true. A bit tricky. But I still think the current implementation is OK.

I'd assume the default to be "autosync all stores" or maybe "autosync no stores".
But we should be able to support both "autosync no stores except this one" and "autosync all stores except this one".

internal/action/show_test.go
dominikschulz: No. Not really ...

AnomalRoil (commented):
Okay, 105/114 I just haven't looked at the gitconfig package itself right now.

internal/action/sync.go
AnomalRoil: An easy fix to have both match the GetBool behaviour would be:
```suggestion
	if as := s.cfg.GetM(mp, "core.autosync"); as != "true" {
```

internal/config/legacy/location.go
```diff
@@ -0,0 +1,63 @@
...
+		cleanName := strings.ReplaceAll(mount, string(filepath.Separator), "-")
+
+		return fsutil.CleanPath(filepath.Join(appdir.UserData(), "stores", cleanName))
+	}
+	// PASSWORD_STORE_DIR support is discouraged.
+	if d := os.Getenv("PASSWORD_STORE_DIR"); d != "" {
+		debug.Log("using value of PASSWORD_STORE_DIR: %s", d)
+
+		return fsutil.CleanPath(d)
+	}
```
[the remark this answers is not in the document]
dominikschulz: Good catch. I'm expecting `GOPASS_HOMEDIR` to always take precendence. Since it's mainly for testing.

internal/config/config.go
```diff
@@ -2,152 +2,223 @@ package config
...
+// NewNoWrites initializes a new config that does not allow writes. For use in tests.
+func NewNoWrites() *Config {
+	return newWithOptions(true)
 }
 
-// Config will return a current config.
-func (c *Config) Config() *Config {
+func newWithOptions(noWrites bool) *Config {
+	c := &Config{
+		cfgs: make(map[string]*gitconfig.Configs, 10),
```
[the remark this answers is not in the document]
dominikschulz: Yes, but most people won't have all of them set.

But it doesn't really matter. I guess we shouldn't use a magic constant here at all.
This will likely never play a performance relevant role - unless someone is creating tons of Config structs.

tests/gptest/gunit.go
```diff
@@ -50,26 +51,30 @@ func NewGUnitTester(t *testing.T) *GUnit {
...
-		"GNUPGHOME":          u.GPGHome(),
-		"GOPASS_CONFIG":      u.GPConfig(),
-		"GOPASS_HOMEDIR":     u.Dir,
-		"NO_COLOR":           "true",
-		"GOPASS_NO_NOTIFY":   "true",
-		"PAGER":              "",
+		"CHECKPOINT_DISABLE":       "true",
+		"GNUPGHOME":                u.GPGHome(),
+		"GOPASS_CONFIG_NOSYSTEM":   "true",
+		"GOPASS_CONFIG_NO_MIGRATE": "true",
```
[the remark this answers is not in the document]
dominikschulz: Now they do :)

tests/config_test.go
```diff
@@ -15,44 +15,40 @@ func TestBaseConfig(t *testing.T) { //nolint:paralleltest
...
-		"parsing",
+		"core.autoimport",
+		"core.showsafecontent",
+		"core.parsing",
 	}
 
 	for _, invert := range invertables { //nolint:paralleltest
 		t.Run("invert "+invert, func(t *testing.T) {
-			out, err = ts.run("config " + invert + " false")
+			out, err = ts.run("config core." + invert + " false")
```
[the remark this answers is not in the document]
dominikschulz: Good catch. They did definitely pass.

internal/store/leaf/read.go
```diff
@@ -30,7 +31,7 @@ func (s *Store) Get(ctx context.Context, name string) (gopass.Secret, error) {
 		return nil, store.ErrDecrypt
 	}
 
-	if !ctxutil.IsShowParsing(ctx) {
+	if !ctxutil.IsShowParsing(ctx) || !config.Bool(ctx, "core.parsing") {
```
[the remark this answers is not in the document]
dominikschulz: IsShowParsing should have been set in the context by the config.
That was always a bit ugly and I'm happy we can reduce that context (ab-)use a bit.

internal/action/sync.go
```diff
@@ -55,12 +58,23 @@ func (s *Action) autoSync(ctx context.Context) error {
...
 
 	ls := s.rem.LastSeen("autosync")
 	debug.Log("autosync - last seen: %s", ls)
-	if time.Since(ls) > time.Duration(autosyncIntervalDays)*24*time.Hour {
+	syncInterval := autosyncIntervalDays
+	if iv := s.cfg.GetInt("autosync.interval"); iv > 0 {
+		syncInterval = iv
+	}
+
+	if time.Since(ls) > time.Duration(syncInterval)*24*time.Hour {
```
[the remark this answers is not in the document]
AnomalRoil: BTW, do we want to support a value of 0 ?
It would mean "run sync on every run", right?

pkg/gitconfig/gitconfig.go
```diff
@@ -0,0 +1,15 @@
...
+var (
+	// SystemConfig is the location of the (optional) system-wide config defaults file.
+	systemConfig = "/etc/gitconfig" // /etc/gopass/config
+	// GlobalConfig is the location of the (optional) global (i.e. user-wide) config file.
+	globalConfig = ".gitconfig"
+	// LocalConfig is the name of the local (per-workdir) configuration.
+	localConfig = "config"
+	// WorktreeConfig is the name of the local worktree configuration. Can be used to override
+	// a committed local config.
+	worktreeConfig = "config.worktree"
```
AnomalRoil, line 12: I think that worktree configs are even worse, since their config is stored in the _original repo_'s `.git` folder under `.git/worktrees/nameofwt/config.worktree`

e.g. if I have a worktree named "myown" located at `/home/anomalroil/code/myown` for gopass it would look like:
```
file:/home/anomalroil/.gitconfig        commit.gpgsign=true
file:/home/anomalroil/code/gopass/.git/config    user.email=anomalroil@users.noreply.github.com
file:/home/anomalroil/code/gopass/.git/worktrees/myown/config.worktree   user.name=TEST
``` 
because in a worktree, the folder `.git` is replaced with a file `.git` specifying the location of its original gitdir:
```
$ cat .git
gitdir: /home/anomalroil/code/drand/.git/worktrees/myown
```
Is this currently taken into account?

main.go
```diff
@@ -290,6 +290,10 @@ func initContext(ctx context.Context, cfg *config.Config) context.Context {
 	// disable this for all terms on this platform
 	if runtime.GOOS == "windows" {
 		color.NoColor = true
+	} else {
+		if cfg.Get("core.nocolor") != "" {
+			color.NoColor = cfg.GetBool("core.nocolor")
+		}
```
AnomalRoil, line 296:
```suggestion
		color.NoColor = cfg.GetBool("core.nocolor")
```
If I understand it correctly, the

internal/action/sync.go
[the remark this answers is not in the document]
dominikschulz: I think it's acceptable for this to overflow after 718 years.

Shouldn't 0 already be supported by this?

internal/config/docs_test.go
```diff
@@ -0,0 +1,137 @@
...
+	// t.Logf("Used: %+v", used)
+
+	for k := range documented {
+		if !used[k] {
+			t.Errorf("Documend but not used: %s", k)
+		}
+	}
+	for k := range used {
+		if !documented[k] {
+			t.Errorf("Used but not documented: %s", k)
```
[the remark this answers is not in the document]
dominikschulz: 😄

pkg/gitconfig/gitconfig.go
```diff
@@ -0,0 +1,15 @@
+package gitconfig
+
+var (
+	// SystemConfig is the location of the (optional) system-wide config defaults file.
+	systemConfig = "/etc/gitconfig" // /etc/gopass/config
+	// GlobalConfig is the location of the (optional) global (i.e. user-wide) config file.
+	globalConfig = ".gitconfig"
+	// LocalConfig is the name of the local (per-workdir) configuration.
+	localConfig = "config"
```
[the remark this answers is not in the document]
dominikschulz: Yes, that's how this is used in e.g. the gitfs backend.

pkg/gitconfig/gitconfig_test.go
```diff
@@ -0,0 +1,424 @@
...
+
+	c := ParseConfig(strings.NewReader(configSampleDocs))
+
+	for k, v := range c.vars {
+		t.Logf("%s => %s\n", k, v)
+	}
+}
+
+func TestGitBinary(t *testing.T) { //nolint:paralleltest
+	t.Skip("not ready, yet") // TODO(gitconfig) make tests pass
```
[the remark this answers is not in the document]
dominikschulz: I did not have time to finish this, yet. But I don't consider it blocking for this PR.
The main purpose is to have this parser for gopass, I'm not aming a full git compatability, yet.

AnomalRoil (commented):
Last one for today, I hope to wrap it up tomorrow!

internal/action/sync.go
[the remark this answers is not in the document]
AnomalRoil: I think it's not, we're only setting the variable if `iv > 0`:
```
	if iv := s.cfg.GetInt("autosync.interval"); iv > 0 {
		syncInterval = iv
	}
```

pkg/gitconfig/gitconfig.go
dominikschulz: No, it isn't. And I wasn't aware of that. Thank you. I don't necessarily plan to support this as part of this PR, but it's definitely something that needs to be documented and should have a TODO. Will add that and then resolve this comment.

main.go
dominikschulz: Right now I'm not sure. I had a reason to add this.

But definitely two TODOs for me here:
* document why we're checking if it's set, if we decide to keep it
* add an IsSet method to the config to make clear what we're doing (better readable than comparing to an empty string, which is likely not obvious)

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,363 @@
...
+	if !present {
+		debug.Log("inserting value")
+
+		return c.insertValue(key, value)
+	}
+
+	debug.Log("updating value")
+
+	return c.rewriteRaw(key, value, func(fKey, sKey, value, comment string) (string, bool) {
+		return fmt.Sprintf("    %s = %s%s", sKey, value, comment), false
```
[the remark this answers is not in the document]
dominikschulz: Yes, this might have been a literal tab and got replaced by auto-format.

I'll add an escaped tab (\t) later.

internal/config/docs_test.go
```diff
@@ -0,0 +1,137 @@
...
+		}
+	}
+}
+
+func usedOpts(t *testing.T) map[string]bool {
+	t.Helper()
+
+	optRE := regexp.MustCompile(`(?:\.Get(?:|Int|Bool)\(\"([a-z]+\.[a-z]+)\"\)|\.GetM\([^,]+, \"([a-z]+\.[a-z]+)\"\)|config\.Bool\(ctx, \"([a-z]+\.[a-z]+)\"\))`)
```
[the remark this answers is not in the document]
dominikschulz: Sorry, I don't follow what you're asking for here. Could you please elaborate?

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,363 @@
...
+			p := strings.Fields(line)
+			if len(p) < 1 {
+				continue
+			}
+			section = p[0]
+			subsection = ""
+			if len(p) > 1 {
+				subsection = strings.ReplaceAll(p[1], "\\", "")
+				subsection = strings.TrimPrefix(subsection, "\"")
+				subsection = strings.TrimSuffix(subsection, "\"")
```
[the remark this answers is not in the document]
dominikschulz: Correct. This is not covering all of the defined cases, yet.

Not sure I can / want to adress this in this PR, but I will:

* add it to the list of known issues / limitations
* add documentation to this method to help reason about it

For `len(p) = 3`: This shouldn't happen.

These formats are supported:
```
[section] -> brackets get trimmed, section is in p[0]
[section "subsection"] -> brackets get trimmed, section is in p[0], subsection in p[1]
```

Everything else is invalid, I think.

dominikschulz:
@AnomalRoil Thanks a lot. I'll probably not get to address the open comments until tomorrow anyway. If you could leave any remaining comments today that'd be fantastic.

Maybe we can wrap this up until the end of the week :smiley:

pkg/gitconfig/config.go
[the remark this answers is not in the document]
AnomalRoil: So
```
[section "name of my subsection with spaces"] ->  trimming of p[1] only, so `subsection = name`
``` 
correct?

internal/config/docs_test.go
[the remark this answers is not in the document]
AnomalRoil: Should we also parse the code to detect all usages of `os.GetEnv(nameofenvvar)` to get a list of all names of all env variables we're using at some point, and make sure they are documented?

pkg/gitconfig/config.go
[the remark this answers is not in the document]
dominikschulz: Ah, sorry. Now I get it. This is indeed something that I should fix before merging.

internal/config/docs_test.go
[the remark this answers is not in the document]
dominikschulz: Got it. Great idea. Not directly related to the gitconfig change, but I'm OK with adding it to the PR.

pkg/gitconfig/gitconfig.go
dominikschulz: Reading https://git-scm.com/docs/git-config#SCOPES I'm not sure if that's correct - but I haven't used this feature before so I might be wrong.

main.go
dominikschulz: There is no default, so I still want to check if it's set.

docs/config.md
```diff
@@ -42,31 +50,48 @@ Variables not exclusively used by gopass
...
+| `core.nocolor`         | `bool`   | Do not use color. | `false` |
+| `core.nopager`         | `bool`   | Do not invoke a pager to display long lists. | `false` |
+| `core.notifications`   | `bool`   | Enable desktop notifications. | `true` |
+| `core.parsing`         | `bool`   | Enable parsing of output to have key-value and yaml secrets. | `true` |
+| `core.readonly`        | `bool`   | Disable writing to a store. Note: This is just a convenience option to prevent accidential writes. Enforcement can only happen on a central server (if repos are set up around a central one). | `false` |
+| `mounts.path`          | `string` | Path to the root store. | `$XDG_DATA_HOME/gopass/stores/root` |
+| `core.showsafecontent` | `bool`   | Only output *safe content* (i.e. everything but the first line of a secret) to the terminal. Use *copy* (`-c`) to retrieve the password in the clipboard, or *force* (`-f`) to still print it. | `false` |
+| `age.usekeychain`      | `bool`   | Use the OS keychain to cache age passphrases. | `false` |
+| `domain-alias.<from>`   | `string` | Alias from domain to the string value of this entry. | `` |
```
AnomalRoil, line 95:
```suggestion
| `domain-alias.<from>`   | `string` | Alias from domain to the string value of this entry. | `""` |
```
Empty code doesn't work in github markdown, so I guess we can stick to the classical way of representing the empty string

internal/action/sync.go
```diff
@@ -32,6 +33,8 @@ func init() {
 		return
 	}
 
+	debug.Log("GOPASS_AUTOSYNC_INTERVAL is deprecated. Please use autosync.interval")
```
AnomalRoil, line 36: I guess this is a silent `debug.Log` instead of a noisy `out.Warning` because of the lack of context in `init`?

internal/action/sync.go
```diff
@@ -181,11 +204,12 @@ func (s *Action) syncMount(ctx context.Context, mp string) error {
 	}
 	syncPrintDiff(ctxno, l, ln)
 
-	debug.Log("Syncing Mount %s. Exportkeys: %t", mp, ctxutil.IsExportKeys(ctx))
+	exportKeys := s.cfg.GetBool("core.exportkeys")
```
AnomalRoil, line 207: Shouldn't this also be a `s.cfg.GetM` since it could be mount-specific `exportkeys` setting?

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+	return set.SortedFiltered(set.Apply(c.Keys(), func(k string) string {
+		section, subsection, _ := splitKey(k)
+		if section != wantSection {
+			return ""
+		}
+
+		return subsection
+	}), func(s string) bool {
+		return s != ""
+	})
```
AnomalRoil, line 357: I guess this would not come up as a prime example of idiomatic Go xD 
I'll blame generics, but fair enough, it works too.

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,379 @@
...
+
+	return c
+}
+
+// LoadConfigFromEnv will try to parse an overlay config from the environment variables.
+// If no environment variables are set the resulting config will be valid but empty.
+// Either way it will not be writeable.
+func LoadConfigFromEnv(envPrefix string) *Config {
+	c := &Config{
+		noWrites: true,
```
AnomalRoil, line 351: I feel like this is missing a `vars: make(map[string]string, 42)`

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+			readonly: true,
+		},
+		global: &Config{
+			path: globalConfigFile(),
+		},
+		local:    &Config{},
+		worktree: &Config{},
+		env: &Config{
+			noWrites: true,
+		},
```
AnomalRoil, line 46: Shouldn't all of these initialize the `vars` map?

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,379 @@
...
+	// can't set env vars
+	if c.readonly {
+		debug.Log("can not write to a readonly config")
+
+		return nil
+	}
+
+	if c.vars == nil {
+		c.vars = make(map[string]string, 16)
+	}
```
AnomalRoil, line 69: This feels like a patch of a symptom of another deeper bug around `c.vars` initialization, no?
Also note that Parse below is initializing with 42 instead of 16. Maybe we should strive to be consistent or have a default magic number as a const?

pkg/gitconfig/config.go
AnomalRoil:
```suggestion
		noWrites: true,
		vars: make(map[string]string, 42),
```

pkg/gitconfig/configs.go
AnomalRoil:
```suggestion
		system: &Config{
			readonly: true,
			vars: make(map[string]string, 42),
		},
		global: &Config{
			path: globalConfigFile(),
			vars: make(map[string]string, 42),
		},
		local:    &Config{
			vars: make(map[string]string, 42),
		},
		worktree: &Config{
			vars: make(map[string]string, 42),
		},
		env: &Config{
			noWrites: true,
			vars: make(map[string]string, 42),
		},
```

pkg/gitconfig/config.go
AnomalRoil: Hopefully my proposed changes above should patch this...

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,379 @@
...
+	for i := 0; i < count; i++ {
+		keyVar := fmt.Sprintf("%s%d", envPrefix+"_CONFIG_KEY_", i)
+		key := os.Getenv(keyVar)
+
+		valVar := fmt.Sprintf("%s%d", envPrefix+"_CONFIG_VALUE_", i)
+		value, found := os.LookupEnv(valVar)
+
+		if key == "" || !found {
+			return &Config{
+				noWrites: true,
```
AnomalRoil, line 370:
```suggestion
				noWrites: true,
				vars: make(map[string]string, count),
```

pkg/gitconfig/config.go
```diff
@@ -0,0 +1,379 @@
...
+// Either way it will not be writeable.
+func LoadConfigFromEnv(envPrefix string) *Config {
+	c := &Config{
+		noWrites: true,
+	}
+
+	count, err := strconv.Atoi(os.Getenv(envPrefix + "_CONFIG_COUNT"))
+	if err != nil || count < 1 {
+		return &Config{
+			noWrites: true,
```
AnomalRoil, line 357:
```suggestion
			noWrites: true,
			vars: make(map[string]string),
```

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// load the system config, if any
+	if os.Getenv(cs.EnvPrefix+"_NOSYSTEM") == "" {
+		c, err := LoadConfig(cs.SystemConfig)
+		if err != nil {
+			debug.Log("failed to load system config: %s", err)
+		} else {
+			debug.Log("loaded system config from %s", cs.SystemConfig)
+			cs.system = c
+			cs.system.readonly = true
```
AnomalRoil, line 77: Not sure why the system config is `readonly`?

pkg/gitconfig/config.go
AnomalRoil: I guess actually we can also just keep it `nil` until we need to write to it, true. But then we need to guard against accessing it when it's nil

pkg/gitconfig/config.go
AnomalRoil: Actually I realized afterwards you were doing things properly in Set and Get to avoid any issue with this, so feel free to disregard all of these.

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+		c.local,
+		c.worktree,
+		c.env,
+	} {
+		if cfg == nil {
+			continue
+		}
+		if cfg.vars == nil {
+			continue
+		}
```
AnomalRoil, line 316:
```suggestion
		if cfg == nil || cfg.vars == nil {
			continue
		}
```

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// load any env vars
+	cs.env = LoadConfigFromEnv(cs.EnvPrefix)
+
+	return cs
+}
+
+func globalConfigFile() string {
+	// $XDG_CONFIG_HOME/git/config
+	return filepath.Join(appdir.UserConfig(), "config")
```
AnomalRoil, line 123: `appdir.UserConfig()` is currently returning `$XDG_CONFIG_HOME/gopass/config` or `$HOME/.config/gopass/config` :
https://github.com/gopasspw/gopass/blob/5981a5a643b682244b100c15a9aa48fb17a04621/pkg/appdir/appdir_xdg.go#L11-L23
and
https://github.com/gopasspw/gopass/blob/5981a5a643b682244b100c15a9aa48fb17a04621/pkg/appdir/appdir.go#L12-L13

So this here is actually returning `$XDG_CONFIG_HOME/gopass/config/config`, right?
I don't feel like it's what we want.

pkg/gitconfig/configs.go
```diff
@@ -0,0 +1,358 @@
...
+	return false
+}
+
+// SetLocal sets (or adds) a key only in the per-directory (local) config.
+func (c *Configs) SetLocal(key, value string) error {
+	if c.local == nil {
+		if c.workdir == "" {
+			return fmt.Errorf("no workdir set")
+		}
+		c.local = &Config{}
```
AnomalRoil, line 248: Shouldn't this set the path of the local config?

pkg/gitconfig/configs.go
dominikschulz: Yeah, it could be rewritten to be a bit simpler ...

pkg/gitconfig/config.go
dominikschulz: Ack.

pkg/gitconfig/config.go
dominikschulz: I think we do that.

pkg/gitconfig/configs.go
dominikschulz: Add some comment.

pkg/gitconfig/configs.go
dominikschulz: This seems correct to me. We used to have the old yaml config in `$XDG_CONFIG_HOME/gopass/config/config.yml` and I just wanted to drop the extension.

pkg/gitconfig/configs.go
dominikschulz: Absolutely. Thank you!

pkg/gitconfig/configs.go
dominikschulz: Actually we don't even have to check for `cfg.vars == nil`. range handles that properly.

AnomalRoil (dismissed):
LGTM!

Maybe just 2 last changes!

Correct the doc in:
https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/docs/commands/config.md

And add the store flag to the `config` command:
https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/internal/action/commands.go#L157-L168

AnomalRoil:
Ah, it seems a few extra documentation places needs updates too:
- https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/docs/usecases/readyonly-store.md (also the name of that file seem wrong, why "ready" and not "read"?)
- https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/docs/setup.md
- https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/docs/config.md (still contains mention of "autoclip")
- https://github.com/gopasspw/gopass/blob/dcea35c0c97ead9529da76797eba24d3ed08bdab/docs/commands/create.md (The example uses `$(gopass config path)`)

dominikschulz:
@AnomalRoil Thanks again. Updated all of these.

AnomalRoil:
The following didn't work for me:
```
gopass config core.safecontent true
echo """this is a test
on two line""" | gopass insert test
gopass show test
```
It displayed the two lines instead of hiding the first line.

internal/action/show.go
```diff
@@ -182,8 +183,8 @@ func (s *Action) showHandleOutput(ctx context.Context, name string, sec gopass.S
 	}
 
 	if pw == "" && body == "" {
-		if ctxutil.IsShowSafeContent(ctx) && !ctxutil.IsForce(ctx) {
-			out.Warning(ctx, "safecontent=true. Use -f to display password, if any")
+		if config.Bool(ctx, "core.showsafecontent") && !ctxutil.IsForce(ctx) {
+			out.Warning(ctx, "core.showsafecontent=true. Use -f to display password, if any")
```
AnomalRoil, line 187:
```suggestion
		if config.Bool(ctx, "core.safecontent") && !ctxutil.IsForce(ctx) {
			out.Warning(ctx, "core.safecontent=true. Use -f to display password, if any")
```

Seems the name didn't match that of the config code?

internal/action/show.go
```diff
@@ -195,8 +196,8 @@ func (s *Action) showHandleOutput(ctx context.Context, name string, sec gopass.S
 		}
 	}
 
-	if IsClip(ctx) && pw != "" {
-		if err := clipboard.CopyTo(ctx, name, []byte(pw), s.cfg.ClipTimeout); err != nil {
+	if (IsClip(ctx) || config.Bool(ctx, "core.showautoclip")) && pw != "" {
```
AnomalRoil, line 199: Same here, the config is `core.autoclip` but we're using `showautoclip` instead
```suggestion
	if (IsClip(ctx) || config.Bool(ctx, "core.autoclip")) && pw != "" {
```

internal/action/show.go
```diff
@@ -255,8 +256,8 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	}
 
 	// everything but the first line.
-	if ctxutil.IsShowSafeContent(ctx) && !ctxutil.IsForce(ctx) {
-		body := showSafeContent(ctx, sec)
+	if config.Bool(ctx, "core.showsafecontent") && !ctxutil.IsForce(ctx) {
```
AnomalRoil, line 259:
```suggestion
	if config.Bool(ctx, "core.safecontent") && !ctxutil.IsForce(ctx) {
```

AnomalRoil (changes requested):

AnomalRoil:
Ah, it seems the problem is actually in the migration path, since in the old `config.yml` file, `showsafecontent` was named `safecontent` only

internal/action/show.go
dominikschulz: We should be using `core.showsafecontent` everywhere.

internal/action/show.go
AnomalRoil: Actually it seems the name change is on purpose, but the migration path doesn't migrate correctly?

internal/action/show.go
AnomalRoil: Actually it seems the name change is on purpose, but the migration path doesn't migrate correctly?

internal/action/show.go
dominikschulz: This is actually intentional. We have `core.autoclip` for generate and `core.showautoclip` for gopass show.

docs/usecases/readonly-store.md
```diff
@@ -17,16 +17,16 @@ To workaround this, we can configure "Collaborators & teams" on GitHub side to g
 Ultimately it turns out that this scenario requires a feature such as a store in readonly mode, where people can configure their local store or a set of sub-stores in readonly mode, to disable the writes and the autosync-on-writes to the store, but they can still pull to sync the latest changes from the remote store. This is not a one-stop solution for the RBAC model of the team sharing store, because we still need GitHub to setup the store access at server-side, but it will provide better usage experience from gopass client side.
 
 Configuration examples:
-```console
+```bash
 # To print the config
-$ gopass config readonly
-readonly: false
+$ gopass config core.readonly
+rcore.eadonly: false
```
AnomalRoil, line 23:
```suggestion
core.readonly: false
```

AnomalRoil: Notice that this doesn't work on my machine:
``` 
$ gopass config core.readonly 
$
```
It displays nothing instead of false.

docs/config.md
AnomalRoil: This still holds I guess ^

AnomalRoil:
I wonder how the migration path would be for old config files that still contain legacy options in their YAML:
```
check_recipient_hash
concurrency
```
and others

dominikschulz:
These old configs will just be caried over with the `core.` prefix. But they will not be used by anything.

PR #2422, opened 2022-11-21
Makefile: install completion files with mode 0644

FollieHiyuki:
These files are sourced by the corresponding shells, so they don't need to be executable.


dominikschulz (approved):
LGTM

Issue #2400, opened 2022-10-29
MacOS Ventura: Failed to remove tempfile 


secwang:
### Summary

Failed to remove tempfile at /var/folders/37/3fy66clj37d268n32wsfm8400000gn/T/gopass-edit3522339400/secret: failed to unmount /dev/disk5 from /var/folders/37/3fy66clj37d268n32wsfm8400000gn/T/gopass-edit3522339400: exit status 1
### Steps To Reproduce

gopass edit
### Expected behavior

exit 0
### Environment

- OS: [e.g. Mac OS X High Sierra, Ubuntu 18.04, Windows 10, ...]  Mac os x ventura
- OS version: [uname -a] Darwin xxxx 22.1.0 Darwin Kernel Version 22.1.0: Sun Oct  9 20:14:54 PDT 2022; root:xnu-8792.41.9~2/RELEASE_X86_64 x86_64
- gopass Version: [gopass version] gopass 1.14.9 go1.19.1 darwin amd64
- Installation method: [e.g. from source, brew, gopass repo] brew

secwang:
is sandbox error?

dominikschulz:
I haven't tried Ventura, yet. Maybe Apple introduced some new "improvements"?

davenaylor:
I upgraded to Ventura this evening and am experiencing this very same issue.

fw8:
I had the same behavior prior to Ventura... and unfortunately upgrading to Ventura doesn't make the error go away.

lionello:
Same error is shown for me after Mac OS update, although `gopass edit` did appear to have worked.

dominikschulz:
@fw8 This is very annoying, but the error reported should not prevent anyone from editing or creating secrets.
If that's the case, and especially if it did happen prior to Ventura, please do open a separate issue with the GOPASS_DEBUG_LOG output. This issue will be used to track the Ventura issue.

opax:
I have the same issue with Ventura, gopass 1.14.10 go1.19.3 darwin arm64.

Executing with debug output looks like this:

```
GOPASS_DEBUG_LOG=1 gopass edit test
Initialized /dev/rdisk5 as a 16 MB case-insensitive HFS Plus volume
Unmount of all volumes on disk5 was successful
Unmount of all volumes on disk5 was successful
Unmount of all volumes on disk5 was successful
Unmount of all volumes on disk5 was successful
Unmount of all volumes on disk5 was successful
[...]

PR #2423, opened 2022-11-21
Unmount tempfile when still mounted

brianjkinney:
Fixes: [#2400](https://github.com/gopasspw/gopass/issues/2400)

RELEASE_NOTES=[BUGFIX] Fix tempfile unmount on macOS Ventura

Seeing an issue where the filesystem is still mounted after the unmountDisk command, which causes the eject to fail. This modification will check if the disk is still in the mount command output to determine if the device is still mounted, and if it is it will run the umount command.


dominikschulz:
Thanks for the help. Should be obsolete by now.

PR #2463, opened 2022-12-11
test: use `T.Setenv` to set env vars in tests

Juneezee:
A testing cleanup.

This PR replaces `os.Setenv` with `t.Setenv`. Starting from Go 1.17, we can use `t.Setenv` to set environment variable in test. The environment variable is automatically restored to its original value when the test and all its subtests complete. This ensures that each test does not start with leftover environment variables from previous completed tests.

Reference: https://pkg.go.dev/testing#T.Setenv

```go
func TestFoo(t *testing.T) {
	// before
	key := "ENV"
	originalEnv := os.Getenv(key)

	assert.NoError(t, os.Setenv(key, "new value"))
	defer assert.NoError(t, os.Setenv(key, originalEnv))
	
	// after
	t.Setenv(key, "new value")
}
```

### DIFF
--- a/internal/store/leaf/store_test.go
+++ b/internal/store/leaf/store_test.go
@@ -17,7 +17,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func createSubStore(dir string) (*Store, error) {
+func createSubStore(t *testing.T) (*Store, error) {
+	t.Helper()
+
+	dir := t.TempDir()
 	sd := filepath.Join(dir, "sub")
 
 	_, _, err := createStore(sd, nil, nil)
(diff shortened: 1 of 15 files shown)


internal/store/leaf/store_test.go
```diff
@@ -17,43 +17,28 @@ import (
...
 	}
 
-	if err := os.Setenv("GOPASS_CONFIG", filepath.Join(dir, ".gopass.yml")); err != nil {
-		return nil, err
-	}
-
-	if err := os.Setenv("GOPASS_HOMEDIR", dir); err != nil {
-		return nil, err
-	}
+	t.Setenv("GOPASS_CONFIG", filepath.Join(dir, ".gopass.yml"))
```
dominikschulz, line 31: Unrelated to this PR but the configs don't use yaml anymore for a while.
I'll need to change that.

dominikschulz (commented):
Very nice PR. Thanks a lot.
These cleanups are incredibly helpful.

Let's wait until the other PR is in to see if all tests pass after that.

Juneezee:
Rebased.

dominikschulz (approved):
Amazing work. Thank you :star_struck:

PR #2467, opened 2022-12-11
Add zsh files completion to otp command

Hazegard:
The `gopass otp`  command does not complete files when using the zsh completion.

This PR intends to add the files completion to the `otp`  command.

Old OTP zsh completion
image

Fixed OTP zsh completion:
image


dominikschulz (approved):
LGTM. Thanks!

PR #2486, opened 2022-12-23
Disable possible git concurrency during fsck

### LINKED ISSUE #2459: Git errors during gopass fsck --decrypt
### Summary
When updating recipients and running a gopass fsck --decrypt, I had some unexpected git errors. In particular,
```
] 524 / 620 [Goooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooopass                         ]  84.52% 
[] Re-encrypting some/really/secret to fix recipients and storage format.                                                                                            
❌ Task failed: failed to commit changes to git: exit status 128: fatal: Unable to create '/home/user/.password-store/.git/index.lock': File exists.
                                                                                                                                                                                       
Another git process seems to be running in this repository, e.g.                                                                                                                       
an editor opened by 'git commit'. Please make sure all processes                                                                                                                       
are terminated then try again. If it still fails, a git process                                                                                                                        
may have crashed in this repository earlier:                                                                                                                                           
remove the file manually to continue. 
```

My initial response was to both check for this file, and also other running git processes, but the file did not exist, and neither was I doing anything else in the password store. I have since tested this some 5-6 more times, and I have once successfully run a whole fsck --decrypt, but the other ~5 times I've had this exact same error occur again, but on some other secret every time.

I suspect what is happening here is that somehow, because gopass spawns so many git "edit,commit" cycles very quickly, that sometimes before one commit can clean up the index.lock file the next commit is already ready to set up the index.lock file. I don't think gopass does any sort of parallel things here, so it may well be that either git, or the file system driver, does some sort of quick return without waiting for the file to actually be deleted.

I don't know if this is true, and I also don't know how to reliably reproduce, since before today I've not seen this issue.

### Steps To Reproduce
1. I really don't know. Maybe have many secrets?
2. Run gopass fsck --decrypt

### Expected behavior
Everything works

### Environment

- OS: Gentoo 
- OS version: 5.15.74
- gopass Version: 1.14.4
- Installation method: portage

dominikschulz:
By using the default queue some git operations might still have been running when gopass did attempt to update exported keys. Disabling the queue in the inner loop makes more sense since it's more predictable and less brittle and using the queue there wouldn't help much anyway.

Fixes #2459

RELEASE_NOTES=[BUGFIX] Fix possible concurrency issues in fsck.

### DIFF
--- a/internal/queue/background.go
+++ b/internal/queue/background.go
@@ -29,7 +29,8 @@ type Queuer interface {
 	Idle(time.Duration) error
 }
 
-// WithQueue adds the given queue to the context.
+// WithQueue adds the given queue to the context. Add a nil
+// queue to disable queuing in this context.
 func WithQueue(ctx context.Context, q *Queue) context.Context {
 	return context.WithValue(ctx, ctxKeyQueue, q)
 }
@@ -37,7 +38,7 @@ func WithQueue(ctx context.Context, q *Queue) context.Context {
 // GetQueue returns an existing queue from the context or
 // returns a noop one.
 func GetQueue(ctx context.Context) Queuer {
-	if q, ok := ctx.Value(ctxKeyQueue).(*Queue); ok {
+	if q, ok := ctx.Value(ctxKeyQueue).(*Queue); ok && q != nil {
 		return q
 	}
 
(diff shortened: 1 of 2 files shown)


AnomalRoil (approved):
LGTM

PR #2492, opened 2022-12-25
Improve fsck handling and output

### LINKED ISSUE #2494: fsck --decrypt re-encrypts all the files even those with no need to touch
### Summary
fsck --decrypt reencrypts all the files even those with no need to touch.
I mentioned that allready in #2459 and was told to create a separate issue.

### Steps To Reproduce
gopass fsck --decrypt 

### Expected behavior
should only re-encrypt files where the key changed or such but should ignore files without the need for re-encryption.

### Environment
- OS: fedora 37
- OS version: 6.0.14-300.fc37.x86_64
- gopass Version: master
- Installation method: go install from master used the latest package before
- BTRFS

niacdoial:
- better use of git
- better use of error messages
- smarter behavior around importing public keys from gopass to the user's gpg keyring

Fixes #2494


dominikschulz (commented):
This looks interesting. Thank you.

Just some initial feedback for now. Expect more tomorrow.

- What is the purpose / context of this PR? Please try to provide a little bit more explanation what you're trying to do and why, e.g. what is "QOL"?
- Why did you modify the Queue funcs to return a context?
- What is the error collector doing?
- if else is not very idiomatic Go. If possible we'd try to handle the error case (the one that's less likely) in an if clause and the then continue outside of the if clause. In most cases here that's possible since the if error case would already return or continue.
- Go does not use variables with underscores inside (e.g. err_isfatal).
- Returning strings ("F", "NF") to indicate different severities is non-idiomatic as well. We might have special error types for these or maybe a bool.

niacdoial:
hi,
thank you for reviewing this!

- "QOL" is "Quality of life", so not really new features or bugfixes, just removing some friction from things.
  Basically, there are two ideas:
  - make more messages happen at the end of the fsck process, instead of having them interrupt the progress bar (which then needs to be re-drawn, taking more console space, and making everything harder to read)
  - make sure a single call to fsck results in a single git commit, which tells which secrets were re-written. useful when manual intervention is done in the keyring's repo
- the ability to pass around contexts in the queue allows to send info from one task to the next (although, that was useful when fsck had a non-nil queue, as the context object had its own error collector). This goes well with tasks effectively having two contexts, one from the queue and one from the add-task-to-queue caller.
- the error collector helps delay error messages, so they could be displayed at once not interrupting the progress bar at each message

As for the non-go-idiomatic stuff, I'll see what I can do.

niacdoial:
made it more idiomatic, but it seems that the error message for a non `--decrypt` fsck run is really misleading, so I'll need to fix that (or more likely its root cause) tomorrow.

internal/action/delete.go
```diff
@@ -104,7 +107,11 @@ func (s *Action) deleteKeyFromYAML(ctx context.Context, name, key string) error
 	sec.Del(key)
 
 	if err := s.Store.Set(ctxutil.WithCommitMessage(ctx, "Updated Key"), name, sec); err != nil {
-		return exit.Error(exit.IO, err, "Can not delete key %q from %q: %s", key, name, err)
+		if errors.Is(err, store.ErrMeaninglessWrite) {
```
dominikschulz, line 110: A more idiomatic approach would be to handle the returned error first and then the warning, like this:

```
if !errors.Is(err, store.ErrMeaninglessWrite) {
  return ...
}
out.Warningf(...)
```

Both are in the error handling branch of Store.Set anyway and any error but MeaninglessWrite should be returned.

This comment applies to this place and any other instance of this pattern.

internal/action/generate.go
```diff
@@ -354,6 +356,9 @@ func (s *Action) generateSetPassword(ctx context.Context, name, key, password st
 		setMetadata(sec, kvps)
 		_ = sec.Set(key, password)
 		if err := s.Store.Set(ctxutil.WithCommitMessage(ctx, "Generated password for key"), name, sec); err != nil {
+			if errors.Is(err, store.ErrMeaninglessWrite) {
```
dominikschulz, line 359: Do you really want to return with an error if it's an ErrMeaninglessWrite? I don't think so.

internal/backend/storage/fs/store.go
```diff
@@ -60,6 +62,12 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
 	}
 	debug.Log("Writing %s to %s", name, filepath.Join(s.path, name))
 
+	oldvalue, err := os.ReadFile(filepath.Join(s.path, name))
+	//out.Warningf(ctx, "%h\n%h\n%w", oldvalue, value, err)
```
dominikschulz, line 66: Please remove this old debug statement.

internal/backend/storage/fs/store.go
```diff
@@ -60,6 +62,12 @@ func (s *Store) Set(ctx context.Context, name string, value []byte) error {
 	}
 	debug.Log("Writing %s to %s", name, filepath.Join(s.path, name))
 
+	oldvalue, err := os.ReadFile(filepath.Join(s.path, name))
```
dominikschulz, line 65: This is at the core of what this PR is about. I think this block deserves a small comment explaining why we're doing this.

internal/queue/background.go
```diff
@@ -84,9 +84,13 @@ func New(ctx context.Context) *Queue {
 
 func (q *Queue) run(ctx context.Context) {
 	for t := range q.work {
-		if err := t(ctx); err != nil {
+		ctx2,err := t(ctx)
```
dominikschulz, line 87: These files don't seem to be properly formatted.

internal/queue/background.go
```diff
@@ -84,9 +84,13 @@ func New(ctx context.Context) *Queue {
...
+		if err != nil {
 			out.Errorf(ctx, "Task failed: %s", err)
 		}
+		if ctx2 != nil {
```
dominikschulz, line 91: This should also have a short explanation.

internal/store/err.go
```diff
@@ -23,6 +23,8 @@ var (
 	ErrGitNothingToCommit = fmt.Errorf("git has nothing to commit")
 	// ErrEmptySecret is returned if a secret exists but has no content.
 	ErrEmptySecret = fmt.Errorf("empty secret. see https://go.gopass.pw/faq#empty-secret")
+	// ErrMeaninglessWrite is returned if a secret is overwritten with its current (cyphertext) content
```
dominikschulz, line 26: For the sake of consistency, please change `cyphertext` to `ciphertext`. I think that's how we usually call it in the project.

internal/store/leaf/crypto.go
```diff
@@ -56,10 +59,19 @@ func (s *Store) ImportMissingPublicKeys(ctx context.Context, newrs ...string) er
 			debug.Log("Failed to get public key for %s: %s", r, err)
 		}
 
-		if !IsPubkeyUpdate(ctx) && len(kl) > 0 {
```
dominikschulz, line 59: This does change the semantic a bit.
Before we would skip the entire loop if we weren't asked to update keys.

But now only if there are any valid keys for this recipient.
Is that really what we want?

internal/store/leaf/crypto.go
```diff
@@ -162,14 +175,26 @@ func (s *Store) importPublicKey(ctx context.Context, r string) error {
...
 
-		return im.ImportPublicKey(ctx, pk)
+		return nil
 	}
 
-	return fmt.Errorf("public key not found in store")
+	pk, err := s.getPublicKey(ctx, r)
+	if err != nil {
+		return err
+	} else {
```
dominikschulz, line 195: You should just drop the else clause. If the if clause already returned the error then that case won't get there.

Example:

```
if err != nil {
  return err
}
return im.ImportPublicKey(ctx,pk)
```

dominikschulz (changes requested):

internal/action/generate.go
niacdoial: yeah, nil is probably more well-suited. anyway, the rest of the function is not needed is the ciphertext secret is unchanged

internal/queue/background.go
niacdoial: I believe this is now fixed, but I'll let you be the judge of that

internal/store/leaf/crypto.go
niacdoial: the intended change is the following (in pseudocode):
going from:
`if validKeysExist && !wantKeysUpdate {continue}`
to:
`if validKeysExist && (!wantKeysUpdate || bytes.Equals(pubkeyInStore, pubkeyInGPG) ) {continue}`
and the "A or B" part of this condition is moved into the inside of that if block, which only the initial `if validKeysExist` in if statement of that if block.

now that I think about it, the bytewise comparison might not work if there are multiple public keys with that ID. hm…

niacdoial: well that last remark was dumb.
I just got caught up on the name `s.crypto.FindRecipients` (which can take more than one key ID) and on `if kl > 0`.
Just to be sure, do you confirm that `kl > 1` shouldn't be possible there?

internal/store/leaf/link.go
```diff
@@ -36,9 +36,10 @@ func (s *Store) Link(ctx context.Context, from, to string) error {
...
 	// try to enqueue this task, if the queue is not available
 	// it will return the task and we will execute it inline
-	t := queue.GetQueue(ctx).Add(func(ctx context.Context) error {
-		return s.gitCommitAndPush(ctx, to)
+	t := queue.GetQueue(ctx).Add(func(ctx context.Context) (context.Context,error) {
+		return nil,s.gitCommitAndPush(ctx, to)
 	})
 
-	return t(ctx)
+	ctx,err := t(ctx)
```
dominikschulz, line 43: This ctx value is never used.

internal/store/leaf/link.go
```diff
@@ -36,9 +36,10 @@ func (s *Store) Link(ctx context.Context, from, to string) error {
 
 	// try to enqueue this task, if the queue is not available
 	// it will return the task and we will execute it inline
-	t := queue.GetQueue(ctx).Add(func(ctx context.Context) error {
-		return s.gitCommitAndPush(ctx, to)
+	t := queue.GetQueue(ctx).Add(func(ctx context.Context) (context.Context,error) {
+		return nil,s.gitCommitAndPush(ctx, to)
```
dominikschulz, line 40: Most files are not properly formatted. Please use gofmt or goimports.

internal/store/leaf/move.go
```diff
@@ -115,11 +122,12 @@ func (s *Store) directMove(ctx context.Context, from, to string, del bool) error
...
 	// try to enqueue this task, if the queue is not available
 	// it will return the task and we will execute it inline
-	t := queue.GetQueue(ctx).Add(func(ctx context.Context) error {
-		return s.gitCommitAndPush(ctx, to)
+	t := queue.GetQueue(ctx).Add(func(_ context.Context) (context.Context,error) {
+		return nil,s.gitCommitAndPush(ctx, to)
 	})
 
-	return t(ctx)
+	ctx,err := t(ctx)
```
dominikschulz, line 129: This value of ctx is never used.

internal/store/leaf/reencrypt.go
```diff
@@ -55,9 +55,11 @@ func (s *Store) reencrypt(ctx context.Context) error {
...
 					}
 					if err := s.Set(WithNoGitOps(ctx, conc > 1), e, content); err != nil {
-						logger.Printf("Worker %d: Failed to write %s: %s\n", workerId, e, err)
-
-						continue
+						if !errors.Is(err, store.ErrMeaninglessWrite) {
+							logger.Printf("Worker %d: Failed to write %s: %s\n", workerId, e, err)
+							continue
+						}
+						logger.Printf("Worker %d: Writing secret %s is not needed\n")
```
dominikschulz, line 62: printf args are missing.

internal/store/leaf/templates.go
```diff
@@ -116,6 +116,9 @@ func (s *Store) SetTemplate(ctx context.Context, name string, content []byte) er
 	p := s.templatefile(name)
 
 	if err := s.storage.Set(ctx, p, content); err != nil {
+		if errors.Is(err, store.ErrMeaninglessWrite) {
+			return fmt.Errorf("no need to overwrite template")
```
dominikschulz, line 120: If we don't want to overwrite because it already has the right content shouldn't we consider to return nil instead?

internal/store/leaf/write.go
```diff
@@ -69,11 +69,12 @@ func (s *Store) Set(ctx context.Context, name string, sec gopass.Byter) error {
...
 	// try to enqueue this task, if the queue is not available
 	// it will return the task and we will execute it inline
-	t := queue.GetQueue(ctx).Add(func(ctx context.Context) error {
-		return s.gitCommitAndPush(ctx, name)
+	t := queue.GetQueue(ctx).Add(func(_ context.Context) (context.Context,error) {
+		return nil,s.gitCommitAndPush(ctx, name)
 	})
 
-	return t(ctx)
+	ctx, err = t(ctx)
```
dominikschulz, line 76: ctx is not used.

pkg/ctxutil/ctxutil.go
```diff
@@ -153,7 +156,11 @@ func HasGitCommit(ctx context.Context) bool {
 
 // IsGitCommit returns the value of git commit or the default (true).
 func IsGitCommit(ctx context.Context) bool {
-	return is(ctx, ctxKeyGitCommit, true)
+	bv, ok := ctx.Value(ctxKeyGitCommit).(bool)
```
dominikschulz, line 159: The `is` helper should exaclty do this. Why did you replace it?

pkg/ctxutil/ctxutil.go
```diff
@@ -251,17 +258,61 @@ func IsForce(ctx context.Context) bool {
...
+// WithMultiSecretOperation returns a context with a flag set
+func WithMultiSecretOperation (ctx context.Context, bv bool) context.Context {
+	return context.WithValue(ctx, ctxKeyIsMultiSecretOperation, bv)
+}
+
+// IsMultiSecretOperation returns true if the flag is set to true
+func IsMultiSecretOperation (ctx context.Context) bool {
+	if hasBool(ctx,ctxKeyIsMultiSecretOperation) {
+		return is(ctx, ctxKeyForce, true)
+	} else {
```
dominikschulz, line 270: remove the else and outdent the `return false`, please.

pkg/ctxutil/ctxutil.go
```diff
@@ -251,17 +258,61 @@ func IsForce(ctx context.Context) bool {
...
+	}
+}
+
+
+// AddToCommitMessageBody returns a context with something added to the commit's body
+func AddToCommitMessageBody(ctx context.Context, sv string) context.Context {
+	if hasString(ctx, ctxKeyCommitMessageBody) {
+		current_body, ok := ctx.Value(ctxKeyCommitMessageBody).(string)
+		if ok {
+			current_body += "\n"
```
dominikschulz, line 281: no underscores in variable names, please.

pkg/ctxutil/ctxutil.go
```diff
@@ -251,17 +258,61 @@ func IsForce(ctx context.Context) bool {
...
+	if hasBool(ctx,ctxKeyIsMultiSecretOperation) {
+		return is(ctx, ctxKeyForce, true)
+	} else {
+		return false
+	}
+}
+
+
+// AddToCommitMessageBody returns a context with something added to the commit's body
+func AddToCommitMessageBody(ctx context.Context, sv string) context.Context {
```
dominikschulz, line 277: I'm not super happy with this approach. It should work, but it's neither elegant nor efficient.

What you're doing is:
- get an existing string from the context (if remains in the content)
- concatenate it to create a new strings
- add that to the context (the other value is still there)

In most cases this shouldn't be a big deal, but I can imagine very big password stores where this might become an issue.

So what about another approach: Create a new commit message type (either a custom type with a strings.Builder/bytes.Buffer) or just a ref to one of those. If we only store the reference in the context we can manipulate the value and don't need to add it again on every single invocation.

pkg/ctxutil/ctxutil.go
```diff
@@ -271,6 +322,50 @@ func GetCommitMessage(ctx context.Context) string {
 	return sv
 }
 
+// GetCommitMessageFull returns the set commit message (head+body, of either are defined) or an empty string.
+func GetCommitMessageFull(ctx context.Context) string {
+	sv_head, ok_head := ctx.Value(ctxKeyCommitMessage).(string)
```
dominikschulz, line 327: no underscores, please

pkg/ctxutil/ctxutil.go
```diff
@@ -271,6 +322,50 @@ func GetCommitMessage(ctx context.Context) string {
...
+		return sv_head + "\n\n" + sv_body
+	}
+}
+
+// ClearErrorCollector returns a context with where the error collector is emptied.
+func ClearErrorCollector(ctx context.Context) context.Context {
+	return context.WithValue(ctx, ctxKeyErrorCollector, nil)
+}
+// AddToErrorCollector returns a context with a new error logged as a new line into the error collector
+func AddToErrorCollector(ctx context.Context, sv string) context.Context {
```
dominikschulz, line 345: My remarks about the commit message builder also apply here.
Instead of concatenating strings we might create (or use an existing) multi error type that is referenced here.

But: do we really need to store this in the context?
If there other options are too cumbersome that's find, but I do consider using the context as a last restort. It's not very elegant and we've been trying to recude the use of the context in the past already.

internal/action/edit.go
```diff
@@ -61,7 +63,10 @@ func (s *Action) editUpdate(ctx context.Context, name string, content, nContent
 
 	// write result (back) to store.
 	if err := s.Store.Set(ctxutil.WithCommitMessage(ctx, fmt.Sprintf("Edited with %s", ed)), name, nSec); err != nil {
-		return exit.Error(exit.Encrypt, err, "failed to encrypt secret %s: %s", name, err)
+		if !errors.Is(err, store.ErrMeaninglessWrite) {
+			return exit.Error(exit.Encrypt, err, "failed to encrypt secret %s: %s", name, err)
+		}
+		out.Warningf(ctx, "The new value of the password is equal to its current value. Are you sure you want this?")
```
dominikschulz, line 69: This is a question without a choice. If we do not want to give the user a choice we shouldn't include a question in the first place.

internal/store/leaf/templates.go
niacdoial: done

pkg/ctxutil/ctxutil.go
niacdoial: no idea why, but I replaced it back

pkg/ctxutil/ctxutil.go
niacdoial: that part of the code in ctxutil.go was unused anyway, so I just removed it.
(I originally added it before fsck was moved to `WithQueue(nil)`)

dominikschulz (dismissed):
LGTM. Thanks!

dominikschulz:
@niacdoial Thanks a lot. Could you please have a look at the build errors? Then we should be good to go.

internal/store/leaf/fsck.go
```diff
@@ -79,13 +103,31 @@ func (s *Store) fsckLoop(ctx context.Context, path string) error {
...
 		}
 	}
 
-	if err := s.fsckUpdatePublicKeys(ctx); err != nil {
-		out.Errorf(ctx, "Failed to update public keys: %s", err)
+	if errcoll != "" {
+		out.Errorf(ctx, errcoll)
+	}
+	if ctxutil.GetCommitMessageBody(ctx) == "" {
+		out.Errorf(ctx, "Nothing to commit: all secrets seemed to have failed")
```
dominikschulz, line 118: This branch is also executed if there is nothing to commit because nothing was updated. This shouldn't yield an error.

internal/store/leaf/reencrypt.go
```diff
@@ -41,7 +41,7 @@ func (s *Store) reencrypt(ctx context.Context) error {
 		jobs := make(chan string)
 		// We use a logger to write without race condition on stdout
 		logger := log.New(os.Stdout, "", 0)
-		out.Printf(ctx, "Starting reencrypt")
+		out.Println(ctx, "Starting reencrypt")
```
dominikschulz, line 44: There is no out.Println. Use out.Printf, please.

niacdoial: Used `out.Print()` instead, it also exists

niacdoial:
Concerning the build errors, I have no idea what is happening. I know it builds on my side (with nix, build script modified from the official one, under the cut in case it is relevant)

However, the CI build gives me empty errors when I check the "actions" tab of my copy of gopass (no stdout or stderr visible, but a return code of 2 for the main step of the build for all 3 platforms)

<hr/>

```nix
{ lib
, stdenv
, makeWrapper
, buildGo118Module
, fetchurl
, installShellFiles
, git
, gnupg
, xclip
, wl-clipboard
, passAlias ? false
}:

buildGo118Module rec {
  pname = "gopass";
  version = "dev";

  nativeBuildInputs = [ installShellFiles makeWrapper ];

  src = ./src;
#  src = fetchurl {
#    url = "file:///home/niac/gopasss/src";
#    url = "/home/niac/gopasss/src";
#    rev = "ad18f385f3b95ad8f42b8e6146a6adaf3f52ab59";
#    sha256 = "sha256-0000000000000000000000000000000000000000000=";
#  };

#  vendorSha256 = "sha256-ESIlmY3SzmfWLqLxf62dOhAsHPOKc2YYSOhzpJCjvCQ=";
  vendorSha256 = null;

  subPackages = [ "." ];

  ldflags = [ "-s" "-w" "-X main.version=${version}" "-X main.commit=000000000000" ];

  wrapperPath = lib.makeBinPath (
    [
      git
      gnupg
      xclip
    ] ++ lib.optional stdenv.isLinux wl-clipboard
  );

  postInstall = ''
    installManPage gopass.1
    installShellCompletion --zsh --name _gopass zsh.completion
    installShellCompletion --bash --name gopass.bash bash.completion
    installShellCompletion --fish --name gopass.fish fish.completion
  '' + lib.optionalString passAlias ''
    ln -s $out/bin/gopass $out/bin/pass
  '';

  postFixup = ''
    wrapProgram $out/bin/gopass \
      --prefix PATH : "${wrapperPath}" \
      --set GOPASS_NO_REMINDER true
  '';

  meta = with lib; {
    description = "The slightly more awesome Standard Unix Password Manager for Teams. Written in Go";
    homepage = "https://www.gopass.pw/";
    license = licenses.mit;
    maintainers = with maintainers; [ rvolosatovs sikmir ];
    changelog = "https://github.com/gopasspw/gopass/raw/v${version}/CHANGELOG.md";

    longDescription = ''
      gopass is a rewrite of the pass password manager in Go with the aim of
      making it cross-platform and adding additional features. Our target
      audience are professional developers and sysadmins (and especially teams
      of those) who are well versed with a command line interface. One explicit
      goal for this project is to make it more approachable to non-technical
      users. We go by the UNIX philosophy and try to do one thing and do it
      well, providing a stellar user experience and a sane, simple interface.
    '';
  };
}

```

dominikschulz (dismissed):
LGTM. Thanks a lot!

pkg/ctxutil/ctxutil_test.go
```diff
@@ -111,6 +111,22 @@ func TestCommitMessage(t *testing.T) {
 	assert.Equal(t, "", GetCommitMessage(WithCommitMessage(ctx, "")))
 }
 
+func TestCommitMessageBody(t *testing.T) {
+	t.Parallel()
+
+	ctx := context.Background()
+
+	ctx2 = AddToCommitMessageBody(AddToCommitMessageBody(WithCommitMessage(ctx, "foo"),"bar"),"baz")
```
dominikschulz, line 119: `ctx2` needs to be declared with `ctx2 := ...`.

internal/queue/background.go
```diff
@@ -63,7 +63,7 @@ func (n *noop) Idle(_ time.Duration) error {
 }
 
 // Task is a background task.
-type Task func(ctx context.Context) error
+type Task func(ctx context.Context) (context.Context,error)
```
dominikschulz, line 66: These files are still not formatted. Can you run `make fmt`, please?

pkg/ctxutil/ctxutil_test.go
niacdoial: done!
(see comment later on)

internal/queue/background.go
niacdoial: ran gofmt again, but this time I made sure it acted on all files relevant to this pull request (and also made sure it was the one from go 1.18)

niacdoial:
turns out, the nix build script only runs some of the tests, so that's a bunch of fixes I needed to add.

And CI now builds on windows and macos, but still not linux. (still exit code 2, no stdout, no stderr)

(also I found that rcs_test.go depends on the user having a configured global git name/email to work)

dominikschulz:
Temporarily disabling 'gci' seems to unblock the linters.

Feel free to add ' gci'  to the list of disabled linters in `.golangci.yml` and fix the remaining issues.
Once that looks good I hope we can merge the PR.

niacdoial:
The CI works!
(I didn't know what was the right place to look for failed CI runs, so now I got to fix all the lint hints)
it also seems that the build step on linux depends on the linter

dominikschulz (approved):
LGTM. Thanks a lot!

PR #2499, opened 2022-12-28
Add some hooks

dominikschulz:
Taking a lot of inspiration from https://github.com/woodruffw/kbs2. Kudos!

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -79,32 +79,35 @@ This is a list of available options:
 
 | **Option**       | **Type** | Description | *Default* |
 | ---------------- | -------- | ----------- | --------- |
+| `age.usekeychain`      | `bool`   | Use the OS keychain to cache age passphrases. | `false` |
+| `audit.concurrency`    | `int`    | Number of concurrent audit workers. | `` |
+| `autosync.interval`      | `int`   | AutoSync interval in days. | `3` |
 | `core.autoclip`        | `bool`   | Always copy the password created by `gopass generate`. Only applies to generate. | `false` |
 | `core.autoimport`      | `bool`   | Import missing keys stored in the pass repository without asking. | `false` |
 | `core.autosync`        | `bool`   | Always do a `git push` after a commit to the store. Makes sure your local changes are always available on your git remote. | `true` |
+| `core.clear-after` | `bool` | This setting determines wheter or not the clipboard is cleared after `gopass -c`. Setting this to `false` takes precedence over the value of `core.cliptimeout` | `true` |
 | `core.cliptimeout`     | `int`    | How many seconds the secret is stored when using `-c`. | `45` |
-|  `core.clear-after` | `bool` | This setting determines wheter or not the clipboard is cleared after `gopass -c`. Setting this to `false` takes precedence over the value of `core.cliptimeout` | `true` |
 | `core.exportkeys`      | `bool`   | Export public keys of all recipients to the store. | `true` |
 | `core.nocolor`         | `bool`   | Do not use color. | `false` |
 | `core.nopager`         | `bool`   | Do not invoke a pager to display long lists. | `false` |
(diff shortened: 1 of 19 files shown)


docs/hooks.md
```diff
@@ -0,0 +1,44 @@
...
+`gopass` exposes some hook-able events during it's invocation lifecycle. This allows uses to inject additional functionality of perform addition logging.
+
+## Hook API
+
+All hooks are subject to the follwing constraints:
+
+* Hooks do not inherit `STDIN` or `STDOUT` from the parent process.
+* Hooks do inherit `STDERR` from the parent process and may use it to print anything they want.
+* Hooks always run from the `password-store` directory.
+* Hooks are run with the `GOPASS_HOOK=1` in their environment and with `GOPASS_CONFIG_DIR` set to the configuration directory the original `gopass` command was started.
```
AnomalRoil, line 12:
```suggestion
* Hooks are run with the `GOPASS_HOOK=1` in their environment and with `GOPASS_CONFIG_DIR` set to the configuration directory with which the original `gopass` command was started.
```

docs/hooks.md
```diff
@@ -0,0 +1,44 @@
...
+* An exit from a hook (or execution failure) cases the entire `gopass` command to fail.
```
AnomalRoil, line 13:
```suggestion
* An exit from a hook (or execution failure) causes the entire `gopass` command to fail.
```

AnomalRoil: What if it's a post-hook? Would the whole operation still fail and it reverts to the initial state?

docs/config.md
```diff
@@ -83,6 +83,7 @@ This is a list of available options:
 | `core.autoimport`      | `bool`   | Import missing keys stored in the pass repository without asking. | `false` |
 | `core.autosync`        | `bool`   | Always do a `git push` after a commit to the store. Makes sure your local changes are always available on your git remote. | `true` |
 | `core.cliptimeout`     | `int`    | How many seconds the secret is stored when using `-c`. | `45` |
+|  `core.clear-after` | `bool` | This setting determines wheter or not the clipboard is cleared after `gopass -c`. Setting this to `false` takes precedence over the value of `core.cliptimeout` | `true` |
```
AnomalRoil, line 86: Isn't setting `core.cliptimeout` to 0 doing that already?

internal/action/delete.go
```diff
@@ -107,5 +108,5 @@ func (s *Action) deleteKeyFromYAML(ctx context.Context, name, key string) error
 		return exit.Error(exit.IO, err, "Can not delete key %q from %q: %s", key, name, err)
 	}
 
-	return nil
+	return hook.Invoke(ctx, "delete.post-rm", key+"@"+name)
```
AnomalRoil, line 111: Is the `@` symbol really not authorised in keys currently?

I tested it and it seems a key can contain the symbol "@" right now, so this might not be a great separator, since people might be using email addresses as keys.

AnomalRoil (commented):
That's a cool idea, I'm looking forward to it :D 

Here are a few comments while I'm at it, sorry I know it's still a draft.

docs/hooks.md
dominikschulz: Good question. Some operations are hard to revert. I don't think I can easily manage that complexity without a transactional backend. So: no. The command will fail but everything that's been done will stay as it is.

docs/config.md
dominikschulz: No. We're using a default of 45 in that case.

https://github.com/gopasspw/gopass/blob/master/pkg/clipboard/clipboard.go#L47

internal/action/delete.go
dominikschulz: Yeah, you're right. How should we signal to the hook that this is a key? Maybe just as a separate arg?!

AnomalRoil: I feel a separate argument would be best, yes.

PR #2505, opened 2023-01-02
Use custom multi error type for fsck instead of string concat

dominikschulz:
RELEASE_NOTES=n/a

### DIFF
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -193,7 +193,7 @@ func (s *Store) fsckLoop(ctx context.Context, path string) error {
 
 	// print out any deferred warnings (if any)
 	if warnings.Len() > 0 {
-		out.Errorf(ctx, warnings.String())
+		out.Warning(ctx, warnings.String())
 	}
 
 	if ctxutil.GetCommitMessageBody(ctx) == "" {
@@ -300,7 +300,7 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) (string, error)
 	merr = s.fsckCheckRecipients(ctx, name)
 	if merr.ErrorOrNil() != nil {
 		if merr.Severity == errsFatal {
-			_ = errs.Append(merr.Severity, fmt.Errorf("Checking recipients for %s failed:\n    %w", name, err))
+			_ = errs.Append(merr.Severity, fmt.Errorf("Checking recipients for %s failed:\n    %w", name, merr))
 		} else {
 			_ = errs.Append(merr.Severity, merr)
 		}
--- a/internal/store/leaf/fsck.go
+++ b/internal/store/leaf/fsck.go
@@ -257,7 +257,7 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) (string, error)
 		// the only errsNonFatal error from that function are missing/extra recipients,
 		// which isn't much of an error since we have yet to correct that.
 		recpNeedFix = true
-		errs.Append(merr.Severity, merr)
+		_ = errs.Append(merr.Severity, merr)
 	}
 
 	// make sure we are actually allowed to decode this secret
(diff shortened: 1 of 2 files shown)


internal/store/leaf/fsck.go
```diff
@@ -227,61 +292,61 @@ func (s *Store) fsckCheckEntry(ctx context.Context, name string) (string, ErrorS
...
 	}
 
-	errs, err = s.fsckCheckRecipients(ctx, name)
-	if err != nil {
-		if errs == errsFatal {
-			errcoll += fmt.Errorf("Checking recipients for %s failed:\n    %w", name, err).Error()
+	merr = s.fsckCheckRecipients(ctx, name)
+	if merr.ErrorOrNil() != nil {
+		if merr.Severity == errsFatal {
+			_ = errs.Append(merr.Severity, fmt.Errorf("Checking recipients for %s failed:\n    %w", name, err))
```
AnomalRoil, line 303: Do you really want to have `err` here in that `fmt.Errorf`?

internal/store/leaf/fsck.go
```diff
@@ -120,19 +181,23 @@ func (s *Store) fsckLoop(ctx context.Context, path string) error {
...
 		}
+
+		ctx = ctxutil.AddToCommitMessageBody(ctx, msg)
 	}
 
-	if errcoll.Len() > 0 {
-		out.Errorf(ctx, errcoll.String())
+	// print out any deferred warnings (if any)
+	if warnings.Len() > 0 {
+		out.Errorf(ctx, warnings.String())
```
AnomalRoil, line 196: Are these warnings or errors? 
I think we also have `out.Warningf` for the former.

internal/store/leaf/fsck.go
dominikschulz: Thanks. My naming is terrible ...

internal/store/leaf/fsck.go
```diff
@@ -27,6 +27,67 @@ const (
...
+	}
+}
+
+type fsckMultiError struct {
+	Severity ErrorSeverity
+	Errors   []error
+}
+
+func (f *fsckMultiError) IsError() bool {
+	return len(f.Errors) > 0
```
AnomalRoil, line 47: What if all `Errors`'s severity is `0`? Is it still an error?

internal/store/leaf/fsck.go
```diff
@@ -27,6 +27,67 @@ const (
...
+		return ""
+	}
+
+	var sb strings.Builder
+	fmt.Fprintf(&sb, "[%s] ", f.Severity.String())
+	msgs := make([]string, 0, len(f.Errors))
+	for _, e := range f.Errors {
+		msgs = append(msgs, e.Error())
+	}
+	sb.WriteString(strings.Join(msgs, ", "))
```
AnomalRoil, line 61: Wouldn't it be best to wrap errors using `%w` and `Fprintf` instead?

```suggestion
	addComma := false
	for _, e := range f.Errors {
		if addComma {
			fmt.Fprint(%sb, ", ")
		}
		fmt.Fprintf(%sb, "%w", e.Error())
		addComma = true
	}
```

AnomalRoil, line 61: Wouldn't it be best to wrap errors using `%w` and `Fprintf` instead?

```suggestion
	addComma := false
	for _, e := range f.Errors {
		if addComma {
			fmt.Fprint(%sb, ", ")
		}
		fmt.Fprintf(%sb, "%w", e.Error())
		addComma = true
	}
```

internal/store/leaf/fsck.go
```diff
@@ -180,36 +245,36 @@ type convertedSecret interface {
...
 	// make sure we are actually allowed to decode this secret
 	// if this fails there is no way we could fix anything
 	if !IsFsckDecrypt(ctx) {
 		if !recpNeedFix {
-			return "", errsNil, nil
+			return "", nil
 		}
 
-		return "", errsFatal, fmt.Errorf(errcoll + "\nRun fsck with the --decrypt flag to re-encrypt it automatically, or edit this secret yourself.")
+		return "", errs.Append(errsFatal, fmt.Errorf("Run fsck with the --decrypt flag to re-encrypt it automatically, or edit the secret %s yourself.", name)).ErrorOrNil()
```
AnomalRoil, line 270: I'm a bit puzzled here.
If we're not running with `--decrypt`, we would get one error and abort because of the `errsFatal` at the first entry with missing recipients, right? 
Or maybe worse, one per entry.

But then we don't continue and so we don't know for instance if there are MIME leftovers, unless we're running with `--decrypt`, right?

dominikschulz: Yes, I think so. But seriously: This code is a mess. It's been like that since I first wrote it. We're making baby steps to improve it, but not by much. Eventually we'll get there ... (but that will need more untangling and reporting changes).

internal/store/leaf/fsck.go
dominikschulz: I don't think `%w` is valid outside of `fmt.Errorf`.

PR #2506, opened 2023-01-02
Start new auditor

dominikschulz:

### DIFF
--- a/internal/audit/audit.go
+++ b/internal/audit/audit.go
@@ -5,14 +5,12 @@ package audit
 
 import (
 	"context"
-	"errors"
 	"fmt"
+	"sync"
 	"time"
 
-	"github.com/fatih/color"
 	"github.com/gopasspw/gopass/internal/backend"
 	"github.com/gopasspw/gopass/internal/config"
-	"github.com/gopasspw/gopass/internal/notify"
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
@@ -21,20 +19,6 @@ import (
 	"github.com/muesli/crunchy"
 	"github.com/nbutton23/zxcvbn-go"
 )
-
-// auditedSecret with its name, content a warning message and a pipeline error.
-type auditedSecret struct {
-	name string
-
-	// the secret's content as a string. Needed for checking for duplicates.
-	content string
-
-	// message to the user about some flaw in the secret.
-	messages []string
-
-	// real error that something in the pipeline went wrong.
-	err error
-}
 
 type secretGetter interface {
 	Get(context.Context, string) (gopass.Secret, error)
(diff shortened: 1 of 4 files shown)


pkg/gitconfig/config.go
```diff
@@ -296,7 +296,7 @@ func parseConfig(in io.Reader, key, value string, cb parseFunc) []string {
 		// These are odd but we should still support them.
 		k, v, found := strings.Cut(line, " = ")
 		if !found {
-			debug.Log("no valid KV-pair on line: %q", line)
+			// debug.Log("no valid KV-pair on line: %q", line)
```
AnomalRoil, line 299: No debug here?

pkg/gitconfig/config.go
```diff
@@ -383,7 +383,7 @@ func ParseConfig(r io.Reader) *Config {
 	c.raw.WriteString(strings.Join(lines, "\n"))
 	c.raw.WriteString("\n")
 
-	debug.Log("processed config: %s\nvars: %+v", c.raw.String(), c.vars)
+	// debug.Log("processed config: %s\nvars: %+v", c.raw.String(), c.vars)
```
AnomalRoil, line 386: no debug?

internal/audit/report.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+	s.Warnings = append(s.Warnings, msg)
+	r.secrets[name] = s
+}
+
+func newReport() *ReportBuilder {
+	return &ReportBuilder{
+		secrets:    make(map[string]SecretReport, 512),
+		duplicates: make(map[string]set.Set[string], 512),
+		sha1sums:   make(map[string]set.Set[string], 512),
```
AnomalRoil, line 123: Why use `set.Set`? I don't see how we could have twice the same name (i.e. paths) provided to either.

internal/audit/report.go
```diff
@@ -0,0 +1,146 @@
...
+type ReportBuilder struct {
+	sync.Mutex
+
+	secrets map[string]SecretReport
+	// SHA512(password) -> secret names
+	duplicates map[string]set.Set[string]
+
+	// HIBP
+	// SHA1(password) -> secret names
+	sha1sums map[string]set.Set[string]
```
AnomalRoil, line 52: I commented elsewhere already, but re-iterating that I don't see how we could get two secrets with the same path so I don't see why we need `set.Set` for these?

internal/audit/audit.go
```diff
@@ -119,177 +138,124 @@ func Batch(ctx context.Context, secrets []string, secStore secretGetter, expirat
...
-			checked <- as
+	sec, err := a.s.Get(ctx, secret)
+	if err != nil {
+		debug.Log("Failed to check %s: %s", secret, err)
 
-			continue
+		a.r.AddError(secret, err)
+		if sec != nil {
+			a.r.AddPassword(secret, sec.Password())
 		}
```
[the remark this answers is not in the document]
AnomalRoil: In any case it's missing the check `sec.Password()` is not empty unlike done below.

internal/audit/report.go
dominikschulz: This map is from sha1sum to secret name(s).

If we have multiple secrets that contain the same secret (very likely scenario) we need to handle that.

internal/audit/report.go
dominikschulz: See my other comment

pkg/gitconfig/config.go
dominikschulz: We're way to noisy already. Until we have better log filtering I want to disable some of them.

pkg/gitconfig/config.go
dominikschulz: Too noisy

internal/audit/report.go
AnomalRoil: But then we really need to map from 1 string to a slice of strings. `map[string][]string` should be enough, no?

dominikschulz: Yes, that would work, too.

internal/audit/audit.go
```diff
@@ -119,177 +138,124 @@ func Batch(ctx context.Context, secrets []string, secStore secretGetter, expirat
...
-		as := auditedSecret{
-			name: secret,
-		}
 		// check for context cancelation.
 		select {
 		case <-ctx.Done():
-			as.err = errors.New("user aborted")
-			checked <- as
-
 			continue
```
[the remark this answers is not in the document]
dominikschulz: No. We need to ensure to consume all secrets from the chan otherwise the writer might block.

internal/audit/audit.go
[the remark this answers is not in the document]
dominikschulz: No. If we get an error but still a non-nil secret (not sure how that could happen), we still want to record the password.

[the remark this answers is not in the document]
AnomalRoil: You're doing a 
```
        // do not check empty secrets.
	if sec.Password() == "" {
		return
	}
	
	// add the password for the duplicate check
	a.r.AddPassword(secret, sec.Password())
``` 
below but not here, so in case of empty password (e.g. an entry used only for its keys) and of an error, it might be incorrectly flagged.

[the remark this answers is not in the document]
dominikschulz: I think you are right, but I agree that it's not a high priority right now.
The reports, even after this PR, are still not great. I want to improve them, but there is already a lot in this PR and at some point I should stop breaking things ;)

PR #2532, opened 2023-01-29
Fix tests and roll over updater key

dominikschulz:
The updater key expired.


AnomalRoil (approved):
LGTM.

But that error I had never seen:

> ERROR: Some modules have unrecognized licenses:
  github.com/russross/blackfriday/v2: path '/home/runner/go/pkg/mod/github.com/russross/blackfriday/v2@v2.1.0/LICENSE.txt'

PR #2550, opened 2023-02-09
Default to using the root stores backends for newly added mounts

### LINKED ISSUE #2515: gopass does not remember which encryption and storage it currently uses
Related discussion: #2022 

------

It is possible to create a new password store from inside a running gopass session by typing `setup`. However, this seems to have a fallback to Git and GPG.

## Reproducing the problem(s):

1. Initialize gopass with Fossil and age: `gopass setup --crypto age --storage fossilfs`
2. Run gopass
3. Type `setup`

### Expected behavior:

A new password store with the current options (Fossil and age) is created.

### Actual behavior:

gopass defaults to Git and GnuPG.

dominikschulz:
Fixes #2515

RELEASE_NOTES=[ENHANCEMENT] Default to the root stores backends for mounts.

### DIFF
--- a/internal/action/action_test.go
+++ b/internal/action/action_test.go
@@ -39,7 +39,7 @@ func newMock(ctx context.Context, path string) (*Action, error) {
 	c := cli.NewContext(cli.NewApp(), fs, nil)
 	c.Context = ctx
 	if err := act.IsInitialized(c); err != nil {
-		return nil, err
+		return act, err
 	}
 
 	return act, nil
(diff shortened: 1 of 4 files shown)


AnomalRoil (commented):
I wonder if the tests are failing because the `cli` flags are racy... We're passing a `map[string]string` to the `CliCtxWithFlags` function, which just passes it further to the CLI library for parsing. 

I see that running `go test -race ./...` blows up in the `TestInitContext`...

Issue #2555, opened 2023-02-13
`gopass create` fails due to template error


ixil:
### Summary
`gopass create` fails with: `could not write default templates: failed to write default template .gopass/create/0-websites.yml: meaningless write`.
The `0-websites.yml` template already exsits under the `password-store\.gopass\create\0-websites.yml`
I'm on windows, gopass installed using `scoop`

### Steps To Reproduce
`gopass create` ...?

### Environment
gopass installed using `scoop`

- OS: Win 11
- gopass Version: gopass 1.15.4 go1.19.5 windows amd64

dominikschulz:
Oh, I know where this is coming from. We need to catch this specific error. And maybe investigate why it's trying to write the template (again) at all.

ixil:
I've tried deleting the templates `0-websites.yml` and `1-pin.yml` - now it's a git commit error.

dominikschulz:
Something is odd here and I wonder if it's a windows issue. This should only happen if gopass can't parse the templates.

Could you try it again with the GOPASS_DEBUG_LOG variable set to some log file location?

ixil:
Two attemepts, `gopass create` when the templates existed, and then again after I deleted the templates - which interestiingly enough gets to the wizard, but there are no options (other than to quit with q obviously).

```
2023/02/14 09:42:06.530450 gitconfig/configs.go:156	gitconfig.(*Configs).loadGlobalConfigs	loaded global config from C:\Users\olive\AppData\Roaming\gopass\config
2023/02/14 09:42:06.531067 gitconfig/configs.go:67	gitconfig.(*Configs).LoadAll	Loading gitconfigs for &{Preset:<nil> system:0xc00007ca00 global:0xc00007ca40 local:0xc00007ca80 worktree:0xc00007cac0 env:0xc00007cb00 workdir: SystemConfig:/etc/gopass/config GlobalConfig: LocalConfig:config WorktreeConfig:config.worktree EnvPrefix:GOPASS_CONFIG_ NoWrites:false} ...
2023/02/14 09:42:06.531067 gitconfig/configs.go:73	gitconfig.(*Configs).LoadAll	failed to load system config: open /etc/gopass/config: The system cannot find the path specified.
2023/02/14 09:42:06.531587 gitconfig/configs.go:156	gitconfig.(*Configs).loadGlobalConfigs	loaded global config from C:\Users\olive\AppData\Roaming\gopass\config
[...]

PR #2558, opened 2023-02-15
fix: template write error on windows

ckolumbus:
harmonize internal path representations so that
path matching uses always backslash notation under windows.

fixes: #2555

### DIFF
--- a/internal/create/wizard.go
+++ b/internal/create/wizard.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


internal/create/wizard.go
```diff
@@ -88,7 +89,12 @@ func New(ctx context.Context, s backend.Storage) (*Wizard, error) {
 }
 
 func (w *Wizard) parseTemplates(ctx context.Context, s backend.Storage) ([]Template, error) {
-	tpls, err := s.List(ctx, tplPath)
+	var tplPath_int string
```
dominikschulz, line 92: Using underscores in variable names is uncommon in Go. Please use `camelCase`.

dominikschulz:
This PR is failing some tests. Feel free to re-open if you get it to build cleanly.

ckolumbus:
I waited for your feedback... I'll check everything again now that the solution is ok with you

PR #2568, opened 2023-03-11
remove $ from shell snippets

stefanotorresi:
Starting all the shell code blocks with `$` negates the ability to quickly copy-paste the commands (which is made very convenient by GitHub via a button over the code block).


dominikschulz (approved):
Thanks!

PR #2569, opened 2023-03-12
feat: add core.autopush to separate push/autosync behavior

sudoforge:
```
feat: add core.autopush to separate push/autosync behavior

This change adds a `core.autopush` configuration option (which defaults
to a value of `true`). This new configuration option is used in
post-write-ish actions, to determine if the remote repository should be
pushed to. In doing this, we support workflows where a user may want to
always push to the remote, but disable the behavior of "sync", which
fetches updates from remotes for all mounts (including the root store).

Closes: gopasspw/gopass#2551
```


sudoforge:
For sure. I had to wrestle with CI a bit (cough cough, if it builds, it works! ... right, test pipelines? right?), but sans any input from you, we're ready to merge this whenever you have the time for a proper review.

dominikschulz (approved):
LGTM. Thanks!

sudoforge:
woohoo! i'm glad this made it in, it's been a personal pet peeve for a while.

please feel free to cc me on any issue that comes up related to this config property.

PR #2596, opened 2023-05-27
fuzz ParseAKV

hickford:
Further fuzz tests could test round trip etc


AnomalRoil (approved):
LGTM, it's a good idea, we should fuzz more of our parsers.

dominikschulz:
This does only add the entry point for fuzzing, not the actual fuzzing, correct?

I haven't kept fully task of recent fuzzing changes in Go.

dominikschulz:
@AnomalRoil Shall we try to get covered by oss-fuzz? I'm not sure if we meet the bar, but we can try.

PR #2597, opened 2023-05-31
Adding screen parsing for OTP QR code 

AnomalRoil:
This adds the capability to parse the content of the (multiple) display of the user to detect and add an OTP QR code to a given entry.

Usage:

```
$ gopass otp -snip testentry
⚠ Scanning screen n°0
✅ Area scanned on screen n°0: (0,0)-(3440,1440)
⚠ Found a qrcode, checking.
⚠ Not an otpauth:// QR code, please make sure to only have your OTP qrcode displayed.
⚠ Scanning screen n°1
✅ Area scanned on screen n°1: (0,0)-(2560,1440)
⚠ Found a qrcode, checking.
✅ Found an otpauth:// QR code on screen n°1 ((0,0)-(2560,1440))
(Over)writing otpauth URL in key 'otpauth'? [Y/n/q]:
Value written, carrying on to display OTP value from it.
824573
```

I also updated our dependencies, while I was at it, by running `go get -u -t ./...` and `go mod tidy`.


AnomalRoil:
What should we do about the license @dominikschulz ? To me it seems okay since we already have deps using the BSD license, but not sure.

The generator part with the WTFPL license is fine too.

dominikschulz:
BSD is definitely fine with me. WTFPL should be ok as well.

dominikschulz:
This is fine. We won't do that 🙂

AnomalRoil:
Should I add WTFPL to the `.license-lint.yml` and add xgb to our license allowlisted_modules in there too?

dominikschulz (approved):
LGTM

nit: Usually I prefer to split updating dependencies and code changes, but this time we can leave it as is.

PR #2606, opened 2023-06-15
gitfs: support git submodules

0x450x6c:
Hello.

I use git submodule as a store, and gopass didn't recognize it as a git repo.

With this change gopass can recognize git submodules as a repository, so `gopass sync` works. (tested locally)

### DIFF
--- a/internal/backend/storage/gitfs/git.go
+++ b/internal/backend/storage/gitfs/git.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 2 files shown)


dominikschulz:
Oh my, I don't like git submodules. The UX is terrible. Let me see how much complexity this would add ...

dominikschulz (changes requested):
Sorry for the long delay. The code itself looks fine, but we need two things before we can merge it:

1. Comments, explaining what is happening. I'm not deeply familiar with submodules and some explanation is necessary to make sense of this.
2. Tests, showing and validating the implementation.

dominikschulz:
No progress in a year. Closing this for now. But feel free to re-open if you want to get back to this.

PR #2617, opened 2023-07-11
[Fix] Using the correct env prefix for env settings

AnomalRoil:
Currently our config options can be set using the `GOPASS_CONFIG__CONFIG_COUNT`  option followed by `GOPASS_CONFIG__CONFIG_KEY_0` and same for values, but this seems to be a bug.

It should be `GOPASS_CONFIG_CONFIG_KEY_0` without 2 underscores according to our documentation.

This PR fixes it.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,7 +11,7 @@ import (
 )
 
 var (
-	envPrefix    = "GOPASS_CONFIG_"
+	envPrefix    = "GOPASS_CONFIG"
 	systemConfig = "/etc/gopass/config"
 )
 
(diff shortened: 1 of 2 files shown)


AnomalRoil:
We might consider to rename `GOPASS_CONFIG_CONFIG_KEY_0` into `GOPASS_CONFIG_OVERRIDE_KEY_0` and same for the rest too. WDYT?

dominikschulz:
This change looks good. Thanks for taking care.

Let me think about the proposed rename for a bit.

dominikschulz:
The rename should be fine, but we can do that in another PR.

PR #2625, opened 2023-07-30
Correctly handle new multiline secrets

### LINKED ISSUE #2614: Copying empty secret when multiline is inserted
### Summary
After adding multiline password (also when it is only one line) the first line is empty and there I am not able to copy password via gopass command. I have to edit it firstly and manually remove first empty line. Then it is working as expected.

### Steps To Reproduce
```
$ gopass insert -m personal/test
# vim opens
pass1
# :wq to save and quit

$ gopass personal/test
Secret: personal/test


pass1

$ gopass -c personal/test

Error: empty secret. see https://go.gopass.pw/faq#empty-secret
```

### Expected behavior
```
$ gopass personal/test
Secret: personal/test

pass1

$ gopass -c personal/test
✔ Copied personal/test to clipboard. Will clear in 45 seconds.
```

### Environment
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.2 LTS"

$ uname -a
Linux mcuper-P14s 5.19.0-46-generic #47~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Wed Jun 21 15:35:31 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

$ gopass version
gopass 1.15.5 go1.20.2 linux amd64

$go install github.com/gopasspw/gopass@latest

dominikschulz:
This commit fixes as small issue in how multi-line secrets are handled. Before they were always written in to the secret body completly ignoring the first line that contains the password. Now we do respect that correctly. To implement that properly we need to have some additional code to satisfy the io.Writer assumptions around the AKV secret type.

Also this fixes some non-hermetic tests that showed up during testing of this change.

Fixes #2614

### DIFF
--- a/internal/action/insert.go
+++ b/internal/action/insert.go
@@ -185,6 +185,7 @@ func (s *Action) insertGetSecret(ctx context.Context, name, pw string) (gopass.S
 
 // insertYAML will overwrite existing keys.
 func (s *Action) insertYAML(ctx context.Context, name, key string, content []byte, kvps map[string]string) error {
+	debug.Log("insertYAML: %s - %s -> %s", name, key, content)
 	if ctxutil.IsInteractive(ctx) {
 		pw, err := termio.AskForString(ctx, name+":"+key, "")
 		if err != nil {
@@ -200,12 +201,15 @@ func (s *Action) insertYAML(ctx context.Context, name, key string, content []byt
 		if err != nil {
 			return exit.Error(exit.Encrypt, err, "failed to set key %q of %q: %s", key, name, err)
 		}
+		debug.Log("using existing secret %s", name)
 	} else {
 		sec = secrets.New()
+		debug.Log("creating new secret %s", name)
 	}
 
 	setMetadata(sec, kvps)
 
+	debug.Log("setting %s to %s", key, string(content))
 	if err := sec.Set(key, string(content)); err != nil {
 		return exit.Error(exit.Usage, err, "failed set key %q of %q: %q", key, name, err)
 	}
--- a/pkg/gopass/secrets/akv.go
+++ b/pkg/gopass/secrets/akv.go
@@ -333,7 +333,7 @@ func (a *AKV) SafeStr() string {
 
 // pwWriter is a io.Writer that will extract the first line of the input stream and
 // then write it to the password field of the provided AKV. The first line can stretch
-// multiple chunks but once the first line hass been completed any writes to this
+// multiple chunks but once the first line has been completed any writes to this
 // writer will be silently discarded.
 type pwWriter struct {
 	a       *AKV
--- a/pkg/gopass/secrets/akv.go
+++ b/pkg/gopass/secrets/akv.go
@@ -308,14 +308,15 @@ func (a *AKV) Write(buf []byte) (int, error) {
 	var w io.Writer
 	w = &a.raw
 
-	// If the body is empty before writing we must threat the first line
+	// If the body is empty before writing we must treat the first line
 	// of the newly written content as the password or multi-line insert
-	// and similar operations will fail. For regular command line usage
-	// this is actually a non-issue since the gopass process will exit
-	// after writing and when reading the secret it will be (correctly)
-	// parsed. But for tests, library and maybe REPL usage this is an issue.
+	// and similar operations (like insert from stdin) might fail.
+	// For regular command line usage this is actually a non-issue since
+	// the gopass process will exit after writing and when reading the secret
+	// it will be (correctly) parsed. But for tests, library and maybe REPL
+	// usage this is an issue.
 	if a.raw.Len() == 0 {
-		w = io.MultiWriter(&a.raw, &pwWriter{a: a})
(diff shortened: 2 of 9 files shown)


pkg/gopass/secrets/akv.go
```diff
@@ -310,3 +330,32 @@ func (a *AKV) FromMime() bool {
...
+type pwWriter struct {
+	a       *AKV
+	buf     strings.Builder
+	written bool
+}
+
+func (p *pwWriter) Write(buf []byte) (int, error) {
+	n := len(buf)
+	if p.written || p.a == nil {
+		return n, nil
```
AnomalRoil, line 347:
```suggestion
		return 0, nil
```
Since no writes are done in that case, no?

pkg/gopass/secrets/akv_test.go
```diff
@@ -310,3 +341,21 @@ func FuzzParseAKV(f *testing.F) {
...
+	assert.NoError(t, err)
+
+	_, err = p.Write([]byte("bar\n"))
+	assert.NoError(t, err)
+
+	// but anything after the first line is discarded
+	_, err = p.Write([]byte("baz\n"))
+	assert.NoError(t, err)
+
+	assert.Equal(t, "foobar", a.Password())
```
AnomalRoil, line 360: Shouldn't the test also test the content of Body?

pkg/gopass/secrets/akv.go
```diff
@@ -298,7 +305,20 @@ func (a *AKV) Body() string {
 
 // Write appends the buffer to the secret's body.
 func (a *AKV) Write(buf []byte) (int, error) {
-	return a.raw.Write(buf)
+	var w io.Writer
+	w = &a.raw
+
+	// If the body is empty before writing we must threat the first line
```
AnomalRoil, line 311:
```suggestion
	// If the body is empty before writing we must treat the first line
```

pkg/gopass/secrets/akv.go
```diff
@@ -298,7 +305,20 @@ func (a *AKV) Body() string {
...
+	// of the newly written content as the password or multi-line insert
+	// and similar operations will fail. For regular command line usage
```
AnomalRoil, line 313: I don't understand what you mean by "and similar operations will fail"?

pkg/gopass/secrets/akv.go
```diff
@@ -310,3 +330,32 @@ func (a *AKV) FromMime() bool {
...
+		return n, nil
+	}
+
+	i := bytes.Index(buf, []byte("\n"))
+	if i > 0 {
+		p.buf.Write(buf[0:i])
+		p.a.password = p.buf.String()
+		p.written = true
+
+		return n, nil
```
AnomalRoil, line 356: Shouldn't this be 
```suggestion
		return i, nil
```
Since it only writes the pw?

AnomalRoil (changes requested):
I'm surprised we need such a "complex" solution, I would have expected removing the newline in NewAKV would already solve the multi-line issue, but I guess you tried it and it had side effects.

Just a few comments around the Write returns, since usually we'd return the number of bytes actually written.

dominikschulz:
I did try the simpler approach first, indeed. It had side-effects for some of our tests. Unfortunately. I'll address the comments later.

pkg/gopass/secrets/akv.go
dominikschulz: Actually this did raise some questions with me that I didn't have a good answer for and I ditched the MultiWriter approach for a pass-through wirter. I think that's easier to reason about.

pkg/gopass/secrets/akv_test.go
dominikschulz: Yes, that's more comprehensive.

pkg/gopass/secrets/akv.go
dominikschulz: Like insert from STDIN. That should be affected here as well. Updated the comment.

pkg/gopass/secrets/akv.go
dominikschulz: See my comment above. I think the approach was not great.

AnomalRoil (commented):
Just a little comment about the logs.

I'm not sure we needed the extra writer implementation since we do the `Write` call on the whole buffer anyway and not in a way we could stream with a reader/writer but lgtm

internal/action/insert.go
```diff
@@ -185,6 +185,7 @@ func (s *Action) insertGetSecret(ctx context.Context, name, pw string) (gopass.S
 
 // insertYAML will overwrite existing keys.
 func (s *Action) insertYAML(ctx context.Context, name, key string, content []byte, kvps map[string]string) error {
+	debug.Log("insertYAML: %s - %s -> %s", name, key, content)
```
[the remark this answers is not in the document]
dominikschulz: these should be `safeString` types, so they shouldn't leak I think?

But if they do we can remove these.

PR #2631, opened 2023-07-31
Update Makefile to include BSD in crosscompile

dominikschulz:
This should help to catch cross-platform issues before we attempt to build a release in the future.


AnomalRoil:
How about just running `goreleaser build` in our build GHA?

AnomalRoil (approved):

dominikschulz:
Oh, I see. `goreleaser build` != `goreleaser release --snapshot`. That might actually work.

AnomalRoil:
I think it would. We'd just need to tell it to skip tag verification

PR #2635, opened 2023-08-04
[BUGFIX] Use goreleaser build for crosscompile

dominikschulz:
This should make sure that we catch all platforms we target during release.


AnomalRoil (approved):
LGTM. Surprisingly, it works locally. Not sure why GHA don't like it.

PR #2647, opened 2023-09-04
[docs] add/update choco, scoop, winget instructions

dpprdan:
This adds/updates Windows installation instructions with chocolatey, scoop, and winget.

The [scoop](https://github.com/ScoopInstaller/Main/blob/master/bucket/gopass.json) and [choco](https://github.com/lyze237/GopassChoco/blob/master/gopass.nuspec) manifests do not include the git and gpg4win dependencies, so they have to be installed separately. Should we add those to the README? 
There is a `choco install gpg4win` there already, which is why I added `winget install git` and `winget install gpg4win`. The alternative would be to add the dependencies to the manifests so that `choco/scoop install gopass` would also install the dependencies.

~~The winget manifest already includes both dependencies, but the winget cli does not support installing those just yet ([the next minor version will add this functionality](https://github.com/microsoft/winget-cli/issues/163#issuecomment-1697930113)).~~ winget probably needs some more time to properly support dependencies (i.e. I am currently failing at submitting the manifest with dependencies included due to (IMHO) overzealous checks).

BTW is this warning still valid?
https://github.com/gopasspw/gopass/blob/439bc00268a67139b69cb22c7d083ab6d4add040/docs/setup.md?plain=1#L204

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ pacman -S gopass
 [![Scoop version](https://img.shields.io/scoop/v/gopass)](https://github.com/ScoopInstaller/Main/blob/master/bucket/gopass.json)
 
 ```shell
-# WinGet#
+# WinGet
 winget install Git.Git
 winget install GnuPG.Gpg4win
 winget install gopass.gopass
--- a/README.md
+++ b/README.md
@@ -77,7 +77,9 @@ pacman -S gopass
 [![Scoop version](https://img.shields.io/scoop/v/gopass)](https://github.com/ScoopInstaller/Main/blob/master/bucket/gopass.json)
 
 ```shell
-# WinGet
+# WinGet#
+winget install Git.Git
+winget install GnuPG.Gpg4win
 winget install gopass.gopass
 # Chocolatey
 choco install gpg4win
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -62,7 +62,7 @@ choco install gpg4win
 
 ```ps1
 scoop install git
-# add Extras bucket
+# add Extras bucket, if you haven't already
 # scoop bucket add extras
 scoop install gpg4win
 ```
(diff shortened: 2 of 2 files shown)


docs/setup.md
```diff
@@ -51,9 +51,28 @@ brew install gnupg2 git
...
+choco install git
+choco install gpg4win
+```
+
+[Scoop](https://scoop.sh), 
+
+```ps1
+scoop install git
+# add Extras bucket
+# scoop bucket add extras
```
AnomalRoil, line 66: Is the second one really meant to be commented?
Does it require it or not?

AnomalRoil:
You're right, I think the warning can go, Windows support is there to stay.

docs/setup.md
dpprdan: I assumed that most scoop users have it activated already. But you're right, it is not really clear. How about

```suggestion
# add Extras bucket, if you haven't already
# scoop bucket add extras
```

dpprdan:
I still wonder what the best strategy re the dependencies/backends is (on Windows). gopass installs fine without them and you can also run it without them. age is also an alternative to gpg, but I don't know how well it works in Windows. So I see three options:

1. do not add dependencies to the manifests, and
		a. just add git and gpg4win them to the installation instructions.
		b. add git, gpg, and, alternatively age (not sure whether it is in any of the package manager repos, though) to the installation instructions.
2. add git and gpg4win to the manifests so they get automatically installed with gopass (where possible, otherwise 1a - I assume that this should work with choco and scoop)

README.md
```diff
@@ -77,7 +77,9 @@ pacman -S gopass
 [![Scoop version](https://img.shields.io/scoop/v/gopass)](https://github.com/ScoopInstaller/Main/blob/master/bucket/gopass.json)
 
 ```shell
-# WinGet
+# WinGet#
```
dominikschulz, line 80: The trailing `#` shouldn't be there, I think.

dominikschulz (approved):

README.md
dpprdan: oops, you're right!

```suggestion
# WinGet
```

dominikschulz (approved):
Thanks!

PR #2651, opened 2023-09-07
[feat] Add XCKD pwgen config options

### LINKED ISSUE #2595: make xkcdsep  and xkcdlang config options
### Summary

It would be nice if    --xkcdsep  and --xkcdlang could be set in config (generate.generator.lang or similar)

### Steps To Reproduce

Feature request, so no reproducing

### Expected behavior

if set, `gopass generate foo` would create an xkcd password with the configured seperator and the configured language, thus making it obsolete to provide those flags each time

dominikschulz:
This commit adds three new config options to influence the default behaviour of the XKCD password generator.

* pwgen.xkcd.sep for the word separator (string)
* pwgen.xkcd.lang for the language (string)
* pwgen.xkcd.len for the number of words (int)

Fixes #2595

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -317,33 +317,41 @@ func (s *Action) generatePasswordForRule(ctx context.Context, c *cli.Context, le
 // generatePasswordXKCD walks through the steps necessary to create an XKCD-style
 // password.
 func (s *Action) generatePasswordXKCD(ctx context.Context, c *cli.Context, length string) (string, error) {
-	xkcdSeparator := " "
+	sep := config.String(c.Context, "pwgen.xkcd.sep")
 	if c.IsSet("sep") {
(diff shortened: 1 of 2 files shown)


AnomalRoil (commented):
LGTM, but should we also change it here:
https://github.com/gopasspw/gopass/blob/9666b0a8500efbb81968af76900651ad340cdcfa/internal/action/generate.go#L326-L329
to not ask for the length when one is configured?

Also note that I think the `gopass generate -g xkcd nameofentry` command isn't affected by this change since it doesn't use `xkcdGen`:
https://github.com/gopasspw/gopass/blob/9666b0a8500efbb81968af76900651ad340cdcfa/internal/action/generate.go#L346

So this effectively only affects `gopass pwgen` which is arguably less useful than `gopass generate`, no?

PR #2652, opened 2023-09-07
Use default length defined in config file

sylvainpelissier:
Takes into account the configuration `generate.length` by the wizard will creating a new password.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"strconv"
@@ -8,6 +9,11 @@ import (
 
 	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/gitconfig"
+)
+
+const (
+	DefaultLength     = 24
+	DefaultXKCDLength = 4
 )
 
 var (
@@ -236,4 +242,28 @@ func (c *Config) Keys(mount string) []string {
 	}
 
 	return nil
+}
+
+// DefaultLengthFromEnv will determine the password length from the env variable
+// GOPASS_PW_DEFAULT_LENGTH or fallback to the hard-coded default length.
+// If the env variable is set by the user and is valid, the boolean return value
+// will be true, otherwise it will be false.
+func DefaultLengthFromEnv(ctx context.Context) (int, bool) {
+	def := DefaultLength
+	cfg := FromContext(ctx)
+
+	if l := cfg.GetInt("generate.length"); l > 0 {
+		def = l
+	}
+
+	lengthStr, isSet := os.LookupEnv("GOPASS_PW_DEFAULT_LENGTH")
+	if !isSet {
+		return def, false
+	}
+	length, err := strconv.Atoi(lengthStr)
+	if err != nil || length < 1 {
+		return def, false
+	}
+
+	return length, true
 }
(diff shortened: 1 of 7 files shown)


internal/config/config.go
```diff
@@ -237,3 +243,30 @@ func (c *Config) Keys(mount string) []string {
...
+// GOPASS_PW_DEFAULT_LENGTH or fallback to the hard-coded default length.
+// If the env variable is set by the user and is valid, the boolean return value
+// will be true, otherwise it will be false.
+func DefaultLengthFromEnv(ctx context.Context) (int, bool) {
+	def := DefaultLength
+	cfg := FromContext(ctx)
+
+	if l := cfg.GetInt("generate.length"); l > 0 {
+		def = l
+	}
```
AnomalRoil, line 257: You could use `config.Int(ctx, "generate.length")`

```suggestion
	if l := config.Int(ctx, "generate.length"); l > 0 {
		def = l
	}
```

internal/config/config.go
```diff
@@ -237,3 +243,30 @@ func (c *Config) Keys(mount string) []string {
...
+	if !isSet {
+		return def, false
+	}
+	length, err := strconv.Atoi(lengthStr)
+	if err != nil {
+		return def, false
+	}
+	if length < 1 {
+		return def, false
+	}
```
AnomalRoil, line 269:
```suggestion
	if err != nil || length < 1 {
		return def, false
	}
```

AnomalRoil (changes requested):

internal/config/config.go
```diff
@@ -10,6 +11,11 @@ import (
 	"github.com/gopasspw/gopass/pkg/gitconfig"
 )
 
+const (
+	DefaultLength     = 24
+	DefaultXKCDLength = 4
```
AnomalRoil, line 16: Do you really need to move that one here?

sylvainpelissier: Yes because in need it inside `DefaultLengthFromEnv` function and I may have import problem if I leave it in the former package. In addition, it sound more like a config part.

AnomalRoil: Since it's exported, how about using this one in the former package and removing it there then?

AnomalRoil (approved):
LGTM

internal/config/config.go
```diff
@@ -10,6 +11,11 @@ import (
 	"github.com/gopasspw/gopass/pkg/gitconfig"
 )
 
+const (
+	DefaultLength     = 24
```
dominikschulz, line 15: `DefaultLength` is very generic. gopass is mainly concered with passwords, but does also handle other data. So we should be more specific here, e.g. `DefaultPasswordLength`?

internal/config/config.go
```diff
@@ -237,3 +243,27 @@ func (c *Config) Keys(mount string) []string {
 
 	return nil
 }
+
+// DefaultLengthFromEnv will determine the password length from the env variable
+// GOPASS_PW_DEFAULT_LENGTH or fallback to the hard-coded default length.
+// If the env variable is set by the user and is valid, the boolean return value
+// will be true, otherwise it will be false.
+func DefaultLengthFromEnv(ctx context.Context) (int, bool) {
```
dominikschulz, line 251: This should be called`DefaultPasswordLengthFromEnv` (or simliar).

dominikschulz (changes requested):

dominikschulz (approved):
LGTM

PR #2655, opened 2023-09-08
Add config generate.strict

l4u:


internal/action/generate.go
```diff
@@ -239,15 +239,29 @@ func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, n
...
 	switch generator {
 	case "memorable":
+		isStrict := false
 		if c.Bool("strict") {
+			isStrict = true
+		} else if cfg.IsSet("generate.strict") {
+			isStrict = cfg.GetBool("generate.strict")
+		}
+
+		if isStrict {
```
dominikschulz, line 249: Can you move the isStrict evaluation to a method and just call that here and below? So we have less duplication.

dominikschulz (changes requested):

internal/action/generate.go
```diff
@@ -524,3 +524,15 @@ func filterPrefix(in []string, prefix string) []string {
 
 	return out
 }
+
+func isStrict(ctx context.Context, c *cli.Context) bool {
+	cfg := config.FromContext(ctx)
+
+	if c.Bool("strict") {
+		return true
+	} else if cfg.IsSet("generate.strict") {
```
dominikschulz, line 533: You don't need the `else` here.

I think you can just remove the else (but keep the if part after moving it to the next line) and the code will work just like now but be a bit more idiomatic.

dominikschulz (changes requested):
Looking very good, once more change, please.

dominikschulz (approved):
LGTM. Thanks a lot for following along.

PR #2658, opened 2023-09-11
[feat] Add --store option to gopass fsck

dominikschulz:
This allows limiting fsck to a specific mount-point to speed up the process for large password stores.

### DIFF
--- a/internal/store/root/fsck.go
+++ b/internal/store/root/fsck.go
@@ -10,11 +10,15 @@ import (
 )
 
 // Fsck checks all stores/entries matching the given prefix.
-func (s *Store) Fsck(ctx context.Context, path string) error {
+func (s *Store) Fsck(ctx context.Context, store, path string) error {
 	var result []error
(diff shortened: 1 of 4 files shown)


internal/store/root/fsck.go
```diff
@@ -10,14 +10,18 @@ import (
 )
 
 // Fsck checks all stores/entries matching the given prefix.
-func (s *Store) Fsck(ctx context.Context, path string) error {
+func (s *Store) Fsck(ctx context.Context, store, path string) error {
```
AnomalRoil, line 13: You'll need to change the usage in our tests. LGTM otherwise

AnomalRoil (approved):

internal/store/root/fsck.go
dominikschulz: Yes, I saw that. Thank you.

I'll fix that once I get a chance to.

PR #2670, opened 2023-09-19
chore: remove extraneous whitespace characters

sudoforge:
This change removes extraneous whitespace (tab, space) characters from lines in `//:zsh.completion`.

### DIFF
--- a/zsh.completion
+++ b/zsh.completion
@@ -14,29 +14,29 @@ _gopass () {
 	      "identities:List identities"
 	      )
 	      _describe -t commands "gopass age" subcommands
-	      
-	      
-	      
+
+
+
 	      ;;
 	  alias)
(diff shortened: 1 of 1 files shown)


sudoforge:
CI failure for `golangci-lint / lint` appears to be caused by a timeout (re-running it will likely resolve it). If a maintainer is not around, I can rebase this commit and force push to my tree in order to trigger a new run (but this will re-run _every_ job, which can/should be avoided if possible).

dominikschulz:
This is a generated file that is overwritten on every release. Please remove the extra whitespaces from the source instead.

sudoforge:
sorry, haven't had much time to address this recently. i'll get back to this sometime in the next week or two.

PR #2681, opened 2023-11-02
[CLEANUP] Migration of options to more appropriate sections

AnomalRoil:
This PR migrates a few "legacy" options into new config sections, since we didn't have sectioning back when we created these. Namely:

| **Legacy option name** | **New option name** | 
|------------------------|---------------------|
| `core.showsafecontent` | `show.safecontent`  | 
| `core.autoclip`        | `generate.autoclip` |
| `core.showautoclip`    | `show.autoclip`     | 

This also adds an easy migration path to our config handling, which should allow us to migrate option names around much more easily in the future. I've also tweaked our doc config test to detect legacy options that are still in use.

Any system level config or Env variables options are **not** migrated. 

This also fixes a bug in our test code, where the root mount path was not properly set in our config, because we used "path:" instead of "path=" to set it in our config.

I've also spotted some wild, weird spaces which I've replaced with regular spaces and I've reformatted our markdown tables since I was playing with our documentation. I've also fixed a few typos here and there.

Since we discussed it previously in #2617, this also migrates away from using `GOPASS_CONFIG_CONFIG_VALUE_n` instead of `GOPASS_CONFIG_VALUE_n` for specifying custom options through Env variables, in order to match how GIT_CONFIG works. 

NB: This is a big PR, but each commit is distinct and I'm happy opening multiple PRs if you prefer.


AnomalRoil:
It seems our linter on CI got some new Linters because we're using `latest` and `enable-all` and so it was complaining about a lot of things, I had to add a huge commit, the second to last because of it to fix them all. 

It seems we also started advertising Go1.21 in our go.mod file, but our CI was still on Go1.20, I've changed it in my last commit.

Maybe I should first do another PR with these two CI commits and then rebase this branch on top. What do you prefer @dominikschulz ?

internal/config/config.go
```diff
@@ -244,6 +263,48 @@ func (c *Config) Keys(mount string) []string {
...
+					found = true
+				}
+				if val := cfg.Get(oldK); !found && val != "" {
+					debug.Log("Found old option %s = %s in config, probably at the worktree or env level, "+
+						"or maybe at the system level cannot migrate it.", oldK, val)
+				}
+			}
+		}
+	}
+	if err := errors.Join(errors.Join(errs...)); err != nil {
```
dominikschulz, line 303: Why do you need to chain `errors.Join` twice?

dominikschulz:
Splitting off the two huge commits would have been nice, but now that I've made my way through this huge PR we can leave it as is.

Thanks a lot, this is a very nice improvement and cleanup effort.

dominikschulz (dismissed):

internal/config/config.go
AnomalRoil: That's a mistake, I refactored that part a couple times and I messed up I guess 😅

dominikschulz (approved):

AnomalRoil:
I'm always very sad there isn't some kind of global git config option to sign-off commits T.T
This should now be all green hopefully.

Issue #2683, opened 2023-11-02
gopass must remove syscall(2) usage on OpenBSD


jrick:
### Summary

gopass built for OpenBSD from the master branch will fail to link to syscall(2) at runtime once [OpenBSD removes syscall(2) from libc](https://marc.info/?l=openbsd-tech&m=169841790407370&w=2).

There are two usages of syscall(2) that I could find: one is due to pledge (will be fixed by https://golang.org/cl/468095), and the other is in readline, which I've [opened a PR for](https://github.com/chzyer/readline/pull/232).  Unfortunately, it doesn't seem that the owner is maintaining that readline library anymore, and a switch to that other fork might be needed, unless you wish to use a module replacement that points to my PR's branch.

### Steps To Reproduce

```
$ pwd
/home/jrick/src/gopass
$ go build
$ ./gopass
gopass:./gopass: undefined symbol 'syscall'
ld.so: gopass: lazy binding failed!
Killed 
```

### Expected behavior

gopass executes without hitting a lazy linking error.

### Environment

- OS: OpenBSD
- OS version: -current with the full syscall(2) removal patch, including removal from libc
- gopass Version: master branch
- Installation method: repo

### Additional context

A working gopass can be built with patches to x/sys and readline.

```
$ pwd
/home/jrick/src/gopass
$ got br
master
$ got status
$ go work init
$ go work use .
$ go work use ../sys # golang.org/cl/468095
$ go work use ../readline # github.com/chzyer/readline/pull/232
$ go build
$ ./gopass

   __     _    _ _      _ _   ___   ___
 /'_ '\ /'_'\ ( '_'\  /'_' )/',__)/',__)
( (_) |( (_) )| (_) )( (_| |\__, \\__, \
'\__  |'\___/'| ,__/''\__,_)(____/(____/
( )_) |       | |
 \___/'       (_)

ð Welcome to gopass!
â  This is the built-in shell. Type 'help' for a list of commands.
gopass> ^D
```
[](url)

jrick:
One other additional note, this patched and working gopass still does reference syscall(2), but I could not find where, and have not seen any other linking crashes.  It might be benign, or I may not have tested thoroughly enough.

```
$ nm -s ./gopass | grep 'U syscall' 
         U syscall
```

dominikschulz:
Thank you. I think we might try switching to https://github.com/ergochat/readline if the author doesn't merge your PR soon.

jrick:
[...]

PR #2684, opened 2023-11-03
Replace chzyer/readline with ergochat/readline

dominikschulz:
The former seemed unmaintained and was lacking patches required for OpenBSD.

See GH-2683.

Some small changes were necessary since the ergochat fork changed the API a bit.


AnomalRoil (dismissed):
LGTM

jrick:
The REPL is working well for me with this on OpenBSD.

dominikschulz:
The problem is that the `clear` command will be broken on Windows since ergochat/readline doesn't expose a `ClearScreen` method: https://github.com/ergochat/readline/issues/36

Not sure if we wan't to merge it was is.

AnomalRoil (approved):
LGTM. You'll need to rebase to sign-off the last commit of just merge as is without satisfying the DCO.

dominikschulz:
The code looks good, but I want to quickly tests it before merging. Should be done soon and then we should try to cut a release.

PR #2685, opened 2023-11-04
[BUGFIX] Disabling the OTP snip screenshot feature on OpenBSD

AnomalRoil:
Disabling the OTP snip screenshot feature on OpenBSD as discussed in #2683.

Also do not rely on our dependency returning an error on Darwin and disable that too if it's not compiled with CGO.

@dominikschulz If you have a Darwin device handy, it'd be cool if you could test the `gopass otp -snip testentry` feature with this branch both with and without CGO enabled to confirm whether it works fine or not.
```
CGO_ENABLED=1 go build
``` 
should do it.

### DIFF
--- a/pkg/otp/screenshot_others.go
+++ b/pkg/otp/screenshot_others.go
@@ -1,5 +1,5 @@
-//go:build !(arm || arm64 || amd64 || 386) || !(linux || windows || darwin || freebsd || netbsd || openbsd)
-// +build !arm,!arm64,!amd64,!386 !linux,!windows,!darwin,!freebsd,!netbsd,!openbsd
+//go:build !(arm || arm64 || amd64 || 386) || !(linux || windows || freebsd || netbsd) || (darwin && !cgo)
+// +build !arm,!arm64,!amd64,!386 !linux,!windows,!freebsd,!netbsd darwin,!cgo
 
 package otp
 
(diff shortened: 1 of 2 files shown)


pkg/otp/screenshot_others.go
```diff
@@ -1,5 +1,5 @@
-//go:build !(arm || arm64 || amd64 || 386) || !(linux || windows || darwin || freebsd || netbsd || openbsd)
-// +build !arm,!arm64,!amd64,!386 !linux,!windows,!darwin,!freebsd,!netbsd,!openbsd
+//go:build !(arm || arm64 || amd64 || 386) || !(linux || windows || freebsd || netbsd) || (darwin && !cgo)
```
jrick, line 1: i suspect this won't build on (darwin && cgo) because of a duplicate ParseScreen definition.

In addition to including the _supported.go file, (darwin && cgo) is not (linux || windows || freebsd || netbsd).  So it would evaluate to true and include this file too.

This needs to be the inverse of the supported build tags:

```
//go:build !(arm || arm64 || amd64 || 386) || !(linux || windows || (cgo && darwin) || freebsd || netbsd)
```

or

```
//go:build !((arm || arm64 || amd64 || 386) && (linux || windows || (cgo && darwin) || freebsd || netbsd))
```

AnomalRoil: Ah right, I was trying to tweak it since this is too complex for +build line, but actually since we're not supporting old Go versions, so we don't need to keep these anyway.

dominikschulz (approved):

PR #2696, opened 2023-11-17
Fixed bash completion command in setup docs

fetwar:
Updated bash completion command in `setup.md` that would not persist past end of shell.

Installing bash completions in the system wide location should be better suggesting by default user-level install in `~/.bashrc` as `gopass` is installed system wide as well.

This should be more inline with best practices for allowing the user to manually set their bash completion commands per https://github.com/scop/bash-completion

Adding completions by default on install would be better in the long term though. Potentially as part of package install by default, as pass does (see [pass makefile](https://github.com/zx2c4/password-store/blob/master/Makefile)) or a builtin command to install completions post download.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -303,11 +303,17 @@ Before migrating to gopass, you may have been using other password managers (suc
 
 ### Enable Bash Auto completion
 
-If you use Bash, you can run one of the following commands to enable auto completion for sub-commands like `gopass show`, `gopass ls` and others.
+If you use Bash, you can use the following command to enable auto completion for all users for sub-commands like `gopass show`, `gopass ls` and others.
 
 ```bash
-source <(gopass completion bash)
+gopass completion bash | sudo tee $(pkg-config --variable=completionsdir bash-completion)
 ```
+
+To enable bash completions for the current user only:
+```bash
+grep -q "source <(gopass completion bash)" ~/.bashrc || echo "source <(gopass completion bash)" >> ~/.bashrc
+```
+
 
 **MacOS**: The version of bash shipped with MacOS may [require a workaround](https://stackoverflow.com/questions/32596123/why-source-command-doesnt-work-with-process-substitution-in-bash-3-2) to enable auto completion. If the instructions above do not work try the following one:
 


dominikschulz (approved):
LGTM

docs/setup.md
```diff
@@ -303,12 +303,18 @@ Before migrating to gopass, you may have been using other password managers (suc
 
 ### Enable Bash Auto completion
 
-If you use Bash, you can run one of the following commands to enable auto completion for sub-commands like `gopass show`, `gopass ls` and others.
+If you use Bash, you can use the following command to enable auto completion for all users for sub-commands like `gopass show`, `gopass ls` and others.
 
 ```bash
-source <(gopass completion bash)
+gopass completion bash | sudo tee $(pkg-config --variable=completionsdir bash-completion)
```
AnomalRoil, line 309: Is pkg-config really usually bundled with bash or with most distros nowadays?

fetwar:
I think @AnomalRoil may have a good point.

## Availability of `pkg-config`

Looking into it `pkg-config` does seem to be quite universal and installed with `bash-completion` at least on debian, but admittedly my understanding of how widespread it is is lacking.

`bash-completion` is a separate package for me on `apt`, whilst it doesn't seem to always be installed with `bash`, it is a recommended package. So I believe that following the instructions here: [github.com/scop/bash-completions](https://github.com/scop/bash-completion/) should get us what we want, since this appears to be the official bash completions package.

Part of the bash package description reads:
```
 The Programmable Completion Code, by Ian Macdonald, is now found in
 the bash-completion package.
```

From my understanding, installing `bash-completion` installs the `pkg-config` variable I use in this PR.

How does gopass find distro default install locations etc normally?

Overall, I agree that finding a more robust way of installing completions is a good idea, but that it shouldn't be up to the user to do it.

---
## Improving Further

There are more options for robust bash completion install I discovered while researching briefly for this PR, but I avoided including them for brevity of the docs. I think this would suit an automated install well though, where error handling can be much more in depth.

To summarise my understanding of bash completions following brief research:

**System wide completions methods:**
- `/etc/bash_completion.d` - the old recommended install location for completions.
    - This initiates all completions on shell start for all users (eager loads), usually sourced within system wide `profile` script.
    - Legacy install location because of the eager loading, but makes a good fallback compared to just not having completions at all.
- `/usr/share/bash-completion/completions` - the new install location for completions.
    - Initiates dynamically when required, allowing for faster shell startup times afaik
    - Not a fixed location, can change distro to distro, which is why I believe the info I followed from [scoop/bash-completions](https://github.com/scop/bash-completion) mentions using `pkg-config` to find it's location. However, I don't trust `pkg-config` to always be installed.

**User specific completion methods:**
- `~/.bashrc`
    - Initiates on shell start (eager loads) and is user specific
    - User specific is a good method to install, however for system wide apps I think system wide makes more sense
- `~/.bash_completions`
    - Mentioned as good abstracted location compared to having completions directly within `~/.bashrc`, but also eagerly loads.

There's a few others mentioned in the scop/bash-completions github, but I will be the first to admit this isn't my area of expertise.

I think that adding proper install of bash completions automatically when installing the software is the best method, where we can check that packages like `pkg-config` and `bash-completion` (to provide the variable I mention in the docs) are installed, using the locations provided if so, falling back to defined fallback paths if not.

Is there a `make` Makefile that is run for install that we can edit to include this? I can't see why bash completions shouldn't be installed by default as part of the package. Notably installing `pass` via `make` or `apt` (presumably most other methods, although personally untested) does install completions.

fetwar:
Found an issue in this PR, silly typo on my end, will amend and re-open or create another if needed

Instead of
```
gopass completion bash | sudo tee $(pkg-config --variable=completionsdir bash-completion)
```

It should be
```
gopass completion bash | sudo tee $(pkg-config --variable=completionsdir bash-completion)/gopass
```

dominikschulz:
We can require or recommend the `bash-completion` package in our Deb and RPM packages.

PR #2699, opened 2023-11-18
[enhancement] Add input validation to block illegal mount points

dominikschulz:
This commit adds a check to prevent mounting mount points that can not be used later on.


AnomalRoil (approved):
Lgtm.

PR #2703, opened 2023-11-19
[feature] Add new pwgen options to capitalize and include numbers in

### LINKED ISSUE #2573: [Feature Request] `gopass pwgen -x` should support more options/flags
### Summary
Currently, the default behavior of the -x flag is rather limited and should be extended to support a few additional options for
increased entropy/security, while still being easy to type/pseudo-remember

# Steps To Reproduce
If you do -x with no separator, you get passwords such as:

```
gopass pwgen -x
craftsman division enunciate empty
```

If you do -x --sep "", you in turn get:
```
gopass pwgen -x --sep ""
BlusteryWombPasswordUnloader
```

and lastly, if you do -x --sep "#' you get:
```
gopass pwgen -x --sep "#"
aviation#backward#schnapps#frequent
```

### Expected behavior
One of the most basic changes that could be made would be to allow Capitalization and Number support on
the xkcd style passwords.

The way it works in `bw` is at least a start:
```
bw generate -cup --includeNumber --words 6 --separator ^
Gravity^Dish^Regalia5^Treadmill^Vexingly^Drop-down
```

I personally would like to see it allow for `-n` and `-c`, both of which could even allow for `<optional number>` options

It could allow for the result above, where a single capital is introduced at either word[1] or word [:-1] of any 1 word, up to capitals at the first/end of every word (word[1] && word[:-1])

It could also allow for the same with numbers, a number could go at the beginning or end of any 1 word or all words or both.

This would increase entropy by quite a bit, as seen here: https://rumkin.com/tools/password/

`BlusteryWombPasswordUnloader` -- 154 bits of entropy
`craftsman division enunciate empty` -- 166 (This is more secure than no space because space is at least an additional character type)
`aviation#backward#schnapps#frequent` -- 178 bits of entropy
`Gravity^Dish^Regalia5^Treadmill^Vexingly^Drop-down` -- 317 bits of entropy

### Additional context
From a security perspective, most users do not know how to efficiently manage passwords no matter what we do.  Additionally, there are unfortunately many sites that still have horrid implementations of 'no more than N characters long, no symbols other than !@#$' etc..  Using these types of things in combinations at least allows for secure password generation, that
also allows a user to do things like set their ldap passwords or sso passwords that actually require manually typing.  Otherwise, we might as well just go with things like:  `71a_N+,>\AVjj1^Yv]#AL\%gE1~iVPO-`, but even that password is actually only 207 bits of entropy, is 100x harder (not literally, but typing it is much worse than typing Gravity^Dish...

dominikschulz:
xkcd style passwords.

Depends on martinhoefling/goxkcdpwgen#10

Fixes #2573


AnomalRoil (approved):
LGTM, but I feel these kind of tweaks might benefit from having a config option to set the default value. WDYT?

pkg/pwgen/xkcdgen/pwgen.go
```diff
@@ -16,20 +16,21 @@ func Random() string {
...
 
 // RandomLengthDelim returns a random passphrase combined from the desired number
 // of words and the given delimiter. Words are drawn from lang.
-func RandomLengthDelim(length int, delim, lang string) (string, error) {
+func RandomLengthDelim(length int, delim, lang string, capitalize, numbers bool) (string, error) {
 	g := xkcdpwgen.NewGenerator()
 	g.SetNumWords(length)
 	g.SetDelimiter(delim)
-	g.SetCapitalize(delim == "")
+	g.SetCapitalize(delim == "" || capitalize)
```
AnomalRoil, line 28: TIL that we already supported the feature when we set the delimiter the the empty string, nice.

Issue #2673, opened 2023-09-27
gopass config core.notifications false → "Error: Error setting config value: failed to stage config file: …"


tchernomax:
### Summary

```
max@host % gopass config core.notifications false

Error: Error setting config value: failed to stage config file: exit status 128: fatal: pathspec 'config' did not match any files
```

note: the setting is changed even if gopass output an error

### Steps To Reproduce

```
gopass config core.notifications false
```

### Expected behavior

no error

### Environment

- OS: Archlinux
- OS version: Linux mde-claranet 6.5.4-arch2-1 #1 SMP PREEMPT_DYNAMIC Thu, 21 Sep 2023 11:06:39 +0000 x86_64 GNU/Linux
- gopass Version: gopass 1.15.8 go1.21.1 linux amd64
- Installation method: package ( https://archlinux.org/packages/extra/x86_64/gopass/ )

tjex:
I'm also encountering this on Mac 13.5. 

 Not only this, but I have `core.notifications = false` set, but I've started getting notifications (clipboard copy, clear, etc)

shiftleft1:
Also experiencing this in Ubuntu 20.04

AnomalRoil:
I can reproduce on master. 
This is because the `.config/gopass` folder is not using `git` I guess. 

I suspect these might be the cause:

https://github.com/gopasspw/gopass/blob/875cdb97a98bb7e7c05fcb2277a8783f96d5ebfb/internal/action/config.go#L92-L97

Here's the DEBUG logs:
```
2023/11/20 21:35:37.767111 action/config.go:81	action.(*Action).setConfigValue	setting core.notifications to false for ""
2023/11/20 21:35:37.767116 gitconfig/config.go:109	gitconfig.(*Config).Set	set "core.notifications" to "false"
2023/11/20 21:35:37.767120 gitconfig/config.go:118	gitconfig.(*Config).Set	updating value
2023/11/20 21:35:37.767128 gitconfig/config.go:218	gitconfig.(*Config).rewriteRaw	input (core.notifications: false): ---------

[recipients]
	hash = <somebighash>
[mounts]
	path = /home/anomalroil/.local/share/gopass/stores/root
[mounts "sarl"]
	path = /home/anomalroil/.local/share/gopass/stores/sarl
[generate]
	autoclip = true
[core]
	notifications = true

-----------
[...]

PR #2710, opened 2023-11-20
Tag v1.15.10

dominikschulz:


AnomalRoil:
I'd argue we should try and avoid shipping a tagged version with a `replace` statement in its go.mod file, I don't think we're in any hurry here, right? 

Plus I'd like to take a stab at fixing #2673 since I can reproduce it locally before the next release.

dominikschulz:
Good call. Debugging the release tooling is tedious. Let's push this out, indeed.

PR #2716, opened 2023-11-21
Improved config handling and bugfixes

AnomalRoil:
This PR is doing a few changes to the way we handle the Config action, its logging and also improves a few comments.

This is notably fixing #2673 by properly handling all the different cases for a _local_ vs _global_ config change not necessarily leading to a `git add` and `git commit` in the store.
Sadly, there is no way when using the gitconfig `Set` command to know if the option was written to the global vs local config, and so to do this really nicely we would need to change the signature of the `func (c *Config) Set(mount, key, value string) error` function to include a more explicit return value.
But we can fix the problem by properly handling the different errors than can arise during the Add and Commit phases of the config, which is what this PR does. 

This PR also prevents us from setting the global config while reporting a misleading "No existing configuration found" error.
Previously, upon using `gopass config some.name true` on an uninitialized store, we would get the following:
```
🌟 Welcome to gopass!
⚠ No existing configuration found.
☝ Please run 'gopass setup'
true
``` 
Now we get just the expected `true`, since we do support changing the global config without having a store. When trying to do it on a local config, tho, it will now fail:
- for the root store with:
	``` 
	$ gopass config --store "<root>" some.name true
	
	Error: Store <root> seems uninitialized or cannot be initialized
	```
- on a substore, with:
	``` 
	$ gopass config --store "test" some.name true
	Error: Error setting config value: failed to set config value "some.name": substore 'test' is not initialized or doesn't exist
	```

### DIFF
--- a/internal/action/config.go
+++ b/internal/action/config.go
@@ -37,7 +37,7 @@ func (s *Action) Config(c *cli.Context) error {
 
 	// sub-stores need to have been initialized so we can update their local configs.
 	// special case: we can always update the global config. NB: IsInitialized initializes the store if nil.
-	if inited, err := s.Store.IsInitialized(ctx); err != nil || store != "" && !inited {
+	if inited, err := s.Store.IsInitialized(ctx); err != nil || (store != "" && !inited) {
 		return exit.Error(exit.Unknown, err, "Store %s seems uninitialized or cannot be initialized", store)
 	}
 
--- a/internal/action/config.go
+++ b/internal/action/config.go
@@ -34,10 +34,10 @@ func (s *Action) Config(c *cli.Context) error {
 		return exit.Error(exit.Usage, nil, "Usage: %s config key value", s.Name)
 	}
 
-	// need to initialize sub-stores so we can update their configs.
-	// special case: we can always update the root config.
-	if err := s.IsInitialized(c); err != nil && store != "" {
-		return err
+	// sub-stores need to have been initialized so we can update their local configs.
+	// special case: we can always update the global config. NB: IsInitialized initializes the store if nil.
+	if inited, err := s.Store.IsInitialized(ctx); err != nil || store != "" && !inited {
+		return exit.Error(exit.Unknown, err, "Store %s seems uninitialized or cannot be initialized", store)
 	}
 
 	if err := s.setConfigValue(ctx, store, c.Args().Get(0), c.Args().Get(1)); err != nil {
--- a/internal/action/config.go
+++ b/internal/action/config.go
@@ -87,8 +87,8 @@ func (s *Action) setConfigValue(ctx context.Context, store, key, value string) e
 	}
 
 	// in case of a non-local config change we don't need to track changes in the store
-	if level != config.Local {
-		debug.Log("did not set local config (was config level %d) in store '%s', skipping commit phase", level, store)
+	if level != config.Local && level != config.Worktree {
+		debug.Log("did not set local config (was config level %d) in store %q, skipping commit phase", level, store)
 		s.printConfigValues(ctx, store, key)
 
 		return nil
@@ -99,22 +99,30 @@ func (s *Action) setConfigValue(ctx context.Context, store, key, value string) e
 		return fmt.Errorf("storage not available")
 	}
 
-	// notice that we rely on the cfg.Set above having created the local config file here. If it doesn't exist,
-	// st.Add will fail and return an error.
-	switch err := st.Add(ctx, "config"); {
+	configFile := "config"
+	if level == config.Worktree {
+		configFile = "config.worktree"
+	}
+
+	// notice that the cfg.Set above should have created the local config file.
+	if !st.Exists(ctx, configFile) {
+		return fmt.Errorf("local config file %s didn't exist in store, this is unexpected", configFile)
+	}
+
+	switch err := st.Add(ctx, configFile); {
 	case err == nil:
 		debug.Log("Added local config for commit")
 	case errors.Is(err, istore.ErrGitNotInit):
-		debug.Log("Skipping staging of local config: %v", err)
+		debug.Log("Skipping staging of local config %q: %v", configFile, err)
 	default:
-		return fmt.Errorf("failed to stage local config file: %w", err)
+		return fmt.Errorf("failed to Add local config %q: %w", configFile, err)
 	}
 
 	switch err := st.Commit(ctx, "Update config"); {
 	case err == nil:
 		debug.Log("Committed local config")
 	case errors.Is(err, istore.ErrGitNotInit), errors.Is(err, istore.ErrGitNothingToCommit):
-		debug.Log("Skipping staging of local config: %v", err)
+		debug.Log("Skipping Commit of local config: %v", err)
 	default:
 		return fmt.Errorf("failed to commit local config: %w", err)
 	}
--- a/internal/action/config.go
+++ b/internal/action/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	"github.com/gopasspw/gopass/internal/action/exit"
+	"github.com/gopasspw/gopass/internal/config"
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/internal/set"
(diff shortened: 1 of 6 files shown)


pkg/gitconfig/configs.go
```diff
@@ -145,6 +145,24 @@ func (cs *Configs) loadGlobalConfigs() string {
 		locs = append(locs, filepath.Join(appdir.UserHome(), cs.GlobalConfig))
 	}
 
+	// if we already have a global config we can just reload it instead of trying all locations
+	// but since gitconfig.New() already sets the global path to the globalConfigFile() one, we need to check
+	// its raw length to be sure it wasn't just the default empty config struct and was previously loaded
+	if cs.global != nil && cs.global.raw.Len() > 0 {
```
dominikschulz, line 151: Maybe we want to add something like `IsEmpty()` to the config struct? Would it make this easier to read?

internal/action/config.go
```diff
@@ -89,10 +89,29 @@ func (s *Action) setConfigValue(ctx context.Context, store, key, value string) e
 		return fmt.Errorf("storage not available")
 	}
 
-	if err := st.Add(ctx, "config"); err != nil && !errors.Is(err, istore.ErrGitNotInit) {
+	// in case of a local config change we want to track changes
+	if !st.Exists(ctx, "config") {
```
dominikschulz, line 93: This seems a little bit hacky. Are you doing this only because s.cfg.Set doesn't tell you if it wrote to a global or local config? In that case we could (should) add an additional return value to config.Set to make this easier.

AnomalRoil: Yes, I am. I'm happy adding a "iota enum" going in maybe reverse order from 0 to x in order of precedence of the configs?
Lower numbers having less precedence than higher ones. (e.g. Env would be the max value of the top of my head)

AnomalRoil: But this specific line is to avoid attempting to interact with a non existent local config file since that causes git to error out.

dominikschulz: The `Set` method is part of the internal package `internal/config`. Feel free to add a second parameter if you need it.

But could you remind me how/where you are planning to use that? Right now I fail to see how that would help this PR.

pkg/gitconfig/configs.go
AnomalRoil: Done

dominikschulz:
My goal was to support worktree configs, but it was rather low priority and at this point I don't remember if I finished that part or not.

But the documentation claims that I didn't complete it: https://pkg.go.dev/github.com/gopasspw/gopass/pkg/gitconfig#hdr-Known_limitations

internal/action/config.go
```diff
@@ -34,10 +35,10 @@ func (s *Action) Config(c *cli.Context) error {
 		return exit.Error(exit.Usage, nil, "Usage: %s config key value", s.Name)
 	}
 
-	// need to initialize sub-stores so we can update their configs.
-	// special case: we can always update the root config.
-	if err := s.IsInitialized(c); err != nil && store != "" {
-		return err
+	// sub-stores need to have been initialized so we can update their local configs.
+	// special case: we can always update the global config. NB: IsInitialized initializes the store if nil.
+	if inited, err := s.Store.IsInitialized(ctx); err != nil || store != "" && !inited {
```
dominikschulz, line 40: Could you add brackets to the if condition? My brain is struggling to remember if || or && is evaluated first.

internal/action/config.go
```diff
@@ -80,20 +81,42 @@ func contains(haystack []string, needle string) bool {
...
 	debug.Log("setting %s to %s for %q", key, value, store)
 
-	if err := s.cfg.Set(store, key, value); err != nil {
+	level, err := s.cfg.SetWithLevel(store, key, value)
+	if err != nil {
 		return fmt.Errorf("failed to set config value %q: %w", key, err)
 	}
 
+	// in case of a non-local config change we don't need to track changes in the store
+	if level != config.Local {
```
dominikschulz, line 90: If we were to add full support for worktree configs this would become wrong, I think.

Both local and worktree configs should be committed to the store, I tink.
But feel free to ignore this, I don't know when/if we will have worktree support.
And even if the gitconfig package has it there is still the question if it makes sense for gopass to expose that feature.

internal/action/config.go
```diff
@@ -80,20 +81,42 @@ func contains(haystack []string, needle string) bool {
...
+	}
+
 	st := s.Store.Storage(ctx, store)
 	if st == nil {
 		return fmt.Errorf("storage not available")
 	}
 
-	if err := st.Add(ctx, "config"); err != nil && !errors.Is(err, istore.ErrGitNotInit) {
-		return fmt.Errorf("failed to stage config file: %w", err)
+	// notice that we rely on the cfg.Set above having created the local config file here. If it doesn't exist,
```
dominikschulz, line 102: Great comment. But one idea: Should we maybe check if the file exists and gracefully return if not?

internal/action/config.go
```diff
@@ -80,20 +81,42 @@ func contains(haystack []string, needle string) bool {
...
+		debug.Log("Added local config for commit")
+	case errors.Is(err, istore.ErrGitNotInit):
+		debug.Log("Skipping staging of local config: %v", err)
+	default:
+		return fmt.Errorf("failed to stage local config file: %w", err)
 	}
-	if err := st.Commit(ctx, "Update config"); err != nil && !errors.Is(err, istore.ErrGitNotInit) {
-		return fmt.Errorf("failed to commit config: %w", err)
+
+	switch err := st.Commit(ctx, "Update config"); {
```
dominikschulz, line 113: This commit error handling (needs to) happens in quite a few places.
I wonder if we should have some kind of wrapper that handles these error cases.

Something like `git.TryAdd`, `git.TryCommit` and maybe `git.TryAddAndCommit` which would intentionally not fail if Git is not initialized or if there is nothing to commit.

internal/backend/storage/fs/store.go
```diff
@@ -151,7 +151,7 @@ func (s *Store) Exists(ctx context.Context, name string) bool {
 	}
 	path := filepath.Join(s.path, filepath.Clean(name))
 	found := fsutil.IsFile(path)
-	debug.Log("Checking if %s exists at %s: %t", name, path, found)
+	debug.Log("Checking if '%s' exists at %s: %t", name, path, found)
```
dominikschulz, line 154: nit: I think you could use `%q` instead of `'%s'`.

pkg/gitconfig/configs.go
```diff
@@ -145,6 +145,22 @@ func (cs *Configs) loadGlobalConfigs() string {
 		locs = append(locs, filepath.Join(appdir.UserHome(), cs.GlobalConfig))
 	}
 
+	// if we already have a global config we can just reload it instead of trying all locations
+	if !cs.global.IsEmpty() {
+		if p := cs.global.path; p != "" {
```
dominikschulz, line 150: super-nit: I think you could combine both if statements to remove a level of nesting: `if p := cs.global.path; !cs.global.IsEmpty() && p != ""` (not sure if that's much better).

internal/action/config.go
AnomalRoil: That's what I was doing earlier, happy to add it back.

internal/action/config.go
AnomalRoil: This is an easy fix, adding it

internal/action/config.go
AnomalRoil: I'd prefer to keep such a change for another PR I think

internal/backend/storage/fs/store.go
AnomalRoil: Nice, I never use `%q`, I should.

pkg/gitconfig/configs.go
dominikschulz: Oh. Yes, then you are right of course.

internal/action/config.go
dominikschulz: Sure thing!

dominikschulz (approved):
LGTM. Thanks!

PR #2718, opened 2023-11-21
[BUGFIX] Allow to move shadowed entries into their own folder

### LINKED ISSUE #892: [FR] gopass move to rename/move directories
### Summary
Currently `gopass move` only supports moving password entries around. Is it possible to also implement moving whole subpaths/directories around? For git backend this is as simple as `git mv && git commit && git push`, which is my workaround for now.

AnomalRoil:
This should fix #892 hopefully!

### DIFF
--- a/internal/store/root/move.go
+++ b/internal/store/root/move.go
@@ -42,14 +42,7 @@ func (r *Store) move(ctx context.Context, from, to string, del bool) error {
 	subFrom, fromPrefix := r.getStore(from)
 	subTo, _ := r.getStore(to)
 
-	srcIsDir := r.IsDir(ctx, from)
-	dstIsDir := r.IsDir(ctx, to)
-
-	if srcIsDir && r.Exists(ctx, to) && !dstIsDir {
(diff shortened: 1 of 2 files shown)


internal/store/root/move.go
```diff
@@ -124,13 +117,21 @@ func (r *Store) move(ctx context.Context, from, to string, del bool) error {
...
+
+	if srcIsDir && r.Exists(ctx, to) && !dstIsDir {
+		return fmt.Errorf("destination is a file")
+	}
+
 	entries := []string{from}
 	// if the source is a directory we enumerate all it's children
 	// and move them one by one.
-	if r.IsDir(ctx, from) {
+	if srcIsDir {
```
AnomalRoil, line 134: This was the culprit, but since we didn't really use `moveFromTo` anywhere else, I figured I could as well change its signature since these two variables weren't being used above in `move`

dominikschulz (approved):
LGTM. Thank you!

PR #2723, opened 2023-11-24
Add storage.TryAdd and TryCommit methods

dominikschulz:
This PR adds new Try methods to the RCS backends. Each of the new methods will attempt to Add, Commit or Push but ignore any non-critical errors.

This simplifies the call sites since they no longer have to have lenghty error handling.

### DIFF
--- a/internal/backend/storage/fossilfs/fossil.go
+++ b/internal/backend/storage/fossilfs/fossil.go
@@ -207,6 +207,21 @@ func (f *Fossil) Add(ctx context.Context, files ...string) error {
 	return f.Cmd(ctx, "fossilAdd", args...)
 }
 
+// TryAdd adds the listed files to the fossil index.
+func (f *Fossil) TryAdd(ctx context.Context, files ...string) error {
+	err := f.Add(ctx, files...)
+	if err == nil {
+		return nil
+	}
+	if err == store.ErrGitNotInit {
+		debug.Log("Fossil not initialized. Ignoring.")
+
+		return nil
+	}
+
+	return err
+}
+
 // HasStagedChanges returns true if there are any staged changes which can be committed.
 func (f *Fossil) HasStagedChanges(ctx context.Context) bool {
 	s, err := f.getStatus(ctx)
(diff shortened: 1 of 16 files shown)


internal/backend/storage/fossilfs/fossil.go
```diff
@@ -293,6 +328,27 @@ func (f *Fossil) Push(ctx context.Context, remote, branch string) error {
...
+		return nil
+	}
+
+	switch err {
+	case store.ErrGitNotInit:
+		debug.Log("Git not initialized. Ignoring.")
+
+		return nil
+	case store.ErrGitNoRemote:
+		debug.Log("Git has no remote. Ignoring.")
```
AnomalRoil, line 344: I guess these should read more like:
```suggestion
		debug.Log("Fossil not initialized. Ignoring.")

		return nil
	case store.ErrGitNoRemote:
		debug.Log("Fossil has no remote. Ignoring.")
```

And I'm not sure if the `store.ErrGitNoRemote` and `ErrGitNotInit` actually match any of the Fossil errors?

internal/backend/storage/fossilfs/fossil.go
```diff
@@ -251,6 +266,26 @@ func (f *Fossil) Commit(ctx context.Context, msg string) error {
...
+	if err == nil {
+		return nil
+	}
+	if err == store.ErrGitNothingToCommit {
+		debug.Log("Nothing to commit. Ignoring.")
+
+		return nil
+	}
+	if err == store.ErrGitNotInit {
+		debug.Log("Git not initialized. Ignoring.")
```
AnomalRoil, line 281: Same as below, not sure how Fossil errors match our `Git` errors?
```suggestion
	if err == store.ErrGitNothingToCommit {
		debug.Log("Nothing to commit. Ignoring.")

		return nil
	}
	if err == store.ErrGitNotInit {
		debug.Log("Fossil not initialized. Ignoring.")
```

AnomalRoil (approved):
Beside the Linter issues and the Fossil errors and logs, LGTM

internal/backend/storage/fossilfs/fossil.go
dominikschulz: The errors are oddly named (i.e. we should rename them) but the semantics match Fossil, e.g. as Fossil repo can be uninitialized or there could be nothing to commit ....

AnomalRoil (approved):
LGTM

PR #2724, opened 2023-11-25
[BUGFIX] Try to always honor local config for mounts

### LINKED ISSUE #2721: `core.notifications = false` not working in mounts
> Not only this, but I have `core.notifications = false` set, but I've started getting notifications (clipboard copy, clear, etc)
>
>_Originally posted by @tjex in https://github.com/gopasspw/gopass/issues/2673#issuecomment-1779264183_
            

I have confirmed this to be the case for mounts: it seems we are currently not properly handling substore and their config when we are fetching config FromContext.

AnomalRoil:
I realize this is a big PR, but sadly the more I was digging to fix #2721, the more I would uncover new edge cases around our local vs global config handling 😢 

This:
 - adds a new `config.WithMount` that allows to pass to a context the current "mount point" or "store" that we're working on.
 - changes `config.FromContext` to also return the mount point if any is provided in the context.
 - changes `config.Int`, `Bool` and `String` to take into account the mount point if any is provided.
 - attempts to uses `WithMount` in the relevant places, mostly: `Show` and `Generate` are concerned by custom config options, as well as anything that can use the clipboard feature.

Furthermore, in "config related" changes, this also:
 - documents the config options that aren't supported at the local level.
 - prevents local configs from tampering with security features such as `recipients.hash`.
 - fixes `recipients ack` to actually update the `recipients.hash` when it was previously outdated, but the `.gpg-id` were correct.
 - adds a few convenience functions to our internal `config` package.

In the "weird changes" I still did in here, this also:
 - fixes a few typos.
 - aligns the use of `<root>'  to indicate the local root store when using the `--store` flag even in `sync`.
 - fixes a race condition in `rcs_test.go` where `git` was weirdly failing with `failed to add "/tmp/TestInitRCS2737126547/001/git" to git: exit status 128: fatal: Unable to create '/home/anomalroil/code/gopass/.git/index.lock.lock': File exists.` because of some Parallel tests. This might actually indicate a bigger underlying issue: I would expect our tests not to tamper with the local git things, and do everything in `/tmp`....

Let me know if you'd like me to chunk it into multiple PR / commits, it just was a big rabbit hole and it kept on giving more.

This is probably not entirely done yet, e.g. I don't think `core.autopush` is properly respected at the local level yet, might be worth opening a new issue to track a big review of all our config options?


internal/config/context.go
```diff
@@ -3,28 +3,42 @@ package config
...
 
+	"github.com/gopasspw/gopass/pkg/debug"
 	"github.com/gopasspw/gopass/pkg/gitconfig"
 )
 
 type contextKey int
 
 const (
 	ctxKeyConfig contextKey = iota
+	ctxMountPoint
```
dominikschulz, line 14: nit: `ctxKeyMountPoint`

dominikschulz (changes requested):

dominikschulz (approved):

dominikschulz:
Amazing, thanks a lot!

PR #2727, opened 2023-11-25
[bugfix] Fix writes to global config from tests

### LINKED ISSUE #2725: Running `go test` for gopass modifies global config and also removes mounts
### Summary
If you have a mounted substore, e.g. "test":
```
gopass (/home/anomalroil/.local/share/gopass/stores/root)
└── test (/home/anomalroil/.local/share/gopass/stores/test)
``` 
and you run `go test ./...` somehow your config is changed and you'll loose your mount and will have to mount it again.

### Steps To Reproduce
```
$ gopass init --store test
$ gopass mounts
$ git clone https://github.com/gopasspw/gopass && cd gopass
$ go test ./... -count=1
$ gopass mounts
```

### Expected behavior
Running the tests should remain contained to some kind of `/tmp` files and not affect the global developer config for `gopass`.

### Environment
- OS: Ubuntu 22.04
- OS version: `6.5.6-76060506-generic #202310061235~1697396945~22.04~9283e32 SMP PREEMPT_DYNAMIC Sun O x86_64 x86_64 x86_64 GNU/Linux`
- gopass Version: master tip
- Installation method:  from source

dominikschulz:
Fixes #2725


AnomalRoil (changes requested):
Seems there are a few issues on Windows

Nit: maybe the `gptest.NewUnitTester(t)` could return a context too to avoid having to add it to all tests always?

And I guess using `ReadOnly` configs might bite us when testing the config options, since `ReadOnly` doesn't allow to change values even in memory, right?

AnomalRoil (commented):
I think now such tests aren't actually testing what we think they are testing:
https://github.com/gopasspw/gopass/pull/2727/files#diff-2bf3cbedfee37a38c52d5f08662f0b4b441ce2a1d1fa915c18e348d167d106cbL67-R72

And we do set config options in a few tests:
```
internal/action/generate_test.go
41:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "false"))
159:	require.NoError(t, act.cfg.Set("", "generate.autoclip", ov))
161:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "true"))
172:	require.NoError(t, act.cfg.Set("", "generate.autoclip", ov))
174:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "true"))

internal/action/copy_test.go
30:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "false"))
113:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "false"))

internal/action/create_test.go
29:	require.NoError(t, act.cfg.Set("", "core.notifications", "false"))
30:	require.NoError(t, act.cfg.Set("", "core.cliptimeout", "1"))

internal/action/find_test.go
33:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "false"))
68:	require.NoError(t, act.cfg.Set("", "show.safecontent", "true"))
88:	require.NoError(t, act.cfg.Set("", "show.safecontent", "false"))

internal/action/show_test.go
70:	require.NoError(t, act.cfg.Set("", "show.safecontent", "true"))
143:	require.NoError(t, act.cfg.Set("", "show.safecontent", "false"))
418:	require.NoError(t, act.cfg.Set("", "domain-alias.foo.de.insteadOf", "foo.com"))

internal/action/insert_test.go
29:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "true"))
87:	require.NoError(t, act.cfg.Set("", "show.safecontent", "true"))
116:	require.NoError(t, act.cfg.Set("", "show.safecontent", "false"))
139:	require.NoError(t, act.cfg.Set("", "generate.autoclip", "false"))
``` 

and the fact that moving to a `ReadOnly` config didn't break any of these tests probably mean they aren't testing useful things?

AnomalRoil (changes requested):
I think the `ReadOnly` naming is misleading here since it really is a `noWrites` one, given our gitconfig actually does support readonly configs.

internal/config/config.go
```diff
@@ -79,12 +79,17 @@ func New() *Config {
 	return c
 }
 
-// NewNoWrites initializes a new config that does not allow writes. For use in tests.
+// NewReadOnly initializes a new config that does not allow writes. For use in tests.
 // This does not migrate legacy option names to their correct config section.
-func NewNoWrites() *Config {
+func NewReadOnly() *Config {
 	return newWithOptions(true)
```
AnomalRoil, line 85: This probably explains why the tests are all still passing: it's not really a readonly config as we mean it in our gitconfig here:

https://github.com/gopasspw/gopass/blob/d168602f770b00eb2d531ae6337c3c316ee8299c/pkg/gitconfig/config.go#L21-L22

It's actually a `Config` with `noWrites` set (that's what `newWithOptions(noWrites bool)` does) and so it's not really `ReadOnly`.

main_test.go
```diff
@@ -176,8 +175,8 @@ func testCommands(t *testing.T, c *cli.Context, commands []*cli.Command, prefix
 func TestInitContext(t *testing.T) {
 	t.Parallel()
 
-	ctx := context.Background()
-	cfg := config.NewNoWrites()
+	ctx := config.NewContextReadOnly()
+	cfg := config.NewReadOnly()
```
AnomalRoil, line 179: The context is supposedly set to the provided config below in initContext in our main, so keeping a plain one is fine here.

pkg/pwgen/pwrules/aliases_test.go
```diff
@@ -22,7 +21,7 @@ func TestLoadCustomRules(t *testing.T) {
 		require.NoError(t, cfg.Set("", "domain-alias."+k+".insteadOf", v))
 	}
 
-	ctx := context.Background()
+	ctx := config.NewContextReadOnly()
 	ctx = cfg.WithConfig(ctx)
```
AnomalRoil, line 25: Here we're using the above config in our context anyway, so a plain bg context is fine.

AnomalRoil (changes requested):
a few more comments

internal/config/config.go
dominikschulz: Shall we revert this rename?

AnomalRoil: Yeah, probably

PR #2731, opened 2023-11-28
[feature] Allow setting autosync.interval in minutes with "m" suffix

quite:
Solves https://github.com/gopasspw/gopass/issues/2730https://github.com/gopasspw/gopass/issues/2730


quite:
@dominikschulz what do you think? This does not work entirely, because the `return`s upon failed parsing does not actually cause the program to exit!

internal/action/sync.go
```diff
@@ -69,13 +73,22 @@ func (s *Action) autoSync(ctx context.Context) error {
 
 	ls := s.rem.LastSeen("autosync")
 	debug.Log("autosync - last seen: %s", ls)
-	syncInterval := autosyncIntervalDays
+	syncInterval := autosyncIntervalMinutes
 
-	if s.cfg.IsSet("autosync.interval") {
-		syncInterval = s.cfg.GetInt("autosync.interval")
+	if intervalStr := s.cfg.Get("autosync.interval"); intervalStr != "" {
+		factor := dayInMinutes
```
dominikschulz, line 79: This looks a bit too complicated. I would suggest we just use [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) unless it's a plain number. In that case we must handle it as if it is a day to keep backwards compatibility.

Maybe like this:

```
if !strings.ContainsAny(intervalStr, "hmsd") {
  intervalStr += "d"
}
syncInterval = time.ParseDuration(intervalStr)
```

dominikschulz (changes requested):

internal/action/sync.go
quite: Quite nice, I did not know about `time.ParseDuration`. But I see that it does not support `d` for day. So if we should do it this way, I guess we need a little helper that adds support for `d`, and suffix-less for backwards compat. What do you say?

dominikschulz: Ahw, you are right. We could use something like this maybe? https://github.com/xhit/go-str2duration

quite: We could, I've pushed another commit attempting that method. I guess it is more featureful, but adding another dependency counts for quite a lot of added complexity to me. But maybe it's fine?

internal/action/sync.go
```diff
@@ -20,11 +21,12 @@ import (
...
 )
 
 var (
-	autosyncIntervalDays = 3
-	autosyncLastRun      time.Time
+	autosyncInterval = time.Duration(3*24) * time.Hour
+	autosyncLastRun  time.Time
 )
 
 func init() {
```
quite, line 32: Should the GOPASS_AUTOSYNC_INTERVAL be given the same treatment? Or we dont care since its deprecated?

internal/action/sync.go
```diff
@@ -69,13 +71,20 @@ func (s *Action) autoSync(ctx context.Context) error {
...
 
-	if s.cfg.IsSet("autosync.interval") {
-		syncInterval = s.cfg.GetInt("autosync.interval")
+	if intervalStr := s.cfg.Get("autosync.interval"); intervalStr != "" {
+		if _, err := strconv.Atoi(intervalStr); err == nil {
+			intervalStr += "d"
+		}
+		duration, err := str2duration.Str2Duration(intervalStr)
+		if err != nil {
+			return exit.Error(exit.Usage, err, "failed to parse autosync.interval %q: %q", intervalStr, err)
```
quite, line 82: This causes no error, because it is not cared for at https://github.com/gopasspw/gopass/blob/master/internal/action/init.go#L44  -- what to do about it?

AnomalRoil: Maybe output a warning using `out.Warn` in case of an error?

quite: But this implies a syntax error with the autosync.interval the config file. And what autosync.interval should we then continue running with?

dominikschulz: As an invalid autosync interval shouldn't be a fatal error we should continue running with a sensible default (e.g. 3 days).

quite: I've pushed a commit that falls back to the default. Other review comments remain...

internal/action/sync.go
dominikschulz: Would be nice to handle it in the same way, but not strictly necessary since it's deprecated.

quite: Let's not do that, forgoing complexity.

dominikschulz (approved):
LGTM. Thank you!

PR #2748, opened 2023-12-21
[cleanup] Add better logging in case no owner key is found

dominikschulz:
See gopasspw/gopass-jsonapi#118

### DIFF
--- a/internal/store/leaf/recipients.go
+++ b/internal/store/leaf/recipients.go
@@ -281,6 +281,7 @@ func (s *Store) OurKeyID(ctx context.Context) string {
 	}
 
 	debug.Log("WARNING: no owner key found in %v", recp)
+	out.Warning(ctx, "No owner key found. Make sure your key is trusted by GPG")
 
 	return ""
 }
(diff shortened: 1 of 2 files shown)


internal/store/leaf/recipients.go
```diff
@@ -270,7 +270,8 @@ func (s *Store) ensureOurKeyID(ctx context.Context, rs []string) []string {
 // OurKeyID returns the key fingprint this user can use to access the store
 // (if any).
 func (s *Store) OurKeyID(ctx context.Context) string {
-	for _, r := range s.Recipients(ctx) {
+	recp := s.Recipients(ctx)
```
innovate-invent, line 273: Can you add a check in the GPG Recipients() to specifically call out "make sure your key is trusted by gpg". This will save people from a rabbit hole of figuring out why this error occurs.

internal/store/leaf/recipients.go
```diff
@@ -279,6 +280,9 @@ func (s *Store) OurKeyID(ctx context.Context) string {
 		return kl[0]
 	}
 
+	debug.Log("WARNING: no owner key found in %v", recp)
+	out.Warning(ctx, "No owner key found. Make sure your key is trusted by GPG")
```
innovate-invent, line 284: Thanks for adding this, but you are mentioning GPG outside of the GPG specific code. This message would appear for all storage drivers.

dominikschulz: Ack. Thank you. I have reworded this a bit, but so far haven't found a better place to put it without doing some ugly changes that I don't want to do for this (the risk of breaking stuff here is just too high).

innovate-invent: I see. I guess the current commit is good then.

PR #2765, opened 2024-01-14
upgrade to ergochat/readline@v0.1.0

slingamn:
I just tagged https://github.com/ergochat/readline/releases/tag/v0.1.0 which has a minor API break. Here's the changelog:

https://github.com/ergochat/readline/blob/a16c9c0b4a38ce80181b7e716074be2532b94493/docs/CHANGELOG.md

Thanks for your patience! Going forward from here, I don't anticipate any more API breaks that would affect gopass :-)


dominikschulz (approved):
LGTM. Thank you!

Issue #2806, opened 2024-02-19
Feature Request: Ignoring passwords in the audit feature


ltecheroffical:
Let's say you have passwords stored but you have passwords you have no control of (for example, your family's streaming accounts) and run a pass audit. Because of that, you have all these warnings about passwords that you cannot change. Here is what I suggest:

We can fix this using a file named `.gopass-audit-ignore` in the password store folder. Every time a scan is run, then Gopass should check that file to see if to exclude it and, if so, ignore it with regex syntax. Here is an example of an example `.gopass-audit-ignore`following these rules:

```
# Lines starting with # get ignored
Ignore_Everything_In_Here/.*
test_folder/ignore_this
```

Hopefully this is detailed enough!
(I tried to add it myself until I found out I needed to learn the code base more)

leogott:
A similar feature would be useful to me, however I propose a slightly simpler solution:

A marker-file (e.g. `.gopass-audit-ignore`) included in any subdirectory of a store could make gopass audit report any file in the same dir and nested dirs as skipped.

It might be better though, to only skip specific checks, per secret, and explicitly so
(
- a shared streaming account might have a weak password which rarely changes, but you'd still want to know about a leaked password
- if you control the streaming password, you *absolutely* want to check for uniqueness
- a SIM-pin or similar will always fail most password-quality and haveibeenpowned checks, but maybe you still want to change it once a year or check for unique-ness
)

ltecheroffical:
> A similar feature would be useful to me, however I propose a slightly simpler solution:
> 
> A marker-file (e.g. `.gopass-audit-ignore`) included in any subdirectory of a store could make gopass audit report any file in the same dir and nested dirs as skipped.
> 
> It might be better though, to only skip specific checks, per secret, and explicitly so (
> 
>     * a shared streaming account might have a weak password which rarely changes, but you'd still want to know about a leaked password
> 
>     * if you control the streaming password, you _absolutely_ want to check for uniqueness
> 
>     * a SIM-pin or similar will always fail most password-quality and haveibeenpowned checks, but maybe you still want to change it once a year or check for unique-ness
>       )

That's a lot less flexible then Regex filtering

PR #2822, opened 2024-03-10
[feat] Add .gopass-audit-ignore support to ignore secrets from audits

dominikschulz:
[feat] Add .gopass-audit-ignore support to ignore secrets from audits
This PR adds a new exclude file that is used during gopass audit to
ignore entries from auditing. The file itself is using RE2 syntax.

Fixes https://github.com/gopasspw/gopass/issues/2806


ltecheroffical:
Thanks!

PR #2840, opened 2024-03-25
[chore]: linting

ldez:
Fix the CI https://github.com/gopasspw/gopass/actions/runs/8400385803/job/23007586512


dominikschulz (approved):
LGTM. Thanks a lot. I was tempted to wait a bit before dropping support for older Go versions but this is such a nice cleanup and we don't promise to support older versions.

ldez:
FYI by defining `go 1.22.0` inside the `go.mod`, you already dropped the older Go versions.

Since go1.21, the version is a hard requirement for the min Go version. (If you try to compile with go1.21, go1.22 will be automatically downloaded and used)

If you want to allow older versions, you should define a lower version inside your `go.mod`.

PR #2841, opened 2024-03-25
[chore] use the same version of golangci-lint

ldez:
There are 2 different version of golangci-lint used inside the project, which lead to CI failure: https://github.com/gopasspw/gopass/actions/runs/8425200501/job/23070742289

https://github.com/gopasspw/gopass/blob/b8f0ff8ff2570963b534d70961919d638c6a763b/Makefile#L136

https://github.com/gopasspw/gopass/blob/b8f0ff8ff2570963b534d70961919d638c6a763b/.github/workflows/golangci-lint.yml#L46


dominikschulz (approved):
Thank you, again 😃

PR #2866, opened 2024-04-10
chore: Update dependencies

dominikschulz:


go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/ProtonMail/go-crypto v1.0.0
 	github.com/atotto/clipboard v0.1.4
 	github.com/blang/semver/v4 v4.0.0
-	github.com/caspr-io/yamlpath v0.0.0-20200722075116-502e8d113a9b
+	github.com/caspr-io/yamlpath v0.0.1
```
AnomalRoil, line 10: Notice that the v0.0.1 has no licence and is missing ~10 commits compared to the latest master.
Maybe we should pin the latest one?

AnomalRoil (changes requested):
either pin master for that dep or add an exception to our LICENSE-LINT I guess.

PR #2907, opened 2024-07-07
Updating our linter and fixing the current linter issues

AnomalRoil:
This should fix our last linter issues in our latest builds that makes it look like our build are failing when they aren't.


dominikschulz (approved):
Thanks!

PR #2946, opened 2024-09-20
feat: Add --qrbody flag to encode the entire body of a secret into a QR

dominikschulz:
code

This comes in handy when storing Wireguard configs in gopass as it allows to import the entire config on a mobile device easily.


AnomalRoil (approved):
LGTM, pretty straightforward.

Issue #2650, opened 2023-09-06
[cleanup] Math/rand.Seed deprecated


n4x2:
### Summary
Since Go 1.20 release, math/rand.Seed is marked "Deprecated"

Looks like five instances to consider updating:
```sh
find . -name "*.go" -type f -print | xargs grep .Seed
```
```sh
./pkg/pwgen/pwgen_test.go:      mrand.Seed(1789)
./pkg/pwgen/rand.go:    rand.Seed(time.Now().Unix() + int64(os.Getpid()+os.Getppid()))
./pkg/fsutil/fsutil.go: rand.Seed(time.Now().UnixNano())
./pkg/gitconfig/config_test.go: rand.Seed(time.Now().Unix())
./internal/action/binary_test.go:       rand.Seed(42)
```

dominikschulz:
Usually "Deprecated" just means it's finally stable ;)

But I'll check what the recommendation is these days.

dominikschulz:
* `pkg/pwgen/rand.go` This one can Go away once we don't need to support Go 1.20 anymore.
* `pkg/fsutil/fsutil.go` Same here.

For the test files we actually rely on detemernistic output, so we will need some more changes. We need to introduce a global variable with a local generator and seed it like this `New(NewSource(seed))`. See https://pkg.go.dev/math/rand#Seed

orangekame3:
@dominikschulz 
Hi, I'd like to take on this issue if that's okay.

dominikschulz:
@orangekame3 Feel free to do so

orangekame3:
@dominikschulz 
Thank you. I've created a PR. Please check it out:
https://github.com/gopasspw/gopass/pull/2675

lhardt:
Hello @dominikschulz ! 

I see the last pull request on this issue is from last year, but I see the  `good-first-issue` tag. 
Do you mind if I open my own PR?

AnomalRoil:
@lhardt Please go ahead. It should be as simple as taking the previous PR and fixing the merge conflicts and then doing the changes that were requested in there if you want. 
Or you can try and do it your way too if you have another idea to do it.

lhardt:
@dominikschulz  Can this be marked as closed following #2953?
I guess #2873  can be closed following #2954 too

dominikschulz:
Fixed by #2953

PR #2950, opened 2024-09-27
[DEPRECATION] Remove references to deprecated rand.Seed (#2650)

lhardt:
Remove references to deprecated rand.Seed (#2650).

### Context:

- From Go 1.20 (Feb 2023) onwards, `math/rand.Seed` is deprecated, and `math/rand` is already seeded by default.
- From Go 1.24 (Feb 2025?) onwards, `math/rand.Seed` [will be a no-op](https://github.com/golang/go/blob/master/src/math/rand/rand.go), breaking the test `TestPwgenNoCrand`. 

See method documentation in the link above for further info.

### Code Changes:

-  We must have our own `Rand` variable, so that its value can be injected in tests;
-  I changed the test name of `TestPwgenNoCrand` to `TestPwgenNoCrandFallback`, to make the expected behaviour clearer;
- `Shred` does not need to seed, nor does `TestLoadFromEnv` in config_test,  
  - But note that `TestPwgenNoCrandFallback` must restore the randomness so that `TestLoadFromEnv` can work properly.

This also removes the side-effect that fsutil's Shred was overwriting the seed set by pwgen.


pkg/pwgen/pwgen_test.go
```diff
@@ -2,13 +2,14 @@ package pwgen
 
 import (
 	"bytes"
-	"crypto/rand"
+	crand "crypto/rand"
 	"fmt"
 	"io"
-	mrand "math/rand"
```
lhardt, line 8: This choice was questioned on orangekame's PR. 

Howver, this aligns `pwgen_test.go` with the style found in `rand.go`, of the same package.

AnomalRoil (approved):
LGTM, thanks!

dominikschulz (approved):

lhardt:
Hi! Done the linting and signoff fixes. Since this is so close to October, would you mind if I reopen the PR on Tuesday?

dominikschulz:
@lhardt No problem at all. See you in #hacktober :)

PR #2953, opened 2024-09-30
[DEPRECATION] Remove references to deprecated rand.Seed

lhardt:
**(Note: reopened this PR after creating and squashing `gofumpt` commit)**

Remove references to deprecated rand.Seed (#2650).

### Context:

- From Go 1.20 (Feb 2023) onwards, `math/rand.Seed` is deprecated, and `math/rand` is already seeded by default.
- From Go 1.24 (Feb 2025?) onwards, `math/rand.Seed` [will be a no-op](https://github.com/golang/go/blob/master/src/math/rand/rand.go), breaking the test `TestPwgenNoCrand`. 

See method documentation in the link above for further info.

### Code Changes:

-  We must have our own `Rand` variable, so that its value can be injected in tests;
-  I changed the test name of `TestPwgenNoCrand` to `TestPwgenNoCrandFallback`, to make the expected behaviour clearer;
- `Shred` does not need to seed, nor does `TestLoadFromEnv` in config_test,  
  - But note that `TestPwgenNoCrandFallback` must restore the randomness so that `TestLoadFromEnv` can work properly.

This also removes the side-effect that fsutil's Shred was overwriting the seed set by pwgen.

### DIFF
--- a/pkg/fsutil/fsutil.go
+++ b/pkg/fsutil/fsutil.go
@@ -10,7 +10,6 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
-	"time"
 
 	"github.com/gopasspw/gopass/pkg/appdir"
 	"github.com/gopasspw/gopass/pkg/debug"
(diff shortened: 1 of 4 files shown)


lhardt:
As mentioned above, this reopens the PR #2950, with `gofumpt` changes.

dominikschulz (approved):

lhardt:
The DCO failure message is:
```Expected "Leo Hardt ..." , but got "Léo Hardt ...".``` -- Can it be ignored or would you like me rebase and signoff again?

dominikschulz:
@lhardt I'm sorry, the check doesn't seem to handle non-ASCII nicely. Let's just ignore the check for now.

Issue #2873, opened 2024-04-26
Trailing whitespace removed when copying password to clipboard


leandrolanzieri:
### Summary
Hi there. I just found that whenever a secret has a trailing whitespace, copying as `gopass show -c test_secret` doesn't copy the whitespace. Showing the secret without copying prints the secret correctly. Also, intermediate whitespaces work fine.
Is this known / intentional? Perhaps having a trailing whitespace is not an expected use case.

### Steps To Reproduce
Create a new secret with a trailing whitespace (e.g., `supersecure `). Try copying with `gopass show -c test_secret`.

### Expected behavior
`supersecure ` ends up on the clipboard.

### Environment

- OS: `Manjaro Linux`
- OS version: `Linux 5.10.211-1-MANJARO #1 SMP PREEMPT Fri Mar 1 16:55:20 UTC 2024 x86_64 GNU/Linux`
- gopass Version: `gopass 1.15.11 go1.21.4 linux amd64`
- Installation method: `yay`

dominikschulz:
Correct. We currently do not support passwords with leading or trailing whitespaces: https://github.com/gopasspw/gopass/blob/master/pkg/gopass/secrets/akv.go#L253

lhardt:
Since key-value pairs are only `key: value`, with a single space (see `kvSep`), it must be fine to remove `TrimSpace` and just remove the first character instead. Do you mind if I check in a PR for that?

lhardt:
Done :)  

See PR #2954

leandrolanzieri:
Great to see this in!

PR #2954, opened 2024-10-01
[ENHANCEMENT] Allow for whitespace-trailing passwords (#2873)

lhardt:
### Summary

- Adding support for whitespace-trailed passwords. This means blocking `AKV` from `Trim`ming the first line.
- Adding support for whitespace-trailed values on AKV. 

### Technical Notes

- The `TrimSpace` on `ParseAKV` only applies to Keys, not Values. So keys must be trimmed, but values are taken as-is.

### DIFF
--- a/pkg/gopass/secrets/akv.go
+++ b/pkg/gopass/secrets/akv.go
@@ -260,15 +260,13 @@ func ParseAKV(in []byte) *AKV {
 			continue
 		}
 
-		line = strings.TrimSpace(line)
-
 		key, val, found := strings.Cut(line, kvSep)
 		if !found {
 			continue
 		}
 
 		key = strings.TrimSpace(key)
-		val = strings.TrimSpace(val)
+		// val = strings.TrimSpace(val)
 		// we only store lower case keys for KV
 		a.kvp[key] = append(a.kvp[key], val)
 	}
--- a/pkg/gopass/secrets/akv.go
+++ b/pkg/gopass/secrets/akv.go
@@ -250,7 +250,7 @@ func ParseAKV(in []byte) *AKV {
 
 		// handle the password that must be in the very first line
 		if first {
-			a.password = strings.TrimSpace(line)
+			a.password = line
 			first = false
 
 			continue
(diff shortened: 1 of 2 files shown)


pkg/gopass/secrets/akv.go
```diff
@@ -260,15 +260,13 @@ func ParseAKV(in []byte) *AKV {
 			continue
 		}
 
-		line = strings.TrimSpace(line)
```
lhardt, line 263: The components of the line were going to be `trim`med anyway.

pkg/gopass/secrets/akv.go
```diff
@@ -260,15 +260,13 @@ func ParseAKV(in []byte) *AKV {
...
-
 		key, val, found := strings.Cut(line, kvSep)
 		if !found {
 			continue
 		}
 
 		key = strings.TrimSpace(key)
-		val = strings.TrimSpace(val)
```
lhardt, line 271: Should I add a more detailed comment of why val wasn't `Trim`med? It would seem arbitrary otherwise.

Do  you normally reference GitHub issues on comments?

pkg/gopass/secrets/akv_test.go
```diff
@@ -121,6 +121,24 @@ func TestSetKeyValuePairToEmptyAKV(t *testing.T) {
 	assert.Equal(t, "bar", v)
 }
 
+func TestAKVTrailingWhitespace(t *testing.T) {
+	t.Parallel()
+	// Expected behaviour is KEY: VALUE, with one space.
+	// Fallback should exist for KEY:VALUE, with no spaces.
+	mlValue := `foobar
```
AnomalRoil, line 128: I think the issue initially was about passwords surrounded by spaces, no just key-values pairs, no?
How does that work for passwords now? A test would be nice.

pkg/gopass/secrets/akv.go
dominikschulz: Remove the dead code and add a reference to the GH Issue, please.

pkg/gopass/secrets/akv_test.go
dominikschulz: Correct, I think the issue was about whitespace in passwords.
Being able to handle them in KV pairs is very nice as well, but either we need to change the description of the PR to match the implementation or add support for whitespace in passwords as well.

lhardt: I had a misunderstanding between KV pairs and the password itself.  But it should also be simple to not `Trim` the password. But I also see no harm in leaving values trimmed as well.  Do you mind if I do both on the same PR?

PR #2960, opened 2024-10-02
[FEATURE] Adding support for `age.Plugin` identities

### LINKED ISSUE #2260: [FR] age with YubiKeys
[age plugin support](https://github.com/FiloSottile/age/releases/tag/v1.1.0-rc.1) is coming along nicely and I'd love to have gopass ready and tested with YKs once 1.1.0 final is released.

Will try to give it a try to iron out potential issues and write down any specifics in our docs.

### LINKED ISSUE #2619: Non-interactive setup
I am working with ansible playbooks to automatically setup machines with gopass, is there a way to do

```
gopass setup --crypto=age --storage=fs 
```

Non-interactively, providing the passphrase as a parameter?

### LINKED ISSUE #2900: age plugin support
### Summary
[age v1.2.0](https://github.com/FiloSottile/age/releases/tag/v1.2.0) got released, exposing a plugin package providing an age plugin client:

> The Go module now exposes a plugin package that provides an age plugin client. That is, Recipient and Identity implementations that invoke a plugin binary, allowing the use of age plugins in Go programs.

gopass should be able to make use of this.

Already mentioned by @dominikschulz in https://github.com/gopasspw/gopass/issues/2059:

> Once age v1.1.0 comes with plugin support I'd like to support that, too. So people can use their hardware tokens (at least YubiKeys) with gopass+age.

AnomalRoil:
Sorry for this being a big, messy PR; I had to touch quite a few things and the `age` plugin support itself is still experimental, so this is currently based on the [age/plugin feature branch from age](https://github.com/FiloSottile/age/pull/580).

On the bright side: yubikey `age` identities do work really nicely as far as I can tell! 

Closes #2900, closes #2260, and fixes partially #2619.

A few points worth mentioning that might warrant some discussion:
- most notably, I had to edit the format we use for storing identities in our key ring, because plugin identities do not allow to necessarily derive the right recipient easily (or include secret key material when doing so), so I am now storing in our identity file the identities in the following format:
   ```
   AGE-PLUGIN-YUBIKEY-1GK...|age1yubikey1qt2r...
   AGE-SECRET-KEY-10CW...
   ``` 
	using `|` to separate the actual identity encoding from the recipient encoding. This has some trickling effects on how we parse age identities
- I had to add the notion of `wrapped` identities and recipients, because plugin.`Identity` and `Recipient` don't have any easy way to go back to their "encoded" string format with the current proposed `age` plugin interfaces. 
- I had to add a sorting function on the recipients when encrypting with age, because `age` by default tries to decrypt in the order of the stanzas, and if we do have a native identity on our machine, we probably want to use that first before using a hardware token.

On a side node, overall, the `age` backend code is fairly messy and it might warrant some big refactoring at some point.

Disclaimer: I have only tested that on my own machine, not with a team using `age` as a backend, and yeah tests are somewhat lacking, I'll try to add a few tests with example identities and recipients in the coming days.

### DIFF
--- a/internal/backend/crypto/age/identities.go
+++ b/internal/backend/crypto/age/identities.go
@@ -1,10 +1,13 @@
 package age
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"errors"
 	"fmt"
+	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 	"sort"
@@ -12,12 +15,39 @@ import (
 	"time"
 
 	"filippo.io/age"
+	"filippo.io/age/agessh"
+	"filippo.io/age/plugin"
 	"github.com/gopasspw/gopass/pkg/appdir"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
 )
 
 var idRecpCacheKey = "identity"
+
+type wrappedIdentity struct {
+	id       age.Identity
+	rec      age.Recipient
+	encoding string
+}
+
+func (w *wrappedIdentity) String() string           { return w.encoding }
+func (w *wrappedIdentity) SafeStr() string          { return w.encoding[:12] }
+func (w *wrappedIdentity) Recipient() age.Recipient { return w.rec }
+
+func (w *wrappedIdentity) Unwrap(stanzas []*age.Stanza) ([]byte, error) {
+	return w.id.Unwrap(stanzas)
+}
+
+type wrappedRecipient struct {
+	rec      age.Recipient
+	encoding string
+}
+
+func (w *wrappedRecipient) String() string { return w.encoding }
+
+func (w *wrappedRecipient) Wrap(fileKey []byte) ([]*age.Stanza, error) {
+	return w.rec.Wrap(fileKey)
+}
 
 // Identities returns all identities, used for decryption.
 func (a *Age) Identities(ctx context.Context) ([]age.Identity, error) {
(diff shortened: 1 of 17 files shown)


internal/backend/crypto/age/identities.go
```diff
@@ -1,24 +1,54 @@
...
+	"filippo.io/age/agessh"
+	"filippo.io/age/plugin"
 	"github.com/gopasspw/gopass/pkg/appdir"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
 	"github.com/gopasspw/gopass/pkg/debug"
 )
 
 var idRecpCacheKey = "identity"
 
+type wrappedIdentity struct {
```
dominikschulz, line 27: I think a wrappedIdentity would deserved a (short) comment.

internal/backend/crypto/age/identities.go
```diff
@@ -52,11 +82,71 @@ func (a *Age) Identities(ctx context.Context) ([]age.Identity, error) {
 	return ids, nil
 }
 
-// IdentityRecipients returns a slice of recipients dervied from our identities.
+func parseIdentity(s string) (age.Identity, error) {
+	switch {
+	case strings.HasPrefix(s, "AGE-PLUGIN-"):
+		sp := strings.Split(s, "|")
```
dominikschulz, line 88: What is `sp[0]` and what is `sp[1]`?

dominikschulz (changes requested):
Nice, thanks a lot.

I think my only ask is to add some of the details from the PR message to the PR itself (e.g. wrappedIdentity and some notes about the recipient format).

dominikschulz (approved):
Very nice!

PR #2961, opened 2024-10-02
[bugfix] Replace ~ with user homedir if $GOPASS_HOMEDIR is not set

### LINKED ISSUE #2916: config: git operations do not work using `~` in `mounts.path`
### Summary
All git operations (`gopass git ...`) give this error: `chdir ~/.password-store: no such file or directory`.
All other commands work.

### Steps To Reproduce
The path in config file uses `~` in place of my home folder:
```dosini
[mounts]
	path = ~/.password-store
```
Git operations work if I use the absolute location of the root store (`/home/user/.password-store`) for the path

### Environment
- OS: Ubuntu 24.04
- gopass Version: 1.15.13
- Installation method: apt repository

cal-andrew:
Fixes #2916

### DIFF
--- a/pkg/fsutil/fsutil.go
+++ b/pkg/fsutil/fsutil.go
@@ -39,12 +39,16 @@ func ExpandHomedir(path string) string {
 
 // CleanPath resolves common aliases in a path and cleans it as much as possible.
 func CleanPath(path string) string {
-	// Only replace ~ if GOPASS_HOMEDIR is set. In that case we do expect any reference
-	// to the users homedir to be replaced by the value of GOPASS_HOMEDIR. This is mainly
-	// for testing and experiments. In all other cases we do want to leave ~ as-is.
+	// Replace ~ with GOPASS_HOMEDIR if set (mainly for testing and experiments),
+	// otherwise replace ~ with user's homedir if set. We expect any reference
+	// to the user's homedir to be replaced with one of these two values.
 	if len(path) > 1 && path[:2] == "~/" {
 		if hd := os.Getenv("GOPASS_HOMEDIR"); hd != "" {
 			return filepath.Clean(hd + path[2:])
+		}
+
+		if home, err := os.UserHomeDir(); err == nil {
+			return filepath.Clean(home + path[1:])
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


AnomalRoil:
Mhhh, I thins this conflicts with the idea of
https://github.com/gopasspw/gopass/blob/486c1654383160cf23f4674a4143187078f9bc59/pkg/fsutil/fsutil.go#L40-L56

The idea being that "~" is replaced with `$GOPASS_HOMEDIR` when it's set instead of `$HOME`, for testing and dev purposes.
Your change feels like it wouldn't do that for git operations anymore (and I'm surprised our tests are passing since they do use `GOPASS_HOMEDIR` IIRC).

Furthermore, the fact that in #2916 it's working for the root store but not substores seems to hint at a deeper problem in how we handle substore vs root store paths.

AnomalRoil:
I think the deeper problem here is that `~` is a shell meta-character, and so paths starting with `~` have been set by incorrectly quoting `"~"` while setting the path, thus preventing it from being expanded  to `$HOME` by the shell itself, demo:
```
$ gopass config mounts.path
/home/anomalroil/.local/share/gopass/stores/root
$ gopass config mounts.path "~/.local/share/gopass/stores/root"
~/.local/share/gopass/stores/root
$ gopass config mounts.path
~/.local/share/gopass/stores/root
$ gopass config mounts.path ~/.local/share/gopass/stores/root
/home/anomalroil/.local/share/gopass/stores/root
$ gopass config mounts.path
/home/anomalroil/.local/share/gopass/stores/root
```
Note that bash does allow to have quoted `~` in `$PATH`, but most shells do not.

But I think the initial idea was to be able to share a `gopass config` across a team, and so supporting `~` makes sense I think.
Just maybe not at the `git` level, but at a upper level?

dominikschulz:
@cal-andrew Not sure if `CleanPath` covers all instances, but I think it would be a good start. Changing this on the `gitfs` doesn't make sense as @AnomalRoil pointed out.

cal-andrew:
The PR now tries to replace ~ in `CleanPath`, and enables testing this in Windows.

dominikschulz (approved):

AnomalRoil (approved):

cal-andrew:
Thank you!

PR #2966, opened 2024-10-09
[bugfix] Copy with trailing slash at destination.

### LINKED ISSUE #2912: `gopass cp` has a different behaviour depending on whether the recursive directory already exist
### Summary
The copy command has a different behaviour when recursively copying secret directories from one store to another.
Even if the `--force` flag is given, it will recursively copy the directory structure even tho a trailing slash is given.

`gopass cp --help` states, that the contents will be flattened, if a trailing slash is found.
However, the contents are not flattened, if the directory already exists in the target store.

I've also tried every combination of trailing slashes, all with the same outcome:
- `gopass cp --force full-store/some/secret/directory empty-store/some/secret/directory`
- `gopass cp --force full-store/some/secret/directory/ empty-store/some/secret/directory`
- `gopass cp --force full-store/some/secret/directory empty-store/some/secret/directory/`
- `gopass cp --force full-store/some/secret/directory/ empty-store/some/secret/directory/`

I could not spot any difference in behavior with/without the `--force` flag at all.

### Steps To Reproduce
1. Two stores; one contains a directory of secrets, one is empty
2. Copy the directory to the empty store `gopass cp --force full-store/some/secret/directory/ empty-store/some/secret/directory/`
3. Verify the entire directory was copied to `empty-store/some/secret/directory/`, e.g. with `gopass ls empty-store`
4. Now run `gopass cp --force full-store/some/secret/directory/ empty-store/some/secret/directory/` again
5. The secrets were now copied to `empty-store/some/secret/directory/full-store/some/secret/directory`

### Expected behavior
The second `gopass cp` command should overwrite the secrets in the existing directory and not copy the directory structure as well.

### Environment

- OS: macOS Sonoma 14.5
- OS version: `Darwin FE-C-014F3 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64`
- gopass Version: `gopass 1.15.13 go1.22.2 darwin arm64`
- Installation method: `brew`

vijay-kesanakurthi:
FIxes #2912

### DIFF
--- a/internal/action/copy.go
+++ b/internal/action/copy.go
@@ -3,6 +3,8 @@ package action
 import (
 	"context"
 	"fmt"
+	"path/filepath"
+	"strings"
 
 	"github.com/gopasspw/gopass/internal/action/exit"
 	"github.com/gopasspw/gopass/pkg/ctxutil"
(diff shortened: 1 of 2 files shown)


dominikschulz (commented):
The implementation looks great. Clean and descriptive. But I was wondering if you could at least add one test case for the new / changed behaviour?

vijay-kesanakurthi:
I have added a test , could you check and give feedback if any changes needed?

dominikschulz (approved):
LGTM. Thanks!

Issue #2949, opened 2024-09-26
gopass setup should validate empty email


jcrben:
### Summary
At least if you're using the gpg backend, it needs an email otherwise it spits out a message like this:

```
2024/09/25 18:53:46 failed to check private keys: failed to create new private key: 
failed to create new private key: 
failed to run command: '/usr/bin/gpg [/usr/bin/gpg --batch --gen-key]': 
"gpg: Generating a RSA/RSA key pair\ngpg: -:8: missing argument\n" - exit status 2
```

Probably obvious to most, but still better to give a friendly validation error.

### Steps To Reproduce
See above

### Expected behavior
Refuse to proceed with empty email and gpg backend?

### Environment

- OS: Ubuntu 24.04
- OS version: Linux lenovo-legion 5.15.153.1-microsoft-standard-WSL2 #1 SMP Fri Mar 29 23:14:13 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: gopass 1.15.14 go1.22.5 linux amd64
- Installation method: manual .deb file

zenc0derr:
Hi @jcrben , Are u working on this issue?

jcrben:
no

gagandeepp:
@jcrben  interested , please assign

jcrben:
@gagandeepp I can't assign, maybe @AnomalRoil or @dominikschulz can assign

gagandeepp:
@jcrben 
![Screenshot 2024-10-07 at 6 14 55 PM](https://github.com/user-attachments/assets/6de77b38-566a-421b-b529-4ca40acdd8dd)

I am able to setup without email as mentioned in the screenshot

gagandeepp:
@dominikschulz can you please see my comments above?

jcrben:
You selected 2, try just hitting enter to go with the default (0).

I just replicated it on a fairly clean Ubuntu 24.04 machine (Framework laptop). 

Steps:
*  brew.sh install per instructions
* brew install --build-from-source gopass
* gopass setup
* hit enter a few times

![image](https://github.com/user-attachments/assets/a4310ab4-182c-4f6e-b5cd-c03fc2872526)

gagandeepp:
@dominikschulz  Please review

PR #2969, opened 2024-10-10
Empty Email check implemented

gagandeepp:
Fixes: https://github.com/gopasspw/gopass/issues/2949

### DIFF
--- a/internal/action/setup.go
+++ b/internal/action/setup.go
@@ -138,8 +138,6 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
 		}
 
 		email, err = termio.AskForString(ctx, "📧 What is your email?", email)
-
-		
 		if err != nil {
 			return err
 		}
--- a/internal/action/setup.go
+++ b/internal/action/setup.go
@@ -140,7 +140,7 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
 		email, err = termio.AskForString(ctx, "📧 What is your email?", email)
 
 		if strings.TrimSpace(email) == "" {
-			return fmt.Errorf("Please enter a valid email address to proceed")
+			return fmt.Errorf("⛔️ Please enter a valid email address to proceed")
 		}
 		if err != nil {
 			return err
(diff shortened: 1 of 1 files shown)


internal/action/setup.go
```diff
@@ -137,6 +138,10 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
 		}
 
 		email, err = termio.AskForString(ctx, "📧 What is your email?", email)
+
+		if strings.TrimSpace(email) == "" {
+			return fmt.Errorf("failed to create a usable key pair")
```
jcrben, line 143: Disclaimer: I'm not a maintainer. But I think this message could be more user-friendly. Something like: "You did not provide an email. An email is required to generate a gpg key pair, so please provide one."

gagandeepp:
@dominikschulz  Please review this

internal/action/setup.go
gagandeepp: Done

internal/action/setup.go
```diff
@@ -137,6 +138,10 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
 		}
 
 		email, err = termio.AskForString(ctx, "📧 What is your email?", email)
+
+		if strings.TrimSpace(email) == "" {
```
dominikschulz, line 142: nit: I think checking for the email string should happen after error handling in the next block, i.e. please move these lines below line 147.

gagandeepp: Done @dominikschulz

jcrben:
What's the story on the age backend? Does the bug also happen there? I don't really know much about age, not sure if it requires an email.

gagandeepp:
> What's the story on the age backend? Does the bug also happen there? I don't really know much about age, not sure if it requires an email.

@jcrben honestly no idea about it

gagandeepp:
@dominikschulz please review

internal/action/setup.go
gagandeepp: @jcrben  Can you please resolve this conversation?

internal/action/setup.go
gagandeepp: @dominikschulz  Can you please resolve this conversation?

dominikschulz (approved):
LGTM

AnomalRoil:
No need for emails with the `age` crypto backend, unless one is using the `gitfs` storage backend with it.

dominikschulz:
@AnomalRoil I don't think we have the infrastructure to plumb through these different backend requirements, yet.

PR #2972, opened 2024-10-15
[n/a] Correcting mistake in search+replace

AnomalRoil:
Seems I was a bit hasty when migrating to Go 1.23 and renaming the function "clear" which is now a builtin into "clearClip".


dominikschulz (approved):
Thanks. I did notice the odd change to some of the `debug.Log` statements but was assuming those were intentional. The others I did actually miss :/

Issue #2963, opened 2024-10-06
Gopass command before setting up gopass


sarojregmi200:
### Summary
It is not a bug, but thought it might help make this project a tiny bit better. I tried setting up the gopass today and installed it successfully and tried to run gopass by running `gopass` binary. And it says it is not yet setup to setup type `gopass init` and then exits the process. 
I think it can be improved. 
![Screenshot 2024-10-06 at 17 41 35](https://github.com/user-attachments/assets/23ef9153-9cf5-417c-8ff4-12faa5c7d022)

### Steps To Reproduce
- install gopass.
- Type gopass in your terminal and see (you should not have it already configured).

### Better behavior
Gopass is not yet setup in your system do you want to set it up? A small message like this will be a smoother onboarding experience for those who are setting it up for the first time.

### Environment
Local, personal pc.

- OS: [Arch linux]
- gopass Version: [1.15.14]
- Installation method: [pacman]

 
>I have just started using the project and have searched for similar issue but have not found one it is what I think is a goodway and what I felt while getting the error message.

dominikschulz:
Thanks for the report. But I'm afraid I don't fully understand how you'd like it to change.

We already provide a hint on what to do (`gopass setup`).
Could you maybe provide an example of how a more welcoming error message could look like?

sarojregmi200:
Sure.

It would be great if it just continued, rather than exiting the program.
Saying something like.
**"Do you want to continue to setup? yes(y) no(n)".**

Something like this will better improve the flow of the program.

PR #2975, opened 2024-10-20
[FEATURE] Ask for setup if not initialized

Juneezee:
Fixes https://github.com/gopasspw/gopass/issues/2963.

### Before

https://github.com/user-attachments/assets/22f764d7-bc0c-4add-9bae-e7d9d2381cdd

### After

https://github.com/user-attachments/assets/eeb03aed-f835-4e19-b93e-7d6ffe827d70

### DIFF
--- a/internal/action/init.go
+++ b/internal/action/init.go
@@ -55,6 +55,15 @@ func (s *Action) IsInitialized(c *cli.Context) error {
 	out.Printf(ctx, logo)
 	out.Printf(ctx, "🌟 Welcome to gopass!")
 	out.Noticef(ctx, "No existing configuration found.")
+
+	contSetup, err := termio.AskForBool(ctx, "❓ Do you want to continue to setup?", true)
+	if err != nil {
+		return err
+	}
+	if contSetup {
+		return s.Setup(c)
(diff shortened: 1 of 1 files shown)


AnomalRoil (approved):
Looks like a better UX, lgtm

internal/action/init.go
```diff
@@ -55,6 +55,15 @@ func (s *Action) IsInitialized(c *cli.Context) error {
 	out.Printf(ctx, logo)
 	out.Printf(ctx, "🌟 Welcome to gopass!")
 	out.Noticef(ctx, "No existing configuration found.")
+
+	contSetup, err := termio.AskForBool(ctx, "❓ Do you want to continue to setup?", true)
```
dominikschulz, line 59: The third parameter to AskForBool is the default. I think that should be false in this case.

If gopass is run non-interactively it will always use the default value and we only want to start the setup flow if the user explicitly asks for it. So please change `true` to `false` and then we should be good.

dominikschulz (changes requested):

dominikschulz (approved):

PR #2980, opened 2024-10-27
Steam TOTP support

nikicat:
This PR adds Steam TOTP encoding support. totp:// url should contain `encoding=steam` query parameter.


nikicat:
Related to #2566

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/gopasspw/gopass
 
 go 1.23.2
 
+replace github.com/pquerna/otp => github.com/nikicat/otp v1.4.0-ojo
```
AnomalRoil, line 5: This would be too much of a supply chain attack risk, and also replace statement in a go.mod aren't meant to be shipped in a library, and some bits of gopass are acting as a library, so mot great.

Any chance you could upstream that to pquerna otp code or provide the derivation in our code directly?

We're trying to generally reduce the number of dependencies, not add new ones.

AnomalRoil (changes requested):

go.mod
nikicat: I'll try to merge this change upstream, that's why I've created a draft PR. I think conversion from a standard digital code format is possible, but is a bit hacky, as it needs to requires at least 8 digits code to be requested and then reencoded to Steam's base26 format.

AnomalRoil: Yeah, sorry I fat fingered the undraft button when trying to authorize the Github Action

Issue #2933, opened 2024-08-31
[age] allow custom paths to ssh files 


fionera:
### Summary
Currently it is not possible to use the age backend and point gopass to a specific key to use, e.g. when having a subdir in .ssh.

### Steps To Reproduce
1. Have a key somewhere not in ~/.ssh
2. Setup gopass with age

### Expected behavior
A flag to specify additional folders/directly select a key

### Environment
- OS: Mac OS X Sequoia
- OS version: Darwin 24.1.0 Darwin Kernel Version 24.1.0: Fri Aug  2 03:23:05 PDT 2024; root:xnu-11215.40.24.0.1~10/RELEASE_ARM64_T6000 arm64 Darwin
- gopass Version: gopass 1.15.13 go1.22.6 darwin arm64
- Installation method: Nix

dominikschulz:
That sounds like something we should support. Good point.

Dxuian:
is this still open?

dominikschulz:
@Dxuian Yes, still open. Although I haven't thought through the implication of this. Feel free to give it a try, but if it unnecessarily complicates the implementation we might need to reject the request. Depends on how the implementation will end up.

PR #2981, opened 2024-10-28
specify ssh dir

JoelLau:
override default ~/.ssh/ directory for ssh paths

- contributes to #2933

### DIFF
--- a/internal/backend/crypto/age/ssh.go
+++ b/internal/backend/crypto/age/ssh.go
@@ -32,14 +32,11 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 		return sshCache, nil
 	}
 
-	// notice that this respects the GOPASS_HOMEDIR env variable, and won't
-	// find a .ssh folder in your home directory if you set GOPASS_HOMEDIR
-	uhd := appdir.UserHome()
-	sshDir := filepath.Join(uhd, ".ssh")
-	if !fsutil.IsDir(sshDir) {
-		debug.Log("no .ssh directory found at %s. Ignoring SSH identities", sshDir)
+	sshDir, err := getSSHDir()
+	if err != nil {
+		debug.Log("asdf: %s", err)
 
-		return nil, fmt.Errorf("no identities found: %w", ErrNoSSHDir)
+		return nil, fmt.Errorf("asdf: %w", err)
 	}
 
 	files, err := os.ReadDir(sshDir)
@@ -67,6 +64,31 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 	debug.Log("returned %d SSH Identities", len(ids))
 
 	return ids, nil
+}
+
+func getSSHDir() (string, error) {
+	preferredPath := os.Getenv("GOPASS_SSHDIR")
+	if preferredPath != "" {
+		return preferredPath, nil
+	}
+	sshDir := filepath.Join(preferredPath, ".ssh")
+	if !fsutil.IsDir(sshDir) {
+		debug.Log("no .ssh directory found at %s. Ignoring SSH identities", sshDir)
+
+		return "", fmt.Errorf("no identities found: %w", ErrNoSSHDir)
+	}
+
+	// notice that this respects the GOPASS_HOMEDIR env variable, and won't
+	// find a .ssh folder in your home directory if you set GOPASS_HOMEDIR
+	uhd := appdir.UserHome()
+	sshDir = filepath.Join(uhd, ".ssh")
+	if !fsutil.IsDir(sshDir) {
+		debug.Log("no .ssh directory found at %s. Ignoring SSH identities", sshDir)
+
+		return "", fmt.Errorf("no identities found: %w", ErrNoSSHDir)
+	}
+
+	return "", nil
 }
 
 // parseSSHIdentity parses a SSH public key file and returns the recipient and the identity.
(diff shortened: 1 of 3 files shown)


JoelLau:
@dominikschulz i realise this doesn't exactly implement the features described in #2933, but would this be a step in the right direction?

internal/backend/crypto/age/ssh.go
```diff
@@ -32,14 +32,11 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
...
-	uhd := appdir.UserHome()
-	sshDir := filepath.Join(uhd, ".ssh")
-	if !fsutil.IsDir(sshDir) {
-		debug.Log("no .ssh directory found at %s. Ignoring SSH identities", sshDir)
+	sshDir, err := getSSHDir()
+	if err != nil {
+		debug.Log("asdf: %s", err)
 
-		return nil, fmt.Errorf("no identities found: %w", ErrNoSSHDir)
+		return nil, fmt.Errorf("asdf: %w", err)
```
AnomalRoil, line 39: That's a weird error?

internal/backend/crypto/age/ssh.go
```diff
@@ -32,14 +32,11 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
...
 
-	// notice that this respects the GOPASS_HOMEDIR env variable, and won't
-	// find a .ssh folder in your home directory if you set GOPASS_HOMEDIR
-	uhd := appdir.UserHome()
-	sshDir := filepath.Join(uhd, ".ssh")
-	if !fsutil.IsDir(sshDir) {
-		debug.Log("no .ssh directory found at %s. Ignoring SSH identities", sshDir)
+	sshDir, err := getSSHDir()
+	if err != nil {
+		debug.Log("asdf: %s", err)
```
AnomalRoil, line 37: Why `asdf`?

internal/backend/crypto/age/ssh.go
```diff
@@ -69,6 +66,31 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 	return ids, nil
 }
 
+func getSSHDir() (string, error) {
+	preferredPath := os.Getenv("GOPASS_SSHDIR")
```
AnomalRoil, line 70: Please add details to the config.md and age.md documentation about how this variable works.

AnomalRoil (changes requested):
While it's not a flag, I think a env variable for that is actually more sensible.

A test would be nice too.

internal/backend/crypto/age/ssh.go
JoelLau: sorry it was just a placeholder. wanted to get your early opinion

dominikschulz (approved):

JoelLau:
@AnomalRoil are there any other changes required for this PR?

JoelLau:
if its not too much trouble, can i get this PR accepted for hacktoberfest by merging this pr or marking it with the `hacktoberfest-accepted` label?

internal/backend/crypto/age/ssh.go
```diff
@@ -69,6 +66,24 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 	return ids, nil
 }
 
+func getSSHDir() (string, error) {
+	preferredPath := os.Getenv("GOPASS_SSH_DIR")
+	sshDir := filepath.Join(preferredPath, ".ssh")
+	if fsutil.IsDir(sshDir) {
```
dominikschulz, line 72: I think we should also check if `preferredPath` is not empty here. Otherwise we might end up with `.ssh` (i.e. .ssh in the current pwd) w/o users asking for it.

dominikschulz (changes requested):

internal/backend/crypto/age/ssh.go
JoelLau: thanks @dominikschulz , i've added the check

internal/backend/crypto/age/ssh.go
```diff
@@ -69,6 +66,24 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 	return ids, nil
 }
 
+func getSSHDir() (string, error) {
+	preferredPath := os.Getenv("GOPASS_SSH_DIR")
+	sshDir := filepath.Join(preferredPath, ".ssh")
+	if fsutil.IsDir(sshDir) && sshDir != "" {
```
dominikschulz, line 72: This should actually check if prefrerredPath is empty. Otherwise we might still pick up a .ssh dir does exist in the current pwd.

dominikschulz (changes requested):

internal/backend/crypto/age/ssh.go
```diff
@@ -69,6 +66,24 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 	return ids, nil
 }
 
+func getSSHDir() (string, error) {
+	preferredPath := os.Getenv("GOPASS_SSH_DIR")
+	sshDir := filepath.Join(preferredPath, ".ssh")
+	if fsutil.IsDir(sshDir) && preferredPath != "" {
```
AnomalRoil, line 72: Could you swap both conditions?
`os.Getenv` is most likely to return an empty string for most users in most of the case, so let's not run `IsDir` if we don't need to.

AnomalRoil (changes requested):

dominikschulz (approved):

PR #2993, opened 2024-11-02
[CLEANUP] Replace experimental `maps` and `slices` with stdlib

Juneezee:
The experimental functions are now available in the standard library in Go 1.23 [^1].

The only difference is that `maps.Keys` and `maps.Values` in the standard library return an iterator, while `maps.Keys` and `maps.Values` in the `golang.org/x/exp` package return a slice.

[^1]: https://go.dev/doc/go1.23#new-unique-package

### DIFF
--- a/internal/backend/registry.go
+++ b/internal/backend/registry.go
@@ -1,12 +1,14 @@
 package backend
 
 import (
+	"cmp"
 	"context"
 	"fmt"
-	"sort"
+	"maps"
+	"slices"
 	"sync"
 
-	"golang.org/x/exp/maps"
+	"github.com/gopasspw/gopass/internal/set"
 )
 
 var (
@@ -73,10 +75,7 @@ func (r *Registry[K, V]) BackendNames() []string {
 	r.RLock()
 	defer r.RUnlock()
 
-	names := maps.Keys(r.nameToBackend)
-	sort.Strings(names)
-
-	return names
+	return set.SortedKeys(r.nameToBackend)
 }
 
 func (r *Registry[K, V]) Backends() []V {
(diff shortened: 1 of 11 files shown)


pkg/gitconfig/gitconfig_test.go
```diff
@@ -343,7 +342,8 @@ func TestParseComplex(t *testing.T) {
 
 	c := ParseConfig(strings.NewReader(configSampleComplex))
 
-	assert.Contains(t, maps.Keys(c.vars), "core.sshCommand")
+	_, ok := c.vars["core.sshCommand"]
+	assert.True(t, ok)
```
AnomalRoil, line 346: why not use `SortedKeys` here?

AnomalRoil (commented):
A few questions I guess

pkg/gitconfig/gitconfig_test.go
Juneezee: The old code

```go
assert.Contains(t, maps.Keys(c.vars), "core.sshCommand")
```

was just testing whether the `core.sshCommand` key exists in the `c.vars` map or not.

Obviously we can do `assert.Contains(t, set.SortedKeys(c.vars), "core.sshCommand")` too to test whether `core.sshCommand` exists in `c.vars` or not. But `_, ok := c.vars["core.sshCommand"]` does the job in a much simplier way

internal/backend/registry.go
```diff
@@ -96,11 +95,10 @@ func (r *Registry[K, V]) Prioritized() []V {
...
 
 	bes := maps.Values(r.backends)
-	sort.Slice(bes, func(i, j int) bool {
-		return bes[i].Priority() < bes[j].Priority()
-	})
 
-	return bes
+	return slices.SortedFunc(bes, func(a, b V) int {
+		return cmp.Compare(a.Priority(), b.Priority())
+	})
```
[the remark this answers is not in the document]
Juneezee: We can't use `slices.Sorted`.

image

You can see the types permitted by `cmp.Ordered` here: https://pkg.go.dev/cmp#Ordered. We are using generics so it is not possible to use `slices.Sorted` here.

dominikschulz (approved):
Thanks!

Issue #3012, opened 2024-12-15
Bug Report: code check `call.Err !=nil` but return a nil value error `err`


alingse:
### Summary

do you mean `return call.Err` or just `return nil` to ignore the call.Err ?

### Environment

- OS: [e.g. Mac OS X High Sierra, Ubuntu 18.04, Windows 10, ...]
- OS version: [uname -a]
- gopass Version: [gopass version]
- Installation method: [e.g. from source, brew, gopass repo]

### Additional context

I create a linter to detect code that returns a non-relevant nilness error bug. I checked the top 1000 GitHub Go repositories and found this, all result listed in https://github.com/alingse/sundrylint/issues/4

dominikschulz:
Please provide more details.

alingse:
@dominikschulz 

https://github.com/gopasspw/gopass/blob/bd408d6cd47aaccdcc56fed606b21581d7ef63e4/internal/notify/notify_dbus.go#L31-L34
```go
	if err != nil {
		debug.Log("DBus failure: %s", err)

		return err
	}

	obj := conn.Object("org.freedesktop.Notifications", "/org/freedesktop/Notifications")
	call := obj.Call("org.freedesktop.Notifications.Notify", 0, "gopass", uint32(0), iconURI(ctx), subj, msg, []string{}, map[string]dbus.Variant{"transient": dbus.MakeVariant(true)}, int32(3000))
	if call.Err != nil {
		debug.Log("DBus notification failure: %s", call.Err)

		return err
	}
```

check `if call.Err != nil` but just return a already nil value `err`

dominikschulz:
Ha, looks legit, indeed. Thanks.

Denis-Kuso:
I’d like to work on fixing this. Seems like just a small correction in the variable return. Will submit a PR soon.

PR #3015, opened 2024-12-19
[BUGFIX] replace return of wrong error variable

Denis-Kuso:
This fixes the issue #3012 where the error returned by the `Notify()` call would incorrectly return a nil value (from `dbus.SessionBus()`), instead of the correct non-nil value (from `call.Err`).

### DIFF
--- a/internal/notify/notify_dbus.go
+++ b/internal/notify/notify_dbus.go
@@ -31,7 +31,7 @@ func Notify(ctx context.Context, subj, msg string) error {
 	if call.Err != nil {
 		debug.Log("DBus notification failure: %s", call.Err)
 
-		return err
+		return call.Err
 	}
 
 	return nil


Denis-Kuso:
I noticed that errors returned by the `Notify` function are ignored throughout
the codebase, perhaps indicating that notification failures aren’t considered
critical. Given this, I didn't add tests for the obj.Call error path,
as it doesn't seem necessary (nor is it currently tested) to propagate the
error or affect the application's flow. Let me know if you'd prefer otherwise.

dominikschulz (approved):
Thank you.

dominikschulz:
> I noticed that errors returned by the `Notify` function are ignored throughout the codebase, perhaps indicating that notification failures aren’t considered critical. Given this, I didn't add tests for the obj.Call error path, as it doesn't seem necessary (nor is it currently tested) to propagate the error or affect the application's flow. Let me know if you'd prefer otherwise.

Correct choice. For most instances I don't consider the returned errors from this method important since notifications and mostly best effort anyway. But I still think it's important that the method correctly returns errors - if any - so the callers can decide if and how to handle them.

PR #3027, opened 2025-01-03
[BUGFIX] Check if any usable key matches on clone

14mRh4X0r:
This commit changes the check whether we have access to the cloned repository from checking whether *all* of our usable keys are included to whether *any* of our usable keys are included.


dominikschulz (changes requested):
Thank you very much. Just add a short comment what recpSet.Choose(idSet.Contains) does. I had to go back and forth a few times.

dominikschulz (approved):

AnomalRoil (approved):

PR #3029, opened 2025-01-06
[bugfix] Don't check for autosync on manual triggered sync (#3026)

### LINKED ISSUE #3026: autosync = false breaks manual sync
Setting `core.autosync = false` breaks manual syncing with `gopass sync` command. It then just does nothing (it seems):

    19:44:07 mcn@x2b ~ grep sync .config/gopass/config 
    autosync=false
    
    19:44:54 mcn@x2b ~ time gopass sync
    🚥 Syncing with all remotes ...
    ✅ All done
    
    real	0m0,437s
    user	0m0,270s
    sys	0m0,179s
    
In pretty much no time it "syncs" and states all being done, while actually not at all syncing. When setting `core.autosync = true` it then does it also manually:

    19:44:57 mcn@x2b ~ grep sync .config/gopass/config 
    autosync=true
    
    19:45:39 mcn@x2b ~ time gopass sync
    🚥 Syncing with all remotes ...
    [<root>] 
       gitfs pull and push ... OK (no changes)
       done
       …
    
I prefer it not autosyncing, because I find it annoying to wait for the syncing of >20 stores when I only want to use the command line tool. So I have set `autosync.interval = 9999` to work around this issue. Is there another way to do that?

gopass 1.15.15 go1.23.4 linux amd64 on NixOS unstable with `home.packages = with pkgs; [ gopass gopass-jsonapi ];`

TM2500:
* [bugfix] Don't check for autosync on manual triggered sync

Fixes #3026

### DIFF
--- a/internal/action/sync.go
+++ b/internal/action/sync.go
@@ -85,6 +85,7 @@ func (s *Action) autoSync(ctx context.Context) error {
 	debug.Log("autosync - interval: %s", syncInterval)
 
 	if time.Since(ls) > syncInterval {
+		ctx = ctxutil.WithInteractive(ctx, true)
 		err := s.sync(ctx, "")
 		if err != nil {
 			autosyncLastRun = time.Now()
@@ -161,10 +162,12 @@ func (s *Action) sync(ctx context.Context, store string) error {
 func (s *Action) syncMount(ctx context.Context, mp string) error {
 	// using GetM here to get the value for this mount, it might be different
 	// than the global value
-	if as := s.cfg.GetM(mp, "core.autosync"); as == "false" {
-		debug.Log("not syncing %s, autosync is disabled for this mount", mp)
+	if ctxutil.HasAutoSync(ctx) {
+		if as := s.cfg.GetM(mp, "core.autosync"); as == "false" {
+			debug.Log("not syncing %s, autosync is disabled for this mount", mp)
 
-		return nil
+			return nil
+		}
 	}
 
 	ctxno := out.WithNewline(ctx, false)
--- a/internal/action/sync.go
+++ b/internal/action/sync.go
@@ -46,7 +46,7 @@ func init() {
 
 // Sync all stores with their remotes.
 func (s *Action) Sync(c *cli.Context) error {
-	return s.sync(ctxutil.WithGlobalFlags(c), c.String("store"))
+	return s.sync(ctxutil.WithGlobalFlags(c), c.String("store"), false)
 }
 
 func (s *Action) autoSync(ctx context.Context) error {
@@ -85,8 +85,7 @@ func (s *Action) autoSync(ctx context.Context) error {
 	debug.Log("autosync - interval: %s", syncInterval)
 
 	if time.Since(ls) > syncInterval {
-		ctx = ctxutil.WithAutoSync(ctx, true)
-		err := s.sync(ctx, "")
+		err := s.sync(ctx, "", true)
 		if err != nil {
 			autosyncLastRun = time.Now()
 		}
@@ -97,7 +96,7 @@ func (s *Action) autoSync(ctx context.Context) error {
 	return nil
 }
 
-func (s *Action) sync(ctx context.Context, store string) error {
+func (s *Action) sync(ctx context.Context, store string, isAutosync bool) error {
 	// we just did a full sync, no need to run it again
 	if time.Since(autosyncLastRun) < 10*time.Second {
 		debug.Log("skipping sync. last sync %ds ago", time.Since(autosyncLastRun))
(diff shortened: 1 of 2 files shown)


TM2500:
Additional Tests are needed to verify this bugfix.
As soon as they are created, the draft state is removed from the PR.

internal/action/sync.go
```diff
@@ -85,6 +85,7 @@ func (s *Action) autoSync(ctx context.Context) error {
 	debug.Log("autosync - interval: %s", syncInterval)
 
 	if time.Since(ls) > syncInterval {
+		ctx = ctxutil.WithInteractive(ctx, true)
```
AnomalRoil, line 88: It looks like you're relying on the interactive context key in the end, instead of the newly added autosync one, no?

pkg/ctxutil/ctxutil.go
```diff
@@ -489,3 +490,18 @@ func IsHidden(ctx context.Context) bool {
...
+
+// HasAutoSync returns true if syncs are triggered by autosync.
+func HasAutoSync(ctx context.Context) bool {
+	bv, ok := ctx.Value(ctxKeyAutoSync).(bool)
+	if !ok {
+		return false
+	}
+
+	return bv
+}
```
AnomalRoil, line 507: Do you really need these?
It looks like you're relying on the interactive context key above, instead of that newly added autosync one, no?

internal/action/sync.go
```diff
@@ -161,10 +162,12 @@ func (s *Action) sync(ctx context.Context, store string) error {
 func (s *Action) syncMount(ctx context.Context, mp string) error {
 	// using GetM here to get the value for this mount, it might be different
 	// than the global value
-	if as := s.cfg.GetM(mp, "core.autosync"); as == "false" {
-		debug.Log("not syncing %s, autosync is disabled for this mount", mp)
+	if ctxutil.HasAutoSync(ctx) {
```
AnomalRoil, line 165: Wait, you're setting the interactive key above, not the autosync one, no?

internal/action/sync.go
AnomalRoil: You might mean to use 
```suggestion
		ctx = ctxutil.WithAutoSync(ctx, true)
```
there, no?

internal/action/sync.go
TM2500: I propably mixed something up. I need to take a look at it again.
I hope to include meaningful tests as well, as there are none for sync.

internal/action/sync.go
TM2500: Yes, I intended to use the new context.

TM2500:
For tests I would have to write new tests utils, that support fake-remote git repos as stores. This may take me a while to figure out how to include it in the available tooling.
The bug is fixed with the last two commits.

dominikschulz:
@TM2500 Thanks for the PR. I won't press for tests if they don't provide more value than maintanence overhead. Feel free to mark this ready for review as is or whenever you think it's ready. Then I'll give it a good look and let you know if I think anything is amiss.

TM2500:
> @TM2500 Thanks for the PR. I won't press for tests if they don't provide more value than maintanence overhead. Feel free to mark this ready for review as is or whenever you think it's ready. Then I'll give it a good look and let you know if I think anything is amiss.

Yes, I think tests are an order of magnitude more work, as we would have to find a way to generate dummy-remote stores in our test suite.

I will mark this for review and the PR transitions from draft to a regular PR.

Thank you.

internal/action/sync.go
```diff
@@ -161,10 +162,12 @@ func (s *Action) sync(ctx context.Context, store string) error {
 func (s *Action) syncMount(ctx context.Context, mp string) error {
 	// using GetM here to get the value for this mount, it might be different
 	// than the global value
```
AnomalRoil, line 164: you could have moved the comment inside of the first `if` I guess

AnomalRoil (dismissed):
LGTM

dominikschulz (commented):
You are adding a bunch of code just to pass a bool in to the sync method.
Wouldn't it easier to just pass an additional bool to s.sync to signal that it's being invoked from autosync? Or is there something I'm missing?

TM2500:
> You are adding a bunch of code just to pass a bool in to the sync method. Wouldn't it easier to just pass an additional bool to s.sync to signal that it's being invoked from autosync? Or is there something I'm missing?

You are right, as long as we only depend on the autosync-context only within the sync package, we could just use a function-parameter.
Since there is currently no need for it in other packages, I will rewrite the Patch.

TM2500:
I refactored to function arguments.
This is ready to merge.

dominikschulz (approved):
Much simpler. Thanks a lot!

PR #3048, opened 2025-02-08
Roll over self-updater key for 2025

### LINKED ISSUE #3047: Update release signing key
The signing key expired and we did notice this too late (again).

I'm working on a key rollover and adding some tests and instructions that should help with this the next time.

dominikschulz:
We missed to supply a new key before the old key expired again and one of the tests started to fail once the old key was expired.

This is a more comprehensive approach to address this going forward. Not perfect, yet. But should put us (especially myself) in a better position next time this comes up.

Changes in this PR:

* The public part of the new key that I have generated. The private part will be injected into GHA by the time this is being merged.
* A sloppy write-up of the steps I did this time so I have less trial and error the next time we need to roll over keys.
* A copy of the verification test that mocks the verification time 6 Months into the future. That way we should notice before the key is actually expired. This isn't perfect, because we will then have unrelated test failures 6 Months before the key expires, but at least this gives me ~1.5y to figure out a better way of giving an advance notice.

Fixes #3047


internal/updater/README.md
```diff
@@ -0,0 +1,60 @@
...
+* As usual send a PR and get this merged. Consider kicking off a new release, if that makes sense.
+* STOP HERE. For a seamless key rollover we need to wait until most users had a chance to update to a version that has both the old and the new keys. So if possible wait a few months at least. Keep the GH issue open and assigned to track that process.
+* After ~5 Months continue here.
+* Regenerate the test signature for verify_test.go
+  * Create a file that contains `gopass-sign-test\n` and run `gpg -u 0xKEYID --armor --output /tmp/testdata.sig --detatch-sign testdata`. Use the correct KEYID (the one of the NEW key).
+  * Hint: Make sure the input only contains one line break, not two.
+  * Paste the content of /tmp/testdata.sig into the `testSignature` in `verify_test.go`. Make sure all tests pass.
+* Navigate to https://github.com/gopasspw/gopass/settings/secrets/actions
+    * Paste the armored private part of the new key into the existing `GPG_PRIVATE_KEY` secret.
+    * Paste the corresponding passphrase into `PASSPHRASE`.
```
AnomalRoil, line 57: BTW, I'm (no longer? or never was) not a maintainer, so I think you're the only one with access to these things for now.

internal/updater/verify.go
```diff
@@ -6,63 +6,63 @@ import (
...
+3GoLrUPPMgEvjpgOSiKDYvfqa92x+IlWz5rmug2zT5H+/UmizgexyCfRbVlTfi/8
+LgAC95fFvk6mo/s0IwZ4m87whlywFkGYEwmbXGhs29f/qZ7ZJPFOW7BZc8ipvrUe
+rTASZuFDwYIMDaFD/aT9wgn27P/UHsqFW0PbVxm44gS90Q4xTx2XTBmJg4S/3Dwn
+1JZ70RVzsU4kL0tVQ5GDzKvN2SBhHsr5POBTxbrVW1+HATXeRGv0orqccHwmFaPh
+OO4szdamDmhzgr9mdVv0gHg9cyTizvNiH026FYRwJmATPj1sjAnnjscZPKBeKiNO
+fT1TaQbilUs+PL7VNI6d2uAPwQ==
+=bs0I
 -----END PGP PUBLIC KEY BLOCK-----
 `),
+	// 2023 key - 0x560D8522
```
AnomalRoil, line 65: any reason to keep the 2023 key but not the 2024 one?

internal/updater/verify_test.go
```diff
@@ -37,3 +37,22 @@ func TestGPGVerify(t *testing.T) {
...
+	}
+	defer func() {
+		timeNow = func() time.Time {
+			return time.Now()
+		}
+	}()
+
+	ok, err := gpgVerify(testData, testSignature)
+	require.NoError(t, err, "If TestGPGVerify succeeds but this test fails the self-updater key is about to expire. Please open an issue to update the key. Thank you.")
+	assert.True(t, ok)
```
AnomalRoil, line 57: an alternative might be to just rely on:
`func (sig *Signature) SigExpired(currentTime time.Time) bool`
https://github.com/ProtonMail/go-crypto/blob/d703f494c90d2fa80fdbbaf33c084a7225411b55/openpgp/packet/signature.go#L760-L762
for the signature returned by `PrimarySelfSignature`:
https://github.com/ProtonMail/go-crypto/blob/d703f494c90d2fa80fdbbaf33c084a7225411b55/openpgp/keys.go#L887-L892


Something like:
```suggestion
// we only check the latest pubkey
keyList, err := openpgp.ReadArmoredKeyRing(bytes.NewReader(pubkey[0]))
if err != nil || len(keyList) < 1 {
	debug.Log("failed to read public key: %q", err)
	return false, fmt.Errorf("failed to read public key: %w", err)
}

sig, _ := keyList[0].PrimarySelfSignature()
require.True(t, !sig.SigExpired(time.Now().AddDate(0,6,0)), "If TestGPGVerify succeeds but this test fails the self-updater key is about to expire. Please open an issue to update the key. Thank you.")
```

So that we don't need to change `gpgVerify` and use a global variable.

AnomalRoil (changes requested):

internal/updater/verify.go
dominikschulz: There is no 2024 key. When I started this feature I did settle for a 2y expire. Happy to discuss and change that going forward.

internal/updater/verify_test.go
dominikschulz: Great suggestion. I'll try that.

internal/updater/README.md
dominikschulz: Maybe we should change that. But this discussion should not happen inside a PR. Maybe Slack or some other place.

internal/updater/verify.go
AnomalRoil: Ah right, but my question was then:
why delete the key starting with `mQGNBGAH4iEBDAC6ZXN/hzyrB8GA8KdQEasGOrri86GDsyyyRPHP1` instead of storing it like the below one?

internal/updater/README.md
AnomalRoil: Yeah, wasn't intending to discuss that here, but wanted to raise the point that the bus factor/list of core maintainers is relatively low.

internal/updater/verify_test.go
dominikschulz: That didn't seem to work. The self-sig didn't have an expriation in my case. Something else must have been expired (subkey, signature, ...).

internal/updater/verify.go
dominikschulz: Why I wasn't deleting the expired key? Good point. For a proper key rollover we can't do that, but since it already expired I can kill it.

AnomalRoil (approved):

Issue #3034, opened 2025-01-09
show.autoclip with show.safecontent results in empy clip


letfadude:
### Summary
when calling `gopass show -c [key]` with safecontent=true results in password copied to clipboard and not showing in stdout
if configured with autoclip && safecontent and calling `gopass show [key]` results in empty string in clipboard and no stdout
only without setting safecontent autoclip will work correctly but password is shown in stdout 

### Steps To Reproduce
```
# with autoclip and safecontent true

$ gopass show [key]
Secret: [key]

url: [url]
username: [username]

$ xsel -b -o
\ No newline at end of selection

# with safecontent = false and autoclip = true

$ gopass show key
✔ Copied key to clipboard. Will clear in 45 seconds.
Secret: key

[secret]
url: [url]
username: [username]

$ xsel -b -o
[secret]
\ No newline at end of selection
```

### Expected behavior

autoclip and safecontent true should result in password saved to clipboard and not displayed in stdout

### Environment

- OS: Linux Mint 21.3 (Virginia)
- OS version: Linux fadude-tp 5.15.0-130-generic #140-Ubuntu SMP Wed Dec 18 17:59:53 UTC 2024 x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: gopass 1.15.15 go1.23.3 linux amd64
- Installation method: Debian Repo according to documentation

dominikschulz:
I agree the expected behavior.

letfadude:
https://github.com/gopasspw/gopass/blob/b6140bcba065e3ab088822e770f986509cb2951e/internal/action/show.go#L274

if check would be extended to respect not only `IsAlsoClip()` but also `IsClip()`

letfadude:
i will take a closer look at this and submit a PR asap

PR #3053, opened 2025-02-18
[BUGFIX] Fixed password not saving to clipboard with safecontent and autoclip true

letfadude:
Fixed an issue where enabling both AutoClip and SafeContent prevented passwords from being saved to the clipboard. Added tests to cover this case.

Signed-off-by: Alex Wallner <wallneralex7789@gmail.cpm>

### DIFF
--- a/internal/action/show.go
+++ b/internal/action/show.go
@@ -273,7 +273,7 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	// everything but the first line.
 	if config.Bool(ctx, "show.safecontent") && !ctxutil.IsForce(ctx) && ctxutil.IsShowParsing(ctx) {
 		body := showSafeContent(sec)
-		if IsAlsoClip(ctx) {
+		if IsAlsoClip(ctx) || config.Bool(ctx, "show.autoclip"){
 			return pw, body, nil
 		}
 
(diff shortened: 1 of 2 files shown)


letfadude:
fixes #3034

internal/action/show.go
```diff
@@ -273,7 +273,7 @@ func (s *Action) showGetContent(ctx context.Context, sec gopass.Secret) (string,
 	// everything but the first line.
 	if config.Bool(ctx, "show.safecontent") && !ctxutil.IsForce(ctx) && ctxutil.IsShowParsing(ctx) {
 		body := showSafeContent(sec)
-		if IsAlsoClip(ctx) {
+		if IsAlsoClip(ctx) || config.Bool(ctx, "show.autoclip"){
 			return pw, body, nil
 		}
```
AnomalRoil, line 278: I'm guessing this won't display the warning "⚠ show.safecontent=true. Use -f to display password, if any" and instead will just display "✔ Copied <entry> to clipboard. Will clear in 45 seconds.", right?

But I guess that should be fine, wdyt?

AnomalRoil (approved):
thanks for adding a test

internal/action/show.go
letfadude: yes exactly! 
just shows the ✔ Copied line
I think it might be a little more intuitive

dominikschulz (approved):
Thanks for adding the test. Makes this much nicer to review.

PR #3055, opened 2025-02-24
Ensure that otpauths are properly hidden from being displayed in safecontent mode

### LINKED ISSUE #3051: safecontent and otpauth
Hi,

Recently I moved to use `show.safecontent = True`.
I've noticed that when I run `gopass show X`, where `X` is a secret containing an `otpauth`, the otpauth is shown.
Since the otpauth contains the AES secret, I would expect it not to be shown.

Fale:
Fixes #3051


internal/action/show.go
```diff
@@ -306,7 +306,12 @@ func showSafeContent(sec gopass.Secret) string {
 		}
 	}
 
-	sb.WriteString(sec.Body())
+	for _, l := range strings.Split(sec.Body(), "\n") {
+		if strings.HasPrefix(l, "otpauth://") {
+			continue
+		}
```
AnomalRoil, line 312: This isn't behaving exactly like the rest of the `safecontent` since this is currently just suppressing the lines. To be consistent, I think it should be putting `randAsterisk()` to replace what comes after `otpauth://` IMO.

Also, if the otp entry is a key, this won't hide it, we currently support the keys `otpauth`, `hotp` and `totp`:
https://github.com/gopasspw/gopass/blob/01b6ff1dde17dcd139b5d8c8ca3c52808cd79c60/pkg/otp/otp.go#L30-L47

So you might want to add these to the `isUnsafeKey` function to cover all the cases, both when the otp URL is a key and is in the `Body`.

internal/action/show_test.go
```diff
@@ -119,6 +119,16 @@ func TestShowMulti(t *testing.T) {
 		buf.Reset()
 	})
 
+	t.Run("show entry with oauth field with safecontent enabled", func(t *testing.T) {
```
AnomalRoil, line 122: Thanks for adding a test, that's great!

AnomalRoil (changes requested):
A few comments, but LGTM otherwise.

It might warrant editing the docs too to mention this behaviour?

Maybe a little comment in https://github.com/gopasspw/gopass/blob/master/docs/commands/otp.md and maybe at the bottom of https://github.com/gopasspw/gopass/blob/master/docs/commands/show.md#parsing-and-secrets next to the `unsafe-keys` documentation.

Fale:
Thanks a lot for the feedback @AnomalRoil :).
I've improved the PR by both adding the support for OTP strings in YAML kv, as well as I've added the doc.

dominikschulz (approved):

dominikschulz (approved):

Issue #3017, opened 2024-12-29
bug: gopass shell completion and flags


foxyseta:
### Summary

The built-in shell completion for secrets' paths only work when no flags are being used.

### Steps To Reproduce

1. invoke `gopass`
2. type `show -c `
3. start typing the path of one of your secrets
4. press tab for autocompletion

### Expected behavior

The path of your secret is autocompleted, much like it would happen outside of the shell.

### Environment

- OS: Alpine Linux
- OS version: 3.21.0
- gopass Version: gopass 1.15.15 go1.23.3 linux amd64
- Installation method: Alpine Package Keeper (official distro package manager)

GLobyNew:
@dominikschulz I believe this can be closed due merged pull request

PR #3057, opened 2025-02-25
[ENHANCEMENT] Add support for autocompletion with flags in REPL mode

GLobyNew:
Implemented autocompletion with flags in REPL mode for issue #3017

### DIFF
--- a/internal/action/repl.go
+++ b/internal/action/repl.go
@@ -13,6 +13,31 @@ import (
 	"github.com/urfave/cli/v2"
 )
 
+type mindFlagsCompleter struct {
+	base readline.AutoCompleter
+}
+
+func (f *mindFlagsCompleter) Do(line []rune, pos int) ([][]rune, int) {
+	// Get the portion of the line up to the cursor and split to tokens
+	origInput := string(line[:pos])
+	tokens := strings.Fields(origInput)
+	var filtered []string
+	for _, tok := range tokens {
+		// Skip flag tokens
+		if strings.HasPrefix(tok, "-") {
+			continue
+		}
+		filtered = append(filtered, tok)
+	}
+	sanitized := strings.Join(filtered, " ")
+	// Add additional space if had one on original line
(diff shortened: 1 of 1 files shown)


GLobyNew:
Okay, fixing lint is not a problem
Right now it looks like my code manages to not use `configLocations() []string ` func.
I don't understand full logic where it should be used or should it be deleted. 
Can you tip me - should I just remove this function or somewhere my code breaks core functionality?

internal/action/repl.go
```diff
@@ -13,6 +13,31 @@ import (
...
+			continue
+		}
+		filtered = append(filtered, tok)
+	}
+	sanitized := strings.Join(filtered, " ")
+	// Add additional space if had one on original line
+	if len(origInput) > 0 && origInput[len(origInput)-1] == ' ' {
+		sanitized += " "
+	}
+	return f.base.Do([]rune(sanitized), len([]rune(sanitized)))
```
dominikschulz, line 37: We require a blank link before the return statement to make it stand out better.

internal/action/repl.go
```diff
@@ -13,6 +13,31 @@ import (
...
 
+type mindFlagsCompleter struct {
+	base readline.AutoCompleter
+}
+
+func (f *mindFlagsCompleter) Do(line []rune, pos int) ([][]rune, int) {
+	// Get the portion of the line up to the cursor and split to tokens
+	origInput := string(line[:pos])
+	tokens := strings.Fields(origInput)
+	var filtered []string
```
dominikschulz, line 24: Make it `filtered := make([]string, 0, len(tokens))`, please. This avoids resizing the underlying array later. It's probably not performance relevant, but for consistency. Thanks.

dominikschulz (changes requested):

GLobyNew:
@dominikschulz I have rebased my repo with master and commit changes to pass tests.
All tests are passed, except "make codequality": `func configLocations() ` is unused even in master branch.

dominikschulz (approved):
LGTM. Thanks!

PR #3058, opened 2025-02-27
[chore] Add tpl func tests and fix two small issues

dominikschulz:
Timestamps were not rounded correctly and the truncate function was not trunacting correctly.

### DIFF
--- a/internal/tpl/funcs.go
+++ b/internal/tpl/funcs.go
@@ -288,17 +288,17 @@ func roundDuration(duration any) string {
 	second := uint64(time.Second)
 
 	switch {
-	case u > year:
+	case u >= year:
 		return strconv.FormatUint(u/year, 10) + "y"
-	case u > month:
+	case u >= month:
 		return strconv.FormatUint(u/month, 10) + "mo"
-	case u > day:
+	case u >= day:
 		return strconv.FormatUint(u/day, 10) + "d"
-	case u > hour:
+	case u >= hour:
 		return strconv.FormatUint(u/hour, 10) + "h"
-	case u > minute:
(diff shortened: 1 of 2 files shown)


internal/tpl/funcs_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+func TestDate(t *testing.T) {
+	ts := time.Date(2023, 10, 1, 0, 0, 0, 0, time.UTC)
+	assert.Equal(t, "2023-10-01", date(ts))
+}
+
+func TestTruncate(t *testing.T) {
+	assert.Equal(t, "hello...", truncate(8, "hello world"))
+	assert.Equal(t, "hello", truncate(8, "hello"))
+}
```
AnomalRoil, line 102: Might warrant a test for the case `length == 1` and ` len(sv)== 2`

internal/tpl/funcs.go
```diff
@@ -311,7 +311,7 @@ func date(ts time.Time) string {
 
 func truncate(length int, v any) string {
 	sv := strval(v)
-	if len(sv) < length-3 {
+	if len(sv) < length {
```
AnomalRoil, line 314: I suspect the intention was the avoid the risk of `length-3` below on line 318 being negative when e.g. `len(sv) == 2` but `length == 1` ? But even then that would have needed fixing, good catch.

AnomalRoil (changes requested):

internal/tpl/funcs.go
dominikschulz: No, not initially. But I just ran into that as well.

PR #3059, opened 2025-02-28
Add capability to chain the otp to the password. Fixes #3052

### LINKED ISSUE #3052: clipping password+otp
There are many systems that require to put in the password field both the password and the otp.
Is there a way to clip both the password and the otp concatenated?
If not, I think it would be nice to make it possible, since is a fairly common usecase, imho.

Fale:
Signed-off-by: Fabio Alessandro Locati <me@fale.io>


Fale:
I know this is not the suggested implementation by @AnomalRoil, but it seems fairly straightforward implemented like this, and works also with all the other flags that `otp` has :)

dominikschulz (approved):

PR #3060, opened 2025-02-28
Fix go version in go.mod (invalid version '1.23.2')

Denis-Kuso:


Denis-Kuso:
Feel free to ignore this PR. `go install /some/repo` requires the format to be in `major.minor`, yet when I updated the go.mod file to that format and ran `go mod tidy` it restored the go directive and added the toolchain directive. `go mod tidy` adds the patch version as well, even though after (I think) go 1.21 the directive in go.mod needs to be without patch version.

Issue #3037, opened 2025-01-13
Incorrect progress message on single store manual sync


mcnesium:
mcn@x2a ~ gopass sync --store exma
    🚥 Syncing with all remotes ...
    [exma] 
       gitfs pull and push ... OK (no changes)
       done
    ✅ All done
    mcn@x2a ~ gopass --version
    gopass 1.15.15 go1.23.4 linux amd64

It should not say "Syncing with *all remotes*" but instead "Syncing with store [storename]" when specifying a certain store to sync.

dominikschulz:
That shouldn't be a big problem, but it is confusing and we should fix it.

PR #3076, opened 2025-03-03
[UX] Make single store sync more intuitive / verbose

GLobyNew:
This should resolve https://github.com/gopasspw/gopass/issues/3037

### DIFF
--- a/internal/action/sync.go
+++ b/internal/action/sync.go
@@ -104,7 +104,10 @@ func (s *Action) sync(ctx context.Context, store string, isAutosync bool) error 
 		return nil
 	}
 
-	out.Printf(ctx, "🚥 Syncing with all remotes ...")
+	// check if user asked for single store/remote sync or all remote sync
+	if store == "" {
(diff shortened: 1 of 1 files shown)


internal/action/sync.go
```diff
@@ -118,6 +123,7 @@ func (s *Action) sync(ctx context.Context, store string, isAutosync bool) error
 	// sync all stores (root and all mounted sub stores).
 	for _, mp := range mps {
 		if store != "" {
+			out.Printf(ctx, "🚥 Syncing with store/remote \"%v\"...", store)
```
AnomalRoil, line 126:
```suggestion
			out.Printf(ctx, "🚥 Syncing with store/remote %q...", store)
```
%q will display a double-quoted string already.

AnomalRoil (changes requested):
You could use `if numMPs > 0` before the "All done" print to avoid adding a TODO in the code 👍🏻

dominikschulz (approved):

PR #3083, opened 2025-03-09
[FEATURE] Add option -r/--regex to find

carlsmedstad:
If the option `-r/--regex` is supplied to `gopass find`, the argument is treated as a regex pattern instead of a substring.

### DIFF
--- a/internal/action/find.go
+++ b/internal/action/find.go
@@ -56,7 +56,10 @@ func (s *Action) find(ctx context.Context, c *cli.Context, needle string, cb sho
 	}
 
 	// filter our the ones from the haystack matching the needle.
-	choices := filter(haystack, needle, c.Bool("regex"))
+	choices, err := filter(haystack, needle, c.Bool("regex"))
+	if err != nil {
+		return exit.Error(exit.Usage, err, "%s", err)
+	}
 
 	// if we have an exact match print it.
 	if len(choices) == 1 {
@@ -141,12 +144,15 @@ func (s *Action) findSelection(ctx context.Context, c *cli.Context, choices []st
 	}
 }
 
-func filter(l []string, needle string, regex bool) []string {
+func filter(l []string, needle string, regex bool) ([]string, error) {
 	choices := make([]string, 0, 10)
 	for _, value := range l {
 		if regex {
 			matched, err := regexp.MatchString(needle, value)
-			if err == nil && matched {
+			if err != nil {
+				return nil, err
+			}
+			if matched {
 				choices = append(choices, value)
 			}
 		} else if strings.Contains(strings.ToLower(value), strings.ToLower(needle)) {
@@ -154,5 +160,5 @@ func filter(l []string, needle string, regex bool) []string {
 		}
 	}
 
-	return choices
+	return choices, nil
 }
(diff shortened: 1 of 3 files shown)


internal/action/find.go
```diff
@@ -141,10 +141,15 @@ func (s *Action) findSelection(ctx context.Context, c *cli.Context, choices []st
...
 
-func filter(l []string, needle string) []string {
+func filter(l []string, needle string, regex bool) []string {
 	choices := make([]string, 0, 10)
 	for _, value := range l {
-		if strings.Contains(strings.ToLower(value), needle) {
+		if regex {
+			matched, err := regexp.MatchString(needle, value)
+			if err == nil && matched {
+				choices = append(choices, value)
```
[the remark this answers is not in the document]
carlsmedstad: Reviewed by an AI, that's a first for me :D

Addressed in 2aa5f7a6875a13a9bf8f6eb9173f670f0daffeb3.

internal/action/find.go
```diff
@@ -141,13 +144,21 @@ func (s *Action) findSelection(ctx context.Context, c *cli.Context, choices []st
...
-		if strings.Contains(strings.ToLower(value), needle) {
+		if regex {
+			matched, err := regexp.MatchString(needle, value)
+			if err != nil {
+				return nil, err
+			}
+			if matched {
+				choices = append(choices, value)
+			}
+		} else if strings.Contains(strings.ToLower(value), strings.ToLower(needle)) {
```
dominikschulz, line 158: `else if` is highly unusual in Go and at least in a loop rarely needed. I would suggest to handle the regexp case and at the end of that block add a `continue` so the non-regexp part is skipped in that case.

May if would make sense so swap the order of both cases, but I'm sure about that.

internal/action/find.go
```diff
@@ -141,13 +144,21 @@ func (s *Action) findSelection(ctx context.Context, c *cli.Context, choices []st
 	}
 }
 
-func filter(l []string, needle string) []string {
+func filter(l []string, needle string, regex bool) ([]string, error) {
```
dominikschulz, line 147: The variables `regex` and the package `regexp` are a little bit confusing. Maybe you could name the parameter `reMatch` or something like that?

internal/action/find.go
```diff
@@ -141,13 +144,21 @@ func (s *Action) findSelection(ctx context.Context, c *cli.Context, choices []st
...
 	choices := make([]string, 0, 10)
 	for _, value := range l {
-		if strings.Contains(strings.ToLower(value), needle) {
+		if regex {
+			matched, err := regexp.MatchString(needle, value)
```
dominikschulz, line 151: `regexp.MatchString` will compile the regular expression in every cycle of the loop. That could be pretty expensive. Please compile the RE outside of the loop and then use the compiled RE inside the loop. This will also simplify error handling.

dominikschulz (changes requested):
Good idea, thank you. Please address my feedback and then we're good to go.

internal/action/find.go
carlsmedstad: Good point, renamed in 0a129b92fee1714d0de18f42f81ad91c2629e122.

internal/action/find.go
carlsmedstad: Compiling outside of the loop in: ecc5704cb582488e460718e9f7638fbcfaab1a82

Also replaced the `if else` with an early return.

dominikschulz (approved):

PR #3088, opened 2025-03-14
[DEPENDENCY] Update module dependencies to latest versions

GLobyNew:
I don't know if its okay to do. But I looked up the "OpenSSF Scorecard report" of this project: https://securityscorecards.dev/viewer/?uri=github.com/gopasspw/gopass

And was surprised to see, that gopass is affected to 4 known vulnerabilities. 
Turns out they all can be fixed by just upgrading all modules using two commands:

```
go get -u ./...     
go mod tidy
```

So basically after this merge all four vulnerabilities will be fixed and I wonder how often should deps be upgraded? Or every project has it's own policy about it? 
And what policy does gopass follow?

Thanks!


dominikschulz:
@GLobyNew Thank you. But you don't need to bother, we have automation to update dependencies during releases.

It's ok to pull in updates outside of releases, but that's only best effort.

Issue #2599, opened 2023-06-03
`gopass show -c` is slow


UtkarshVerma:
### Summary
Password decryption is slow when invoked with the `-c` flag.

### Steps To Reproduce
![2023-06-03_11-18](https://github.com/gopasspw/gopass/assets/31820255/b6792814-c3fb-405c-99ab-74738c4a70d8)

### Expected behavior
The time taken for decrypting should be similar to `gpg --decrypt`.

### Environment
- OS: Arch Linux
- OS version: Linux metia 6.1.31-1-lts #&#8203;1 SMP PREEMPT_DYNAMIC Tue, 30 May 2023 14:36:16 +0000
x86_64 GNU/Linux
- gopass Version: gopass 1.15.5 go1.20.3 linux amd64
- Installation method: `pacman`

dominikschulz:
I think we had this reported before, I'll try to find the issue.

dominikschulz:
I haven't been able to reproduce this and this report is getting quite old. Feel free to re-open if it's still an issue, but we will need more information to reproduce.

PR #3089, opened 2025-03-21
[BUGFIX] Run clipboard clearing in a separate goroutine for better performance

GLobyNew:
This should fix https://github.com/gopasspw/gopass/issues/2599

I figured out, that the longer timeout for clearing clipboard, the poorer performance of `show -c` command.

The fix is simple - just run `clearClip` in goroutine. 

`time` test before patch:

```
time ./gopass show -c testEntry
✔ Copied testEntry to clipboard. Will clear in 45 seconds.
./gopass show -c testEntry  1.33s user 0.12s system 143% cpu 1.013 total
```

after the patch:

```
time ./gopass show -c testEntry
✔ Copied testEntry to clipboard. Will clear in 45 seconds.
./gopass show -c testEntry  0.18s user 0.06s system 36% cpu 0.667 total
```

Tested on MBP "14 2023 (Apple M3 Pro / 18 GB RAM).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
         env  
 
     - name: Build and Unit Test
-      run: make travis
+      run: make gha-linux
     - name: Integration Test
       run: make test-integration
 
@@ -129,7 +129,7 @@ jobs:
     - run: git config --global user.email foo.bar@example.org
       
     - name: Build and Unit Test
-      run: make travis-windows
+      run: make gha-windows
 
   macos:
     runs-on: macos-latest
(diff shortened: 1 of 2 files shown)


dominikschulz (commented):
This will not work, I think.

When you run `gopass show some/key` it will show the secret and exit the process.

This un-guarded goroutine will just be terminated.

GLobyNew:
@dominikschulz I don't think I understand what you mean 🤔 

This commits only changes `gopass show -c` behaviour.  The `gopass show` without `-c` flag remains unchanged.
Can you please explain what could go wrong?

Issue #2862, opened 2024-04-09
Confusing "need to unlock" message during setup


knuesel:
Running `gopass setup` after a fresh install of 1.15.13:

```
🌟 Welcome to gopass!
🌟 Initializing a new password store ...
🔐 No useable cryptographic keys. Generating new key pair
🧪 Creating cryptographic key pair (gpg) ...
🎩 Gathering information for the gpg key pair ...
🚶 What is your name? XXX
📧 What is your email? XXX
⚠ Do you want to enter a passphrase? (otherwise we generate one for you) [y/N/q]: # typed Enter for default here
✅ Key pair for gpg generated
Passphrase: XXX
⚠ You need to remember this very well!
⚠ 🔐 We need to unlock your newly created private key now! Please enter the passphrase you just generated.
Do you want to export your public key to XXX? [y/N/q]:
```

It says "need to unlock your newly created private key now! Please enter the passphrase you just generated" but then instead of prompting for the passphrase it immediately asks another question about exporting the public key.

(As a side note, I think it would be great to have a hint in the question about exporting the public key to help the user decide...)

### Environment

- OS: Pop!_OS 22.04
- OS version: Linux pop-os 6.6.10-76060610-generic #202401051437~1709085277~22.04~31d73d8 SMP PREEMPT_DYNAMIC Wed F x86_64 x86_64 x86_64 GNU/Linux
- gopass Version: 1.15.13 go1.22.1 linux amd64
- Installation method: gopass repo

PR #3090, opened 2025-03-21
[CLEANUP] Remove unnecessary notice for unlocking private key after generation

GLobyNew:
Okay, this will be funny and small PR 

This pull request addresses issue https://github.com/gopasspw/gopass/issues/2862

Basically as I undestand logic of GnuPG and gopass it goes like this:

1. Regardless if user typed his own passphrase or asked gopass to generate it will store it in cache of `gpg-agent`
2. Agent has default 600 seconds of caching time. So user will be prompted after 10 minutes.

It will be highly unlikely that user will be initiating his first store for more than 10 minutes. 
And even so he will prompted by passphrase entry interface (e.g pinentry-curses), so no additional notices needed.

### DIFF
--- a/internal/action/setup.go
+++ b/internal/action/setup.go
@@ -197,8 +197,6 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
 
 	out.OKf(ctx, "Key pair for %s generated", crypto.Name())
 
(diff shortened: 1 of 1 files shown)


GLobyNew:
Oh well
Okay, that makes things harder. I'll try to find a solution for this to check if password needed for both gpg and age and will arrange different PR

PR #3092, opened 2025-03-26
[FEATURE] Added option --line/-l to show command (#2884)

### LINKED ISSUE #2884: Feature Request: Allow selecting the line to copy
Let's say your using `pass` and migrate to `gopass` because it seems it's in line for your needs. Now let's say your passwords are orginized like this:

```text
password
username/email
website
``` 

And now you want to copy the email or username, you can't unlike in `pass` where you can copy all lines via the param `-c<line>`

saladuh:
When option `--line` / `-l` is set, interprets the key parameter of the command as a line number, 0 indexed, of the secret, instead of a key for a key value.

Also adds config option `show.autolinekey`, that when set to `true` will always interpret the key parameter as a line number, regardless of whether `--line` / `-l` is set or not. Default value is `false`.
This config option can be overridden with `--line=false`.

Closes #2884


internal/action/commands.go
```diff
@@ -64,6 +64,11 @@ func ShowFlags() []cli.Flag {
 			Name:  "chars",
 			Usage: "Print specific characters from the secret",
 		},
+		&cli.BoolFlag{
+			Name:    "line",
+			Aliases: []string{"l"},
+			Usage:   "Interpret the [key] argument as a line number. 0 indexed. Does nothing if <key> value is not provided.",
```
saladuh, line 70: Not sure if we want this to be 1 indexed or 0 indexed?

saladuh (commented):
Not sure if the command should be 0 or 1 indexed for the line number. What's the preference?

AnomalRoil:
I have a meta-question: 
should we really have a new option and a new flag to handle this case or should we support the key being of the format `:(-?\d+)` to represent a line number instead of a key?

E.g. `gopass show entry :10` would show line 10, `gopass show entry :-1` would show the last line, but `gopass show entry key` would show the key-value for `key`.

This _might_ be a breaking change if people are using :1 or :2 as keys currently (which is supported), since it's valid YAML...
We could work around this issue by using the key if there is an exact match?
e.g. this is valid YAML:
```
---
:key: value
:3: test
:-2: test2
```

Opinion @dominikschulz ?

dominikschulz:
@AnomalRoil I like your idea since it is compact and we can work around the edge cases, as you explained. But I'm not opposed to having a flag, a special key syntax or even both.

saladuh:
@AnomalRoil I think that's a great idea. My intention with this PR was to be as close to the request in #2884, which was to be similar to the functionality in `pass`. I didn't want add a new pair of flags to mimic `-c` and `-C` with an int flag for this feature, as urfave/cli cannot overload on the types of flags, so I reused the key field.

I'm happy to work on this functionality (if @AnomalRoil doesn't mind), keeping both the flag and the `:x` syntax or just the latter.
@dominikschulz Are you happy with the implementation in this PR (provided I clean up a few things), or would you prefer to just have the `:x` syntax?

cc @ltecheroffical as you originally requested the feature, any thoughts?

internal/action/commands.go
dominikschulz: Zero indexed, please.

dominikschulz:
The implementation feels a bit too complicated right now. Maybe this is necessary, but please help me understand the reasoning.

* You are introducing two new context keys just for passing around the information form the --line flag in the same file. Can this not simply be an argument or a helper func that is called with the cli.Context as an argument?
* The line flag is a bool. So if you want to use the key as an offset you would need to call `gopass show --line secret/name 2`. That's quite long. What about making the `--line` flag be an int flag, i.e. `gopass show --line=2 secret/name`?
* I like the idea of using either `:N` or just `N` as the key, trying to parse it and if that works use it as a line number instead of a key. We might also warn if there is an ambiguity.

dominikschulz:
Closing due to inactivity. Feel free to reopen if you intent to finish this.

PR #3104, opened 2025-04-02
[chore] Migrate to golangci-lint v2

dominikschulz:
Migrating golangci-lint to the v2 branch. This requires some config changes and a few code changes to make the updated set of checks pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/gopasspw/gopass
 
-go 1.23.2
+go 1.24.0
 
 require (
 	filippo.io/age v1.2.1-0.20240618131852-7eedd929a6cf
--- a/internal/audit/audit_test.go
+++ b/internal/audit/audit_test.go
@@ -52,7 +52,7 @@ func TestBatch(t *testing.T) {
 
 	require.NoError(t, err)
 	assert.NotNil(t, report)
-	assert.Equal(t, len(secrets), len(report.Secrets))
+	assert.Len(t, len(secrets), len(report.Secrets))
 }
 
 func TestAuditSecret(t *testing.T) {
--- a/internal/audit/audit_test.go
+++ b/internal/audit/audit_test.go
@@ -52,7 +52,7 @@ func TestBatch(t *testing.T) {
 
 	require.NoError(t, err)
 	assert.NotNil(t, report)
-	assert.Len(t, len(secrets), len(report.Secrets))
+	assert.Len(t, secrets, len(report.Secrets))
 }
 
 func TestAuditSecret(t *testing.T) {
--- a/internal/audit/audit_test.go
+++ b/internal/audit/audit_test.go
@@ -32,7 +32,7 @@ func (m *mockSecretGetter) Concurrency() int {
 }
 
 func TestNewAuditor(t *testing.T) {
-	ctx := context.Background()
+	ctx := t.Context()
 	s := &mockSecretGetter{}
 	a := New(ctx, s)
 
@@ -43,7 +43,7 @@ func TestNewAuditor(t *testing.T) {
 }
 
 func TestBatch(t *testing.T) {
-	ctx := context.Background()
+	ctx := t.Context()
 	s := &mockSecretGetter{}
 	a := New(ctx, s)
 
@@ -56,7 +56,7 @@ func TestBatch(t *testing.T) {
 }
 
 func TestAuditSecret(t *testing.T) {
-	ctx := context.Background()
+	ctx := t.Context()
 	s := &mockSecretGetter{}
 	a := New(ctx, s)
 
@@ -67,7 +67,7 @@ func TestAuditSecret(t *testing.T) {
 }
 
 func TestCheckHIBP(t *testing.T) {
-	ctx := context.Background()
+	ctx := t.Context()
 	s := &mockSecretGetter{}
 	a := New(ctx, s)
 
--- a/internal/backend/crypto/gpg/gpgconf/version_test.go
+++ b/internal/backend/crypto/gpg/gpgconf/version_test.go
@@ -44,7 +44,7 @@ func TestSort(t *testing.T) {
 
 			sort.Sort(byVersion(tc.in))
 
-			require.Len(t, len(tc.in), len(tc.out))
+			require.Len(t, tc.in, len(tc.out))
 			for i, v := range tc.out {
 				if !tc.in[i].ver.Equals(v) {
 					t.Errorf("wrong sort order at %d: %s != %s", i, tc.in[i].ver, v)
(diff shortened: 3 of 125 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/gopasspw/gopass
 
-go 1.23.2
+go 1.24.0
 
 require (
```
AnomalRoil, line 5: I'm surprised you got away not adding the annoying toolchain

internal/audit/audit_test.go
```diff
@@ -52,7 +52,7 @@ func TestBatch(t *testing.T) {
 
 	require.NoError(t, err)
 	assert.NotNil(t, report)
-	assert.Equal(t, len(secrets), len(report.Secrets))
+	assert.Len(t, len(secrets), len(report.Secrets))
```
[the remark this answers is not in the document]
AnomalRoil: This looks like a legit comment from the bot.

internal/backend/crypto/gpg/gpgconf/version_test.go
```diff
@@ -44,7 +44,7 @@ func TestSort(t *testing.T) {
 
 			sort.Sort(byVersion(tc.in))
 
-			require.Equal(t, len(tc.in), len(tc.out))
+			require.Len(t, len(tc.in), len(tc.out))
```
AnomalRoil, line 47: This looks suspicious, maybe you meant 
```suggestion
			require.Len(t, tc.out, len(tc.in))
```

See https://pkg.go.dev/github.com/stretchr/testify/require#Len

internal/backend/storage/fossilfs/loader_test.go
AnomalRoil, line 1: Are we really still supporting `fossilfs` and people are actually using it, btw? Might be worth putting a deprecation proposal for it if we're not sure it's still working in 2025.

AnomalRoil (changes requested):

internal/config/legacy/config.go
```diff
@@ -88,12 +91,16 @@ func (c *Config) setConfigValue(key, value string) error {
 
 			return nil
 		case reflect.Bool:
-			switch {
-			case value == "true" || value == "on":
+			switch value {
+			case "true":
+				fallthrough
```
AnomalRoil, line 96: Nice, I always forget it exists xD

internal/store/root/mount.go
```diff
@@ -213,17 +213,11 @@ func (r *Store) checkMounts() error {
 // CleanMountAlias removes all leading and trailing slashes from a mount alias.
 // Note: Slashes inside the alias are valid and will be kept.
 func CleanMountAlias(alias string) string {
-	for {
-		if !strings.HasPrefix(alias, "/") && !strings.HasPrefix(alias, "\\") {
-			break
-		}
+	for strings.HasPrefix(alias, "/") || strings.HasPrefix(alias, "\\") {
 		alias = strings.TrimPrefix(strings.TrimSuffix(alias, "/"), "/")
 		alias = strings.TrimPrefix(strings.TrimSuffix(alias, "\\"), "\\")
```
AnomalRoil, line 218:
```suggestion
		alias = strings.Trim(alias, "/")
		alias = strings.Trim(alias, "\\")
```

internal/store/root/mount.go
```diff
@@ -213,17 +213,11 @@ func (r *Store) checkMounts() error {
...
 	}
-	for {
-		if !strings.HasSuffix(alias, "/") && !strings.HasSuffix(alias, "\\") {
-			break
-		}
+	for strings.HasSuffix(alias, "/") || strings.HasSuffix(alias, "\\") {
 		alias = strings.TrimSuffix(strings.TrimPrefix(alias, "/"), "/")
 		alias = strings.TrimSuffix(strings.TrimPrefix(alias, "\\"), "\\")
```
AnomalRoil, line 222:
```suggestion
		alias = strings.Trim(alias, "/")
		alias = strings.Trim(alias, "\\")
```

pkg/passkey/passkey.go
```diff
@@ -14,17 +14,17 @@ import (
...
 	UserVerified    bool
 	AttestationData bool
 	ExtensionData   bool
 }
 
-// Structure to store information and key of public key credential.
+// Credential is a structure to store information and key of public key credential.
 type Credential struct {
-	Id        string
+	ID        string
```
AnomalRoil, line 27: That's a breaking changes in our public API that might warrant at least minor version bump

AnomalRoil: but tbh it's unused: https://pkg.go.dev/github.com/gopasspw/gopass@v1.15.15/pkg/passkey?tab=importedby 
So it's fine.

AnomalRoil (changes requested):
Pfiou, that was a massive PR, but reviewed 103/103 files  👍🏻

internal/backend/storage/fossilfs/loader_test.go
dominikschulz: It was a test to see if we could have an alternative to Git. But it had some annoying bugs ... we can deprecate it.

dominikschulz:
@AnomalRoil Thanks a lot for reviewing this. I know it's a lot. I was surprised as well by the amount of changes required for this "simple" migration. But there were also a number of regressions that creeped in over time. This should get us to a mostly stable state wrt. tests again. Only some gpg flakieness still remains.

internal/store/root/mount.go
dominikschulz: We only want to remove leading and trailing slashes. strings.Trim does also remove them within the string.

internal/store/root/mount.go
dominikschulz: See my comment above.

internal/store/root/mount.go
AnomalRoil: Its documentation says otherwise:
> Trim returns a slice of the string s with all leading and trailing Unicode code points contained in cutset removed

dominikschulz: Oh. Maybe I was wrong 🫤

pkg/passkey/passkey.go
sudoforge: well, fine... except for all of the private packages using it! :P

IMO, _any_ breaking change should necessitate a major version bump. mildly irritating for a case like this, but this change was not made for formatting reasons (which was why this tree existed), so it could have simply been reverted.

dominikschulz: We did have this discussion when this project was initially published and we have documented our stance in the FAQ: https://github.com/gopasspw/gopass/blob/master/docs/faq.md#api-stability

sudoforge: makes sense to me! i wasn't aware of the documented policy, and this change didn't impact me -- i was simply commenting on it to highlight that it _could_ have broken things for someone who doesn't explicitly version their usage of this.

PR #3108, opened 2025-04-06
fix: downgrade golangci-lint-action to v6.5.2

sudoforge:
Commit 55c2c1448f86e01eaae002a5a3a9624417608d84 upgraded this dependency
to `v7.0.0`, which only supports `v2` of `golangci-lint`. This change
downgrades the action to `v6.5.2` in order to fix the pipeline.


sudoforge:
it looks like additional changes have been made in an attempt to bring the codebase inline with `golangci-lint:v2` in backwards-incompatible ways, which means that this downgrade will never pass.

@dominikschulz has a WIP tree for bringing the codebase in line with `v2` here: #3104

closing.

PR #3113, opened 2025-04-14
Allow connections to Ubuntu repository hosts

dominikschulz:
I'm not 100% sure where these connections are being initiated, but this is an Ubuntu image and those are legit Ubuntu hosts.

Triggered by an Harden Runner email notification I got.


AnomalRoil (approved):
LGTM, probably caused by the change of GHA ubuntu-latest version from 22.04 to 24.04 3 months ago: https://github.com/actions/runner-images/commit/976232da217887825e7541c2635bf39cbbf22654

PR #3122, opened 2025-04-17
[chore] Migrate goreleaser config to v2

dominikschulz:
Minor changes that should have no visible effect.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -119,7 +119,7 @@ test-integration: $(GOPASS_OUTPUT)
 crosscompile:
 	@echo ">> CROSSCOMPILE"
 	@which goreleaser > /dev/null; if [ $$? -ne 0 ]; then \
-		$(GO) install github.com/goreleaser/goreleaser@latest; \
+		$(GO) install go install github.com/goreleaser/goreleaser/v2@latest; \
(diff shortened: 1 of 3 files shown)


dominikschulz:
The zsh completion change is a side-effect of a `gorelease release --snapshot` run. But I think it's ok to include it here.

Makefile
```diff
@@ -119,7 +119,7 @@ test-integration: $(GOPASS_OUTPUT)
 crosscompile:
 	@echo ">> CROSSCOMPILE"
 	@which goreleaser > /dev/null; if [ $$? -ne 0 ]; then \
-		$(GO) install github.com/goreleaser/goreleaser@latest; \
+		$(GO) install go install github.com/goreleaser/goreleaser/v2@latest; \
```
AnomalRoil, line 122: That looks like it would turn into
``` 
go install go install github.com/goreleaser/goreleaser/v2@latest
```

AnomalRoil (changes requested):

Makefile
dominikschulz: Yes, indeed. Copy & Paste mistake :)

PR #3159, opened 2025-05-20
[fix] avoid length prompt when input is within rule boundary

sudoforge:
This change fixes a bug with the generatePasswordForRule() function,
so that users who have provided initial values that are within the
bounds of the domain's rule are not prompted for the length, and rejects
all invalid length values by recursively prompting the user for a valid
length.

### DIFF
--- a/internal/action/generate.go
+++ b/internal/action/generate.go
@@ -167,7 +167,7 @@ func hasPwRuleForSecret(ctx context.Context, name string) (string, pwrules.Rule)
 // generatePassword will run through the password generation steps.
 func (s *Action) generatePassword(ctx context.Context, c *cli.Context, length, name string) (string, error) {
 	if domain, rule := hasPwRuleForSecret(ctx, name); domain != "" && !c.Bool("force") {
-		return s.generatePasswordForRule(ctx, c, length, name, domain, rule)
+		return s.generatePasswordForRule(ctx, length, domain, rule)
 	}
 
 	cfg, mp := config.FromContext(ctx)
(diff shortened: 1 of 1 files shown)


sudoforge:
i wasn't able to find existing tests for this method, and given the potential for (and desire to test) interaction, it's not immediately clear to me if you have an existing pattern for that.

point me in the right direction if you do have an existing solution for testing w/ input prompts :)

sudoforge:
the commit subject here sort of reads like it should be a feature, however, i consider the current behavior that a user is prompted for a length, even if they provide a length that's within the min and max boundaries for a given rule, to be a bug. let me know if you'd like the commit subject changed.

internal/action/generate.go
```diff
@@ -261,21 +262,36 @@ func clamp(mi, ma, value int) int {
...
+		)
+
+		question := fmt.Sprintf(
+			"How long should the password be? (min: %d, max: %d)",
+			rule.Minlen, rule.Maxlen,
+		)
+
+		if iv, err = termio.AskForInt(ctx, question, wl); err != nil {
+			return "", exit.Error(exit.Usage, err, "password length must be a number")
+		}
```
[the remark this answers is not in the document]
sudoforge: well, it's only infinite until the user enters a valid value. i think that's probably okay, right?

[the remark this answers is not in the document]
sudoforge: i guess we could get fancy and redraw the same line, coloring the min/max values, progressing from green to red to communicate "c'mon now, really?"

i'm not sure offhand if that would restrict the shells or terminals we can run in, though.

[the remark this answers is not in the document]
sudoforge: i added this into the tree in 2c79046.

sudoforge:
... and there, i'm done refactoring the title and commit body (for now). i settled on calling this a fix, and i think the updated language makes it more clear that the previous behavior was unexpected.

again, though, if you want me to prefix this with `[feat]` and phrase it as a feature, i'm happy to.

internal/action/generate.go
```diff
@@ -261,21 +261,40 @@ func clamp(mi, ma, value int) int {
...
+			rule.Minlen, rule.Maxlen,
+		)
+
+		var sv string
+
+		if sv, err = termio.AskForString(ctx, question, strconv.Itoa(wl)); err != nil {
+			return "", err
+		}
+
+		return s.generatePasswordForRule(ctx, sv, domain, rule)
```
AnomalRoil, line 295: That's a recursive call? Why can't you just carry on with the rest of the function?
Is it in order to prevent the user from inputting an invalid length, so you'd just keep asking if they do so?

internal/action/generate.go
```diff
@@ -261,21 +261,40 @@ func clamp(mi, ma, value int) int {
...
+	}
+
 	iv = clamp(rule.Minlen, rule.Maxlen, iv)
```
AnomalRoil, line 298: I think this is never something else than `iv` itself given the above `if` conditions, no?

AnomalRoil (changes requested):

internal/action/generate.go
sudoforge: yes. see the initial discussion: https://github.com/gopasspw/gopass/pull/3159#discussion_r2100552140

AnomalRoil: I think it warrants a little comment about why it's recursive in the code itself for future maintainer working on this not to have to dig commits / PRs.

internal/action/generate.go
sudoforge: you're right, good catch. this can be removed.

internal/action/generate.go
sudoforge: i described it in the commit body (`git-blame`, `git-log -L`) but can definitely add a comment just before this return.

internal/action/generate.go
```diff
@@ -261,21 +261,40 @@ func clamp(mi, ma, value int) int {
...
 
-	question := fmt.Sprintf("How long should the password be? (min: %d, max: %d)", rule.Minlen, rule.Maxlen)
-	iv, err := termio.AskForInt(ctx, question, wl)
-	if err != nil {
+	if iv, err = strconv.Atoi(length); err != nil {
 		return "", exit.Error(exit.Usage, err, "password length must be a number")
 	}
 
+	wl := clamp(rule.Minlen, rule.Maxlen, iv)
+	if wl != iv {
```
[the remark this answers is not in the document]
sudoforge: sure - this was also used in the new prompt (probably erroneously); i changed that to `rule.Maxlen` instead, so that this could be removed.

internal/action/generate.go
sudoforge: i added a comment above this statement in d9364941312feebadaba7d6b91c9df98c82bce49.

dominikschulz (approved):
Tested locally. I think it's nice UX now, thanks.

PR #3160, opened 2025-05-20
[testing] use `/usr/bin/env cat` instead of `/bin/cat`

sudoforge:
Hard-coding paths to binaries like `/bin/<foo>` fails on platforms like
NixOS, which do not follow the typical directory hierarchy found on
Linux distributions. Using `/usr/bin/env cat` is much more portable,
however, it delegates path searching to the shell environment, which
**can** be a cause for concern in domains with high security
requirements.


sudoforge:
i'm targeting this one use case of hard coding paths as this is what bit me in local testing of #3159, but there are several other hard-coded paths that i'd be happy to include in this tree if you'd like (some of these are `/usr/local/bin` et. al, and a few are actually `/usr/bin/env`, just including this for a high level overview of potential references):

```
➜ rg '\/bin' --files-with-matches --count-matches 
Dockerfile:5
Makefile:1
fish.completion:1
docs/setup.md:4
pkg/passkey/passkey.go:1
docs/commands/edit.md:2
docs/usecases/gpaste.md:2
internal/editor/edit_others_test.go:1
internal/editor/edit_windows_test.go:1
internal/completion/fish/template.go:1
internal/completion/fish/completion_test.go:1
helpers/proxy/Dockerfile.debian:1
helpers/release/main.go:3
internal/backend/crypto/gpg/gpgconf/utils_others.go:1
internal/backend/crypto/gpg/gpgconf/version_test.go:3
internal/action/env_test.go:2
internal/action/generate_test.go:1
```

dominikschulz (approved):

PR #3162, opened 2025-05-22
[chore] Add tests and comments for hasPwRuleForSecret

dominikschulz:
Mostly to remind myself how those are supposed to operate and document it for my future self.

The lookup part is a little bit error prone, i.e. if the rules for these domains change the test will break. But I promise to take care of that once that happens for the first time.


AnomalRoil (approved):
LGTM

Issue #3148, opened 2025-05-03
Question about workflow


1995parham:
I have some accounts that they are using some SSO provider for their username and password. Consider that websites `ws1` and `ws2` both uses my email username and password for login, but they have different TOTP. I want to ask about your suggestion in this situation.

sudoforge:
hey @1995parham! i'm just another user, not a team member, but i saw this and wanted to leave my thoughts.

a little bit tongue-in-cheek: tell `ws1` and `ws2` to learn what "SSO" means :P

it's **single** sign on - there should be a central identity server that you log into, and `ws1` and `ws2` should get a valid token back.

---

unfortunately, this is a little tricky due to the way your applcations have decided to implement their authentication. the `otp` implementation doesn't support keeping multiple `otpauth` URIs in a single entry, so you have pretty limited options, neither of which is ideal:

- create multiple entries and duplicate all information, or
- create `ws1` with your username and password, and `ws2` with _just_ the TOTP URI, and remember to `gopass otp [-c] ws2` when you need that

this situation touches on something (disconnected from this issue) that i've been thinking about for a while: special URIs that allow for inline references to other entries. that is, something like this:

```
# ws2
pass://ws1
username: pass://ws1:username
otpauth://totp?secret=CUSTOM_TOTP_FOR_WS2
```

where the special URI `pass://` would cause `gopass` to get the value from another entry in the store.

this hyopthetical feature would support this use case quite well (although, i came up with this for other reasons). that said, i don't think this is something on the gopass team's radar, and i should note that it explicitly may not ever be something that they want to support, because of the goal to maintain compatibility with `zxc24/pass` (although i don't know if this is still an ongoing goal).

dominikschulz:
@sudoforge I like your idea for the special `pass://` (or maybe `gopass://`) scheme. If we can make sure it's safe that should be absolutely possible to implement.

As for `pass` compatibility: My main goal is interoperability and to avoid vendor lock-in (we are not a commercial product after all). But anything that still allows for a reasonable migration - and I think that's the case here - is fine with me.

1995parham:
@sudoforge Thanks for your response. I know our company SSO setup is not good at all, but I really enjoyed your solution.
@dominikschulz Thanks for getting back to this ticket. If you think, is it acceptable then I can try to do it in a PR.

sudoforge:
> so, in the implementation that field will run a sub-process and use the output 🤔

i would strongly discourage this.

[...]

PR #3163, opened 2025-05-23
Handle referencing in passwords

1995parham:
Based on the discussion we had in #3148. I've added the referencing feature, so you can defined your passwords as follows:

```
gopass://ws1
username: parham.alvani@snapp.cab
```

Which means password is referencing another secret named `ws1` which you can have another parameters redefined here for example `url`, `username` or `otpauth` could be different.

### DIFF
--- a/internal/action/commands.go
+++ b/internal/action/commands.go
@@ -19,10 +19,6 @@ func ShowFlags() []cli.Flag {
 			Usage:   "Always answer yes to yes/no questions",
 		},
 		&cli.BoolFlag{
-			Name:  "no-follow-ref",
-			Usage: "do not following the gopass:// referenced password",
-		},
-		&cli.BoolFlag{
 			Name:    "clip",
 			Aliases: []string{"c"},
 			Usage:   "Copy the password value into the clipboard",
--- a/internal/action/commands.go
+++ b/internal/action/commands.go
@@ -19,6 +19,10 @@ func ShowFlags() []cli.Flag {
 			Usage:   "Always answer yes to yes/no questions",
 		},
 		&cli.BoolFlag{
+			Name:  "no-follow-ref",
+			Usage: "do not following the gopass:// referenced password",
+		},
+		&cli.BoolFlag{
 			Name:    "clip",
 			Aliases: []string{"c"},
 			Usage:   "Copy the password value into the clipboard",
--- a/docs/config.md
+++ b/docs/config.md
@@ -5,7 +5,7 @@
 Some configuration options are only available through setting environment variables.
 
 | **Option**                   | **Type** | **Description**                                                                                                                                                   |
-|------------------------------|----------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| ---------------------------- | -------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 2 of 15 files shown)


sudoforge (commented):
thanks for picking this up! i tested it locally and it seems to work (example session showocasing this below).

```
➜ export GOPASS_HOMEDIR=$(mktemp -d)


➜ git rev-parse HEAD
514f7b68a1aabb93fb82c99661a31942d9024bc4


➜ go build .


➜ ./gopass generate -g xkcd --xs ' ' --print foo 4
✅ Password for entry "foo" generated
⚠ The generated password is:

gangrene freezable apprehend massager


➜ ./gopass insert --echo bar
Enter password for bar []: gopass://foo
Retype password for bar []: gopass://foo


➜ gpg -d $GOPASS_HOMEDIR/.local/share/gopass/stores/root/bar.gpg 2>/dev/null
gopass://foo


➜ ./gopass show -o bar
gangrene freezable apprehend massager
```

---

some (personal) thoughts:

- `pass://` is less characters to type, but i won't belabour my desire for this too much, and will definitely leave it up to the team
- i think some tests should exist for this

support for these things don't exist in this tree, but i'll be happy to add it in later, when i have some time:
- referencing keys (e.g. `gopass://foo:some-key`)
- references as the value for a key (e.g. `bar.some-key = gopass://foo:some-key`)

---

the current experience when an invalid entry is referenced is a little odd.

```
➜ ./gopass insert --echo bar
An entry already exists for bar. Overwrite it? [y/N/q]: y
Enter password for bar []: gopass://abc
Retype password for bar []: gopass://abc


➜ gpg -d $GOPASS_HOMEDIR/.local/share/gopass/stores/root/bar.gpg 2>/dev/null
gopass://abc


➜ ./gopass show -o bar
⚠ Entry "bar" not found. Starting search...
✅ Found exact match in "bar"

Error: failed to retrieve secret "bar": entry is not in the password store
```

internal/action/commands.go
```diff
@@ -18,6 +18,10 @@ func ShowFlags() []cli.Flag {
 			Aliases: []string{"y"},
 			Usage:   "Always answer yes to yes/no questions",
 		},
+		&cli.BoolFlag{
```
dominikschulz, line 21: Rather than a flag this should be a config option, I think.

dominikschulz (commented):
I like the implementation so far, thanks!

But I'll have to give it a try to see how well it blends in.

internal/action/commands.go
dominikschulz: Good point. For edit for example I think we shouldn't replace the reference. Otherwise that's an confusing user experience, isn't it?

1995parham:
I applied the comments you mentioned. Right now there is a configuration and no flag. Also, following references always disabled on `edit'.

1995parham:
Oh, I saw configuration tests are failing. I will fix it.

dominikschulz (dismissed):

dominikschulz:
@1995parham Can you take a look at the linux build, please? There are some legit (minor) issues. But the CodeQL change you can ignore. That's an false positive.

dominikschulz (approved):

dominikschulz (approved):

PR #3200, opened 2025-08-18
[fix] use WritePassword for secure write

wwwjfy:
So that for systems that support it, passwords won't be captured by clipboard managers.

### DIFF
--- a/pkg/clipboard/clipboard.go
+++ b/pkg/clipboard/clipboard.go
@@ -68,7 +68,7 @@ func CopyTo(ctx context.Context, name string, content []byte, timeout int) error
 }
 
 func copyToClipboard(ctx context.Context, content []byte) error {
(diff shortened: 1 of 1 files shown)


wwwjfy:
What's missing to merge this? The failure on windows should not be related to the change.

dominikschulz:
@wwwjfy No, the windows failure is not related. Actually I tried to fix those first but that turned out to take more time that I had at my disposal. Could you please make sure your commit has a "Signed-Off" line ([DCO](https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---signoff)) and rebase? Then all tests should pass and we can merge this.

wwwjfy:
Got it. Updated :)

TM2500:
@dominikschulz: I'm currently inclined to revert this change until the above mentioned issue in gopasspw/clipboard is resolved.
At least if you plan to release a new version soon.

PR #3269, opened 2025-10-05
x3 optimization for hash

sarff:
- Replaced fmt.Sprintf("%x") with hex.EncodeToString() in all functions
- Added warnings for MD5Hex and SHA1Hex that these algorithms are cryptographically broken

Bench:
before 
image

after
image

😁


sarff:
@dominikschulz Done
thx

PR #3275, opened 2025-10-06
Fixing GPG ID related issues

### LINKED ISSUE #2915: Bug with using multiple gpg keys. .gpg-id is ignored.
### Summary
`gopass` does not honour .gpg-id, and tries to use multiple gpg keys to sign inserted passwords.
The same behavior is not present for `pass`

### Steps To Reproduce
create gpg key_1 with password
create gpg key_2 without password

pass init repository with key_1
pass init -p sub_folder with key_2

`cat ~/.password-store/.gpg-id` shows only key_1
`cat ~/.password-store/sub_folder/.gpg-id` shows only key_2

`pass insert personal/example/first`
`gpg -d personal/example/first.gpg` and then insert wrong passwords. descryption will fail. (Expected)

`gopass insert personal/example/second`
`gpg -d personal/example/second.gpg` and then insert wrong passwords. access will be defined for gpg key_1 (Expected), but key_2 will be used as fallback and secret shown anyway (Not expected!)

### Expected behavior
I expected `gopass` to sign my inserted password in not subfolder, only with `key_1` and fail on inserting wrong passwords for key_1. But it used as fallback my unprotected key_2

### Environment
pass version: v1.7.4
gopass version: gopass 1.15.13-git+HEAD go1.22.5 linux amd64
os version: VERSION="22.04.3 LTS (Jammy Jellyfish)"
installation method: `go install github.com/gopasspw/gopass@latest`

### LINKED ISSUE #3253: force gopass to use special recipient with xxxxxxxx!
my .gpg-id file like this

xxxxxxxx!

and i have multiple subkeys. just want use xxxxxxxx key for my pass store.

when use pass add a test password
gpg --list-packets test.gpg 
command show only the xxxxxxxx subkey used which is what i want

with gopass after add a test password
gpg --list-packets test.gpg
show all my local subkeys?

how can i force gopass to only use the subkey i defined in .gpg-id like pass do?

AnomalRoil:
This fixes #3253 and fixes #2915 however I could not find a proper way of testing neither, because the way our GPG-related tests are setup are fairly remote from our code to parse or manage GPG bytes. 
I have reverted my test commits since they didn't actually test properly the behaviour since I couldn't extract the exact subkey recipients from the ciphertexts.

I have tested it locally and it does solve both...

This also adds a lot of useful debug logs for these kind of issues.

However an unexpected side discovery was that  `IsCheckRecipients` was actually always set to `false`, which most notably means that `func (s *Store) useableKeys` was a no-op. This is because `WithCheckRecipients` was never called with `true` and its default value is `false`...
I have thus removed these useless functions, but this caused disruptions in some of our tests because of the change of behavior of `useableKeys` that suddenly did check for recipients being available in the crypto backend... so I have defaulted to `Warning` about it when none are found. 
So this in theory should cause some sort of change in the behaviour, that I sadly haven't been able to pinpoint more precisely.


internal/backend/crypto/gpg/cli/recipients.go
```diff
@@ -45,6 +46,28 @@ func (g *GPG) FindRecipients(ctx context.Context, search ...string) ([]string, e
 	if gpg.IsAlwaysTrust(ctx) {
 		recp = kl.Recipients()
 	}
+	debug.Log("recp before subkey check: %q", recp)
+
+	// let us support the ! syntax that enforces specific subkey usage
```
dominikschulz, line 51: This is bit too deeply nested for my taste.

I believe we can limit nesting by inverting that match conditions and using `continue` for the non-match cases.

Something like this (pseudo):

```
for _, val := range search {
  str := ...
  if == "" || !strings.HasSuffix(val, "!") {
    continue
  }
  for _, key := range kl {
    for sub := range key.SubKeys {
      if !strings.Contains(...) {
        continue
      }
      recp = slices ...
```

internal/backend/crypto/plain/backend.go
```diff
@@ -63,6 +63,7 @@ func (m *Mocker) FindRecipients(ctx context.Context, keys ...string) ([]string,
 	res := make([]string, 0, len(rs))
 	for _, r := range rs {
 		for _, needle := range keys {
+			debug.Log("checking recipient %q = %q", r, needle)
```
dominikschulz, line 66: Make that at least `debug.V(1)`, please.

dominikschulz (changes requested):
Thanks a lot. I will need to do another round of review and testing later. But a few initial comments.

AnomalRoil:
Yeah, I probably should have done the `IsCheckRecipients` changes in another PR since that was a somewhat unrelated finding and has some weird consequences (knowing that some of our code was basically dead code since it was always `false`)

PR #3306, opened 2025-12-26
fix: use clipboard.WritePassword on darwin

### LINKED ISSUE #3305: Gopass leaks password into MacOS password history
I just noticed that gopass is leaking copied password into the MacOS password history. A quick search showed that this issue was fixed back in 2021 by #1816, but from what I can tell was reintroduced as a regression via [f4ee40ce23ff9e06dbd0ac41fb593f2b4bc286a7](https://github.com/gopasspw/gopass/commit/f4ee40ce23ff9e06dbd0ac41fb593f2b4bc286a7#diff-48d42d0dfa23a922f8b18f4b8fdc605d956c1ead5aa647de354c4eca1c40849eR71). I saw that there was an attempt to fix it, which was reverted in [8522d7d677e8a03d6955891fbddbf83f2a33d0e9](https://github.com/gopasspw/gopass/commit/8522d7d677e8a03d6955891fbddbf83f2a33d0e9). 

From what I can tell by the issue, this appears to be a problem only for Linux. Would it make sense to switch to the `WritePassword` variant for MacOS to prevent leaking passwords to the clipboard history?

patrickpichler:
In order to prevent leaking passwords to the clipboard history on darwin, `clipboard.WritePassword` is being used to copy passwords.

Fixes #3305

### DIFF
--- a/pkg/clipboard/clipboard.go
+++ b/pkg/clipboard/clipboard.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"runtime"
 
 	"github.com/fatih/color"
 	"github.com/gopasspw/clipboard"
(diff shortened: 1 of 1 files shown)


dominikschulz:
Thanks! I would rather use conditional build tags for this.

PR #3357, opened 2026-03-14
feat: warn when system keyring is unavailable

acaloiaro:
## Summary

- Emit a one-time warning when `age.usekeychain` is enabled but no OS keyring provider is available, so users understand why passphrase caching won't persist across sessions
- Warn when writing a passphrase to the OS keyring fails at runtime
- The `cacher.Set` and `askPass.Passphrase` signatures now accept a `context.Context` to support the warning output
- Add tests for `askPass` initialization, `osKeyring` get/set paths, and the user-facing warning

### DIFF
--- a/internal/backend/crypto/age/askpass.go
+++ b/internal/backend/crypto/age/askpass.go
@@ -4,10 +4,12 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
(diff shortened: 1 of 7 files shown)


dominikschulz (changes requested):
LGTM. Please fix the lint issue and then we're good.

PR #3383, opened 2026-04-06
Cleaning up bugs and potential security issues

dominikschulz:
This PR addresses a comprehensive code quality audit of the gopass codebase, fixing security vulnerabilities, confirmed bugs, CLI/UX issues, and implementing several new features and refactors.

**Security Fixes**
* [H-1] Restrict template engine secret access and fix error message information leakage
* [H-2] Bound symlink walk to the store root to prevent directory traversal
* [H-3] Validate that the editor binary exists before invocation
* [C-1] Fix path traversal vulnerability in the fs storage layer
* [C-2] Document and emit warnings on secret exposure via the env command
* [M-3] Add thread safety (mutex) to the SSH identity cache

**Bug Fixes**
* Fix all confirmed bugs identified in the code quality report
* Fix CLI/UX issues identified in the code quality audit
* Fix Idle() TOCTOU race and Add() post-Close() panic in the queue
* Resolve all stale TODOs flagged in the audit (S-5)
* Address issues A-5, S-1, S-2, S-3, S-4, S-6, and I-1 from the audit

**New Features**
* [I-2] Structured, documented exit codes
* [I-3] Machine-readable JSON output support
* [I-4] New gopass doctor diagnostic command
* [I-7] Fully document the wizard/create template format
* Add configurable show.hidden-keys for fine-grained safecontent redaction
* Extend gopass env with --stdin, --file, and --exec modes

**Refactoring**
* [A-1] Split the monolithic Action god object into focused handler types
* [A-2] Replace the stringly-typed context-key config system with typed structs
* Add an architectural plan to split the storage interface

**Documentation**
* Fix all documentation vs. implementation mismatches found in the code quality report
* Document pending security work


AnomalRoil:
Ah wow, big one! If you give me a week I should be able to do a proper code
review. 




-------- Original Message --------

dominikschulz:
@AnomalRoil Yeah, sorry for the amount of changes. Each commit should be self contained but I didn't both to break them up into separate PRs. I'll still need to fix a few failing tests. Thanks!

dominikschulz:
The remaining test failures should be safe to ignore for the review. I'll try to fix them tonight, but it's likely a minor issue only.

The CodeQL issues might be something I need to look at more closely. At first sight it does look like an actual issue.

dominikschulz:
Oh, and there are a number of intentional flag changes which could be considered breaking for some users. So once this is released it should be a minor (Y, in X.Y.Z) version bump with a warning notice in the changelog and release announcement. But I do think those flag changes are justified.

dominikschulz:
Ok, the CodeQL issues seem to be false positives and I can't reproduce the MacOS failures locally.

docs/commands/generate.md
```diff
@@ -39,10 +41,19 @@ Use `--generator` to select one of the available password generators:
...
 |-------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `cryptic`   | The default generator yields cryptic passwords that should work with most sites. Use `--symbols` and `--strict` if the site has specific requirements. Please note that we auto-detect the correct rules for some sites. The length argument specifies the number of characters. |
-| `xkcd`      | Use an [XKCD#936](https://xkcd.com/936/) style password. Use `--lang` and `--sep` to refine it's behaviour. The length argument specifies the number of words.                                                                                                                   |
+| `--xkcd`      | Use an [XKCD#936](https://xkcd.com/936/) style password. Use `--xkcd-lang` and `--xkcd-sep` to refine its behaviour. The length argument specifies the number of words.                                                                                                                   |
```
AnomalRoil, line 44: This looks wrong, the generator is xkcd and is called using `gopass generate -g xkcd entry`


```suggestion
| `xkcd`      | Use an [XKCD#936](https://xkcd.com/936/) style password. Use `--xkcd-lang` and `--xkcd-sep` to refine its behaviour. The length argument specifies the number of words.                                                                                                                   |
```

docs/exit-codes.md
```diff
@@ -0,0 +1,109 @@
...
+| 14 | `Audit` | Audit found one or more issues |
+| 15 | `Fsck` | Integrity check found errors |
+| 16 | `Config` | Configuration error (reserved, not yet emitted) |
+| 17 | `Recipients` | Recipient operation failed |
+| 18 | `IO` | Miscellaneous I/O error |
+| 19 | `GPG` | Miscellaneous GPG error (reserved, not yet emitted) |
+| 20 | `Hook` | Hook execution failed |
+| 21 | `Doctor` | Doctor found one or more failing checks |
+
+## Per-Command Summary
```
AnomalRoil, line 39: Seems this is missing quite a few of the ones listed above, mostly the ones related to initialization or setup. I think `gopass setup` and `gopass init` probably return some of these.

docs/security.md
```diff
@@ -64,3 +64,76 @@ secrets by checking out old revisions from the repository.
 
 Please note that we try to make it hard to lock yourself out from your secrets.
 To ensure that a user is always able to decrypt his own secrets, gopass requires that the user has at least one private key available (that matches the current public keys on file for the password store).
+
+---
+
+## Security Engineering Highlights
```
AnomalRoil, line 70: This feels like it's missing all the things related to recipient hash checks to prevent malicious key additions.

internal/action/pwgen/commands.go
```diff
@@ -45,25 +45,25 @@ func GetCommands() []*cli.Command {
...
+					Name:    "xkcd-lang",
+					Aliases: []string{"lang", "xkcdlang"},
 					Usage:   "Language to generate password from, currently only en (english, default) or de are supported",
 					Value:   "en",
 				},
 				&cli.BoolFlag{
-					Name:    "xkcdcapitalize",
-					Aliases: []string{"xc"},
+					Name:    "xkcd-capitalize",
+					Aliases: []string{"xkcdcapitalize", "xc"},
```
AnomalRoil, line 61: Why keep `xc` here while removing `xl` and `xs` above? The short ones seem practical to me, I'm more likely to use the short ones than the long one if I don't have shell auto-completion on.

internal/action/commands.go
```diff
@@ -38,7 +38,7 @@ func ShowFlags() []cli.Flag {
 		},
 		&cli.BoolFlag{
 			Name:    "unsafe",
-			Aliases: []string{"u", "force", "f"},
+			Aliases: []string{"u"},
```
AnomalRoil, line 41: Wow, that's a big breaking change for users, to me `-f` or `-force` is more intuitive to bypass a restriction (and it's what the text of the unsafe content mode used to tell the user to do when it was removing unsafe content and causing it to be an empty value, so users are definitively used to use `-f`, so we might need to re-instate the warning when safe content suppresses the output, I guess...)

Image

(I have to use `-C=false` since I have`show.autoclip = true` and the warning only triggers when gopass doesn't do anything with the password)

internal/action/doctor.go
```diff
@@ -0,0 +1,213 @@
...
+	"github.com/gopasspw/gopass/internal/backend/storage/gitfs"
+	"github.com/gopasspw/gopass/internal/out"
+	"github.com/gopasspw/gopass/pkg/ctxutil"
+	"github.com/gopasspw/gopass/pkg/fsutil"
+	"github.com/urfave/cli/v2"
+)
+
+// Doctor checks the gopass installation for common issues and prints a
+// diagnostic report. It exits with a non-zero status if any check fails.
+func (s *miscHandler) Doctor(c *cli.Context) error {
```
AnomalRoil, line 19: That's a cool idea. Completion installation might be another thing it could check and warn about, so users are informed they have a suboptimal experience if they haven't installer the autocompletion files?

internal/action/find.go
```diff
@@ -61,6 +61,22 @@ func (s *Action) find(ctx context.Context, c *cli.Context, needle string, cb sho
...
+		// try fuzzy match.
+		cm := closestmatch.New(haystack, []int{2})
+		choices = cm.ClosestN(needle, 5)
+	}
+
+	// JSON output: emit the matches as a JSON array (possibly empty) and return.
+	if c != nil && c.Bool("json") {
+		if len(choices) < 1 {
+			return exit.Error(exit.NotFound, nil, "no results found")
+		}
```
AnomalRoil, line 75: This `len` check is making the about comment about json output lie: it's not possible to have an empty json array right now. Might deserve a test and a design decision, but I'd be in favor of printing an empty json array when find doesn't found anything, while still returning the right error code. Not sure which is most practical to use in practice tho.

internal/action/grep.go
```diff
@@ -44,13 +44,15 @@ func (s *Action) Grep(c *cli.Context) error {
 	for _, v := range haystack {
 		sec, err := s.Store.Get(ctx, v)
 		if err != nil {
-			out.Errorf(ctx, "failed to decrypt %s: %v", v, err)
+			out.Errorf(ctx, "Failed to decrypt %s: %v", v, err)
+			errors++
```
AnomalRoil, line 48: This feels like something we should have caught in tests ^^'' 
Good find!

internal/action/setup.go
```diff
@@ -166,13 +166,10 @@ func (s *Action) initGenerateIdentity(ctx context.Context, crypto backend.Crypto
...
-	// support fully automated setup (e.g. for tests)
+	// support fully automated setup (e.g. for tests or CI via GOPASS_AGE_PASSWORD)
 	//nolint:nestif
-	if ctxutil.HasPasswordCallback(ctx) {
-		pw, err := ctxutil.GetPasswordCallback(ctx)("", true)
-		if err == nil {
-			passphrase = string(pw)
-		}
+	if ap := ctxutil.GetAgePassphrase(ctx); ap != "" {
+		passphrase = ap
```
AnomalRoil, line 172: This feels like a slight design change: we used to allow empty password in this codepath, but we don't anymore now. Although I'm not 100% an empty password if full supported everywhere for age.

internal/action/show.go
```diff
@@ -331,12 +331,19 @@ func showSafeContent(sec gopass.Secret) string {
...
 }
 
-func isUnsafeKey(key string, sec gopass.Secret) bool {
+func isUnsafeKey(ctx context.Context, key string, sec gopass.Secret) bool {
 	duks := []string{"hotp", "otpauth", "password", "totp"}
 	if slices.Contains(duks, key) {
 		return true
 	}
 
+	// Check globally-configured hidden keys (show.hidden-keys, repeatable).
```
AnomalRoil, line 340: Cool addition, thanks!

internal/backend/crypto/age/ssh.go
```diff
@@ -7,25 +7,39 @@ import (
...
+		debug.Log("using sshCache")
+
+		return sshCache, nil
+	}
+	sshCacheMu.RUnlock()
+
+	sshCacheMu.Lock()
+	defer sshCacheMu.Unlock()
+	// Re-check after acquiring the write lock (another goroutine may have
+	// populated the cache between the RUnlock and Lock above).
```
AnomalRoil, line 42: This feels like it could have just been a `Lock` in the first place

```suggestion
	sshCacheMu.Lock()
	defer sshCacheMu.Unlock()
	// Re-check after acquiring the write lock (another goroutine may have
	// populated the cache between the RUnlock and Lock above).
```

I don't feel like this is a super hot path worth the extra code complexity overhead

internal/create/wizard.go
```diff
@@ -291,7 +291,7 @@ func mkActFunc(tpl Template, s *root.Store, cb ActionCallback) func(context.Cont
 					if v.Min > 0 && len(password) < v.Min {
 						return fmt.Errorf("%s is too short (needs %d)", v.Name, v.Min)
 					}
-					if v.Max > 0 && len(password) > v.Min {
+					if v.Max > 0 && len(password) > v.Max {
```
AnomalRoil, line 294: Wow, the wizard wasn't working at all then xD

internal/queue/background.go
AnomalRoil, line 85: This loop feel like it should check whether `ctx.Err() != nil` and probably stop execution if so, no?

internal/queue/background.go
```diff
@@ -94,52 +92,56 @@ func (q *Queue) run(ctx context.Context) {
...
+		close(done)
 	}()
-
 	select {
 	case <-done:
 		return nil
 	case <-time.After(maxWait):
-		return fmt.Errorf("timed out waiting for empty queue")
+		return fmt.Errorf("timed out waiting for queue to drain")
 	}
```
AnomalRoil, line 135: This looks like it could easily cause some go routine leaks of the one just above that's running `q.wg.Wait()`, because tasks added to the queue are causing the `wg` to grow, but when the `Queue` is `Close`d, the `wg` isn't drained since `q.work` is just closed.

AnomalRoil (commented):
Sorry I couldn't do this earlier, here are some comments, feel free to cherry pick them or discard them

dominikschulz:
@AnomalRoil Thanks a lot. Sorry, I didn't exactly know how long to wait and thought ~one week to be reasonable. I will review your feedback!

internal/queue/background.go
dominikschulz: My idea here was to allow certain tasks to be procssed in the background (e.g. git push) but still block on shutdown until they are done.

PR #3388, opened 2026-04-19
fix: recover from panic during shell completion with unknown flags

guillermodotn:
Work around a nil `FlagSet` dereference in [`urfave/cli`](https://github.com/urfave/cli) that crashes gopass during tab-completion with unknown flags.

e.g.
```
gopass --something<TAB>
```

The `recover()` only activates during shell completion, normal usage is unaffected. Will submit a follow-up to remove this once the upstream gets fixed.

### DIFF
--- a/main.go
+++ b/main.go
@@ -89,7 +89,7 @@ func main() {
 	ctx = queue.WithQueue(ctx, q)
 	ctx, app := setupApp(ctx, sv)
 
-	if err := app.RunContext(ctx, os.Args); err != nil {
+	if err := runApp(ctx, app); err != nil {
 		log.Fatal(err)
 	}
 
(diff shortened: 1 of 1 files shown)


dominikschulz (commented):
Do you have an upstream issue we could attach as a reference here?

guillermodotn:
I did some more research and this is fixed in v3 of [`urfave/cli`](https://github.com/urfave/cli). Since v2.27.7 is
the last v2 release don't think we will see this fix soon.

Related closed issues: [#1822](https://github.com/urfave/cli/issues/1822)

Is there any plan to move to v3 in the future?

dominikschulz:
@guillermodotn Thanks. I haven't looked into migrating to v3, but I think we should do it if there are no blockers.

dominikschulz (approved):

PR #3389, opened 2026-04-20
[fix] Support HW Age identities

### LINKED ISSUE #3387: Unable to encrypt to hardware-backed age recipients (p256tag / age1tag1...) because filippo.io/age/tag is not imported
### Summary
When attempting to encrypt secrets to a hardware-backed `age` recipient (like those generated by `age-plugin-tpm` v1.0.0+ or `age-plugin-se`), gopass fails with an unsupported plugin error. 
The cryptography community and `age` standardized hardware tokens (TPM, YubiKey, Secure Enclave) to use a unified `p256tag` recipient format that starts with `age1tag1...` instead of device-specific prefixes (like `age1tpm1...`). 
Native support for this was added to the core `age` library in `v1.3.0` via the `filippo.io/age/tag` package. While `gopass` has updated its `go.mod` to use `filippo.io/age v1.3.1`, the `filippo.io/age/tag` package is never imported in the gopass age crypto backend. Because the native parser isn't registered, `gopass` incorrectly assumes `age1tag1...` is a 3rd-party plugin, looks for an executable named `age-plugin-tag`, and crashes.
### Steps To Reproduce
1. Install the latest version of `age-plugin-tpm` (v1.0.1) or `age-plugin-se`.
2. Generate a new identity: `age-plugin-tpm --generate -o tpm-identity.txt`
3. Convert it to a recipient: `age-plugin-tpm -y tpm-identity.txt > tpm-recipient.txt` (This will output a string starting with `age1tag1...`)
4. Add the recipient to a gopass age store: `gopass recipients add --store my-store $(cat tpm-recipient.txt)`
5. Attempt to insert a secret: `gopass insert my-store/test`
6. The encryption fails because `age` tries to invoke a non-existent plugin named `tag`. 
   *(Note: Symlinking `age-plugin-tag` to `age-plugin-tpm` results in `unsupported plugin name: "tag"` because the plugin explicitly expects to be called `tpm`).*
### Expected behavior
`gopass` should natively recognize `age1tag1...` and `age1tagpq1...` recipients and delegate to the appropriate plugin under the hood, exactly like the standard `age` CLI tool does in v1.3.0+. 
This can likely be fixed by importing `filippo.io/age/tag` in the `internal/backend/crypto/age` package initialization.
### Environment
- OS: Debian 
- OS version: 13 (trixie)
- gopass Version: 1.16.1
### Additional context
- **Standardization Context:** [C2SP/C2SP#96](https://github.com/C2SP/C2SP/pull/96)
- **Age Release Notes (v1.3.0):** [Built-in recipients compatible with hardware plugins](https://github.com/FiloSottile/age/releases/tag/v1.3.0)
- **Workaround:** Users are currently forced to pass a legacy flag (`--tpm-recipient` for `age-plugin-tpm`) to their plugins to generate deprecated strings (e.g. `age1tpm1...`)

dominikschulz:
This adds support for age1tag prefixed identities by using the filippo.io/age/tag package for recipient parsing.

Fixes #3387


.capabilities.json
AnomalRoil, line 1: Pretty sure you didn't mean to commit this, right?

dominikschulz: I'm not 100% sure what to think of capslock, yet. I do like the idea but it can noisy when we need to pull in updates of dependencies or the toolchain. In this case I had to update the file. But I can double check if that's really necessary.

AnomalRoil: It's breaking Github code review for me because of the size of the change afaict.

dominikschulz: @AnomalRoil Done.

AnomalRoil (approved):
lgtm

PR #3410, opened 2026-05-03
[chore] Follow up to PR 3383

dominikschulz:
This should address most of the post-merge comments.

### DIFF
--- a/docs/security.md
+++ b/docs/security.md
@@ -131,6 +131,43 @@ checks for expiration, minimum trust level, and the presence of an encryption
 sub-capability. Expired or untrusted keys are rejected, preventing silent
 encryption to keys that can no longer decrypt.
 
+### Recipient File Integrity (Hash Pinning)
+
+The recipients file (`.gpg-id`) lives inside the git repository and lists the
+public-key fingerprints/IDs that secrets are encrypted to. Because the file is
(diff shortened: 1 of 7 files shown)


docs/security.md
```diff
@@ -131,6 +131,43 @@ checks for expiration, minimum trust level, and the presence of an encryption
...
+public-key fingerprints/IDs that secrets are encrypted to. Because the file is
+git-tracked, a collaborator — or an attacker with write access to the remote
+repository — could push a commit that silently adds their own key, causing
+future encrypt operations to include them as a recipient.
+
+gopass defends against this with a hash-pinning mechanism:
+
+1. **Hash computation.** Every time gopass writes the recipients file (via
+   `gopass recipients add`, `gopass recipients rm`, or `gopass init`), it
+   computes the SHA-256 digest of the serialised recipients list.
```
AnomalRoil, line 146: This isn't true, that's only done when running `gopass recipients ack`, no ?

AnomalRoil (approved):
One last nit, I guess

docs/security.md
dominikschulz: No, this is correct for all three code paths. I just checked to make sure. `gopass recipients ack` is only necessary if someone else on a shared repo changes the recipients so the local (!) checksum is updated.

PR #3420, opened 2026-05-03
fix(secparse): replace panic in MustParse with testing.TB.Fatalf

dimension-zero:
## Summary

`secparse.MustParse` was exported from a public package (`pkg/gopass/secrets/secparse`) and contained a `panic(err)` that fires on a `*secrets.PermanentError` from malformed MIME input. Although the function was documented as test-only, the panic was reachable from any caller — including non-test code that imports the package.

- Replace `panic(err)` with `tb.Fatalf(...)`, using a `testing.TB` parameter
- Add `tb.Helper()` so test failure lines point to the call site, not inside `MustParse`
- Update the three call sites in `pkg/otp/otp_test.go` to pass `t`

The `testing.TB` parameter type-enforces test-only use: production code cannot call this function without a `testing.TB` value, making the intent structurally clear rather than just a comment.

## Test plan

- [x] `go build ./pkg/gopass/secrets/secparse/...` — clean
- [x] `go test ./pkg/otp/...` — passes
- [x] `go test ./pkg/gopass/secrets/...` — passes

### DIFF
--- a/pkg/gopass/secrets/secparse/parse.go
+++ b/pkg/gopass/secrets/secparse/parse.go
@@ -4,6 +4,7 @@ package secparse
 
 import (
 	"errors"
+	"testing"
 
 	"github.com/gopasspw/gopass/internal/out"
 	"github.com/gopasspw/gopass/pkg/debug"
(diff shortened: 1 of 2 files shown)


dominikschulz (approved):
Nice!

PR #3421, opened 2026-05-03
docs(appdir): explain GOPASS_HOMEDIR override in UserHome comment

dimension-zero:
## Summary

`pkg/appdir/appdir.go` — `UserHome()` silently overrides the OS home directory when `GOPASS_HOMEDIR` is set. The existing one-line comment said only "It respects the GOPASS_HOMEDIR environment variable," giving a security auditor no context for why a password manager binary honours an environment variable that redirects all data paths.

The expanded comment explains:
- The variable exists for hermetic test and CI environments
- It should not be set in normal production use
- Setting it silently redirects every config, cache, and data path

No behaviour change.

## Test plan

- [x] `go build ./pkg/appdir/...` — clean
- [x] `go vet ./pkg/appdir/...` — clean

### DIFF
--- a/pkg/appdir/appdir.go
+++ b/pkg/appdir/appdir.go
@@ -50,8 +50,13 @@ func UserData() string {
 	return DefaultAppdir.UserData()
 }
 
-// UserHome returns the user's home dir.
-// It respects the GOPASS_HOMEDIR environment variable.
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
Thoughtful comment, thanks!

Issue #3417, opened 2026-05-03
refactor(ctxutil): freeze and gradually reduce context-threading in pkg/ctxutil


dimension-zero:
## Background

`pkg/ctxutil/` contains 30+ context keys used to pass configuration values, terminal state, credentials, and backend selection through the call stack. `ARCHITECTURE.md` acknowledges this as "pragmatic (read: non-idiomatic)."

The pattern has two practical problems:

1. **Silent failures**: context keys are looked up by unexported types; a wrong key type returns the zero value at runtime rather than a compile error.
2. **Scope creep**: because adding a new context key is frictionless, the pattern expands over time. Credentials in context create a risk of accidental logging if a context is ever passed to a logging call.

## Proposed approach

This does not require a full rewrite. The goal is to stop the pattern from growing and reduce it incrementally.

### Phase 1 — Freeze (low effort)
Add a comment to the `pkg/ctxutil/` package doc: "No new context keys should be added. Use explicit parameters or a config struct instead." This creates a policy without touching any existing code.

### Phase 2 — Extract high-frequency keys (medium effort)
Identify the most-accessed keys (likely: `TerminalWidth`, `Verbose`, `AutoPrint`, `Color`) and extract them into a small `DisplayConfig` struct passed explicitly through `Action`. This removes them from context without rewriting the full stack.

### Phase 3 — Credentials audit (targeted)
Review which ctxutil keys carry credential-adjacent data and confirm none reach log or debug call sites.

## Question for maintainers

Is a freeze policy on new ctxutil keys acceptable as a first step? That is a one-line doc change that prevents the problem from growing while a longer-term reduction plan is discussed.

dominikschulz:
I already have a cleanup for this pending, will try to get this in shape and merged soon. See #3423

PR #3423, opened 2026-05-03
docs(adr): add A-11 — context-threading freeze and reduction plan

dimension-zero:
## Summary

`pkg/ctxutil/` has grown to 30+ context keys threading config, terminal state, and credentials through the call stack. `ARCHITECTURE.md` acknowledges this as non-idiomatic. No architectural decision records what the project intends to do about it, so the pattern continues to expand without a policy boundary.

This PR adds `docs/adr/A-11-context-threading.md` which:

- Documents the three risks: silent failures on wrong key type, credentials in context, and unchecked scope creep
- Recommends a freeze (no new context keys after this ADR) as a zero-effort first step
- Proposes a phased reduction plan (freeze → extract display-config struct → credential audit) that can be worked incrementally without a flag day

No code change. The ADR is a decision record, not an implementation.

Related issue: #3417

## Test plan

- [x] Markdown renders correctly
- [x] ADR number A-11 does not conflict with existing `docs/adr/` files (highest existing is A-10)

### DIFF
--- a/docs/adr/A-11-context-threading.md
+++ b/docs/adr/A-11-context-threading.md
@@ -1,0 +1,79 @@
+# A-11: Context-Threading in pkg/ctxutil
+
+**Status:** active — pattern is in use; reduction is planned but not scheduled  
+**Source:** Architectural Review 2026-05-03 (filed as issue #3417)
+
+---
+
+## Background
+
+`pkg/ctxutil/` defines more than 30 unexported context-key types used to pass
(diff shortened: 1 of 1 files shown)


dominikschulz:
I think I have a pending branch somewhere that removes most of the context (ab)use already. I simply need to get it into shape and get it merged. I would leave this PR open as a reminder to finally do that, but I'm most likely not going to merge it.

Issue #3416, opened 2026-05-03
refactor(debug): replace pkg/debug printf-style logging with log/slog


dimension-zero:
## Background

`pkg/debug/` implements a conditional debug logger using printf-style calls (`debug.Log(...)`, `debug.V(1).Log(...)`). This produces unstructured text output, which is difficult to parse in CI environments or when correlating log lines across operations.

`log/slog` has been part of the Go standard library since Go 1.21. The project already requires Go 1.25 (`go.mod`), so `slog` is available with no new dependency.

## Proposed change

Replace `pkg/debug`'s internal implementation with `log/slog` as the backing logger:

- `debug.Log(format, args...)` → `slog.Debug(...)` (structured key-value pairs where call sites permit, otherwise `fmt.Sprintf` for the message)
- `debug.V(n).Log(...)` → `slog.Debug(...)` with a level filter
- Add a `--log-format json` flag (or `GOPASS_LOG_FORMAT=json` config key) to switch the slog handler to `slog.NewJSONHandler`

## Benefits

- Machine-readable debug output for CI log parsing and structured observability
- No new dependency (stdlib only)
- Existing `--debug` / `GOPASS_DEBUG` activation mechanism unchanged

## Scope

`pkg/debug/` is the implementation package; call sites are spread across `internal/`. A mechanical replacement of the internal implementation is sufficient for the base migration. Structured key-value improvements at individual call sites can follow incrementally.

## Question for maintainers

Is a `log/slog`-backed `pkg/debug` welcome? Happy to scope a PR once there is agreement on the public interface shape.

PR #3424, opened 2026-05-03
docs(adr): add A-12 — structured debug logging via log/slog

dimension-zero:
## Summary

`pkg/debug/` uses printf-style debug output that is unstructured and hard to parse in CI or log aggregators. `log/slog` is available in the stdlib (since Go 1.21; the project already requires Go 1.25), so there is zero dependency cost to migrating.

This PR adds `docs/adr/A-12-structured-logging.md` which:

- Documents the current state and the observability limitation
- Presents three options: full `slog` migration, no change, or JSON-flag-only
- Recommends Option A (migrate `pkg/debug` internally to `slog`, keeping the public `Log`/`V`/`LogN` API unchanged) with a `--log-format json` flag in the same PR
- Notes that structured key-value improvements at individual call sites across `internal/` can follow incrementally

No code change. This ADR is a proposed decision record pending maintainer input.

Related issue: #3416

## Test plan

- [x] Markdown renders correctly
- [x] ADR number A-12 does not conflict with existing `docs/adr/` files

### DIFF
--- a/docs/adr/A-12-structured-logging.md
+++ b/docs/adr/A-12-structured-logging.md
@@ -1,0 +1,83 @@
+# A-12: Structured Debug Logging via log/slog
+
+**Status:** proposed — decision pending maintainer input  
+**Source:** Architectural Review 2026-05-03 (filed as issue #3416)
+
+---
+
+## Background
+
+`pkg/debug/` implements conditional debug output using printf-style calls:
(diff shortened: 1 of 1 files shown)


dominikschulz:
I am open to improvements of the log output, but currently it's primary focus are humans (or maybe Agents) debugging gopass. So structured logging is not a priority. I would like to keep the default format. But adding a separate - optional - output format would be fine with me.

PR #3442, opened 2026-05-25
otp: add noscreenshot build tag and document screen-capture scope

### LINKED ISSUE #3415: audit(passkey): document kbinani/screenshot usage and evaluate opt-out build tag
## Background

`go.mod` includes `github.com/kbinani/screenshot` as a direct dependency. This package captures the contents of display screens — an unusual capability for a password manager binary.

The dependency is used in `pkg/passkey/` for WebAuthn/passkey QR code capture. The passkey feature is relatively new and the screenshot surface area has not been formally documented.

## Concerns

1. **Unexpected capability**: Users auditing the binary permissions or capabilities do not expect a password manager to have screen capture functionality. This may trigger security reviews or policy blocks in enterprise environments.
2. **Attack surface**: A bug in `kbinani/screenshot` or its platform-specific backends could expose screen contents.
3. **No opt-out**: Users who do not use the passkey feature still link against the screenshot library.

## Proposed actions

1. **Audit**: Document exactly which `pkg/passkey/` function calls `screenshot`, under what user-visible conditions, and whether it can be triggered without explicit user action.
2. **Document**: Add a section to `docs/backends/` (or a new `docs/passkey.md`) explaining the screenshot dependency, its purpose, and its scope.
3. **Evaluate build tag**: Assess whether a `//go:build passkeys` tag on `pkg/passkey/` would let users opt out of this surface at compile time without losing other functionality.

A companion ADR issue has been filed for the build tag decision.

dominikschulz:
The kbinani/screenshot dependency is pulled in solely by pkg/otp/ParseScreen, which is invoked only when the user explicitly passes --snip (-s) to gopass otp. It is never triggered automatically.

- Add the 'noscreenshot' build tag: building with -tags noscreenshot omits the screenshot_supported.go file so the library is not linked and the --snip flag falls back to the unsupported-platform stub.
- Document the dependency, the triggering conditions, and the build tag in docs/commands/otp.md.
- Add ADR A-13 recording the audit findings and the decision rationale.

Closes #3415

### DIFF
--- a/docs/adr/A-13-screenshot-build-tag.md
+++ b/docs/adr/A-13-screenshot-build-tag.md
@@ -1,0 +1,84 @@
+# A-13: `noscreenshot` Build Tag for OTP Screen-Capture Feature
+
+**Status:** accepted  
+**Source:** [GitHub Issue #3415](https://github.com/gopasspw/gopass/issues/3415)
+
+---
+
+## Background
+
+`pkg/otp/screenshot_supported.go` imports `github.com/kbinani/screenshot` to capture display
+contents when the user runs `gopass otp --snip`.  This allows gopass to locate an OTP QR code
+that is visible on screen and store the decoded `otpauth://` URL directly into a secret.
+
+`screenshot` is a notable dependency for a password manager because it grants the binary the
+ability to read the full screen.  Users auditing binary capabilities or operating in
+policy-restricted environments (e.g., enterprise security reviews, MDM policies) may need to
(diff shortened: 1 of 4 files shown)


AnomalRoil (commented):
LGTM. I wonder if we should add an `otp.Flags` function that would return the right flags, so that `--snip` isn't even registered as a possible flag when compiled without the tag

dominikschulz:
@AnomalRoil #3445 PTAL

PR #3445, opened 2026-05-25
otp: hide --snip flag when built with noscreenshot tag

dominikschulz:
Add otp.SnipFlags() to the build-tagged screenshot files so that the --snip/-s CLI flag is only registered when the screen-capture feature is actually compiled in. Builds with -tags noscreenshot will no longer show --snip in 'gopass otp --help'.

Addresses review comment by AnomalRoil on PR #3442.


AnomalRoil (approved):
LGTM

PR #3448, opened 2026-05-30
Fix clip flag (-c) for cli/v3 GenericFlag compatibility

### LINKED ISSUE #3444: Copy flag seems broken at head
With the current dev build `show -c` seems to be broken, the content is displayed and not copied. Likely a cli v3 regression. Need to check that before the next release.

guillermodotn:
After the cli/v2 to v3 migration, `cmd.String()` on a `GenericFlag` always returns `""` because it type-asserts to `string`, which fails for `*OptionalInt`. This caused `show -c=N` (copy a specific line to clipboard) to silently ignore the line number, always copying the password instead.

Fixes #3444

### DIFF
--- a/internal/action/show.go
+++ b/internal/action/show.go
@@ -24,6 +24,24 @@ import (
 	"github.com/urfave/cli/v3"
 )
 
+// clipFlagValue returns the string value of the "clip" GenericFlag.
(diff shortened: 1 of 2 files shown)


dominikschulz (approved):
Thanks!

PR #3461, opened 2026-06-03
clone: match GPG recipients by fingerprint, not string equality

### LINKED ISSUE #3028: Incomplete key comparison in clone
### Summary
On `gopass clone` with GPG, gopass checks whether we already have access to the cloned repository by comparing our usable keys to the keys listed in .gpg-id. However, this is a string comparison. While `.gpg-id` allows for any GPG-approved way to specify a key, gopass will use short form keys (e.g. `0x17F3ED51DADD9393`) to compare.

### Steps To Reproduce
1. Have a repository with one of your keys in a non-short form in `.gpg-id`
2. `gopass clone` the repository
3. gopass will reply with "Please ask the owner of the password store to add one of your keys"

### Expected behavior
In step 3: gopass will reply with "Found valid decryption keys. You can now decrypt your passwords."

### Environment
- OS: Arch Linux
- OS version: Linux vischium 6.11.5-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 22 Oct 2024 18:31:38 +0000 x86_64 GNU/Linux
- gopass Version: gopass 1.15.15 go1.23.3 linux amd64
- Installation method: pacman

arpitjain099:
## What

`gopass clone` with the GPG backend decides whether the cloning user already has access to the store by comparing the user's usable keys against the recipients listed in `.gpg-id`. That comparison was a plain string equality: usable keys are reported in short form (for example `0x17F3ED51DADD9393`), while `.gpg-id` accepts any GPG-approved key form (short ID, long ID, full fingerprint, or email).

As a result, a recipient written in any form other than the short ID never matched a usable key, and a perfectly valid recipient was told:

> Please ask the owner of the password store to add one of your keys

instead of:

> Found valid decryption keys. You can now decrypt your passwords.

## Fix

In `cloneCheckDecryptionKeys` the match now normalizes both the recipients and the usable identities to their fingerprint before comparing, using the crypto backend's existing `Fingerprint` resolver (`crypto.Fingerprint`, which resolves any key form via the backend's `findKey`). Short ID, long ID, fingerprint, and email forms of the same key all resolve to the same fingerprint and therefore match.

The check is extracted into a small pure helper, `haveDecryptionKey`, that takes the fingerprint resolver as a parameter so it can be tested without a live GPG keyring. The access check is not weakened: a key that is not actually a recipient still does not match, so a non-recipient is still correctly asked to have their key added. Keys the backend cannot resolve (empty fingerprint) are ignored rather than matched.

## Tests

Added `TestHaveDecryptionKey` covering a usable short-form key against a `.gpg-id` recipient written as short ID, long ID, fingerprint, `0x`-prefixed fingerprint, and email (all match), plus negative cases: a non-recipient key, no recipients, an unknown recipient, and no identities (all correctly rejected). `go test ./internal/action/...`, `gofmt`, and `go vet` are clean.

Fixes #3028

### DIFF
--- a/internal/action/clone.go
+++ b/internal/action/clone.go
@@ -214,8 +214,12 @@ func (s *setupHandler) cloneCheckDecryptionKeys(ctx context.Context, mount strin
 	}
 
 	idSet := set.New(ids...)
-	// Check whether any of our usable keys are in recpSet
-	if _, found := recpSet.Choose(idSet.Contains); found {
+	// Check whether any of our usable keys is listed as a recipient. We can not
+	// rely on plain string equality here: .gpg-id may list a recipient in any
+	// GPG-accepted form (short ID, long ID, full fingerprint or email), while
+	// ListIdentities returns our keys in short form. Normalize both sides to a
+	// fingerprint before comparing so every form of the same key matches.
+	if haveDecryptionKey(ctx, crypto.Fingerprint, recpSet.Elements(), idSet.Elements()) {
 		out.Noticef(ctx, "Found valid decryption keys. You can now decrypt your passwords.")
 
 		return nil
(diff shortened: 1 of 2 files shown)


dominikschulz:
Thanks for the PR. I'm afraid this might overlap with some other pending changes on mine that just have been merged. Please rebase this onto master to see if there is any meaningful diff left.

arpitjain099:
Rebased onto master as suggested. This change targeted `cloneCheckDecryptionKeys` in internal/action/clone.go, and the recent team-workflow refactor removed that function (clone.go now has `cloneJoinTeam` and the decryption-key check no longer lives there), so there is no clean target left to rebase onto - effectively superseded. I did not see fingerprint normalization in the new clone/setup paths, so if the original concern still applies (matching .gpg-id recipients written as long-ID/fingerprint/email against short-form usable keys in the new join flow), I am happy to re-target the fix there. Otherwise this can be closed - whichever you prefer.

arpitjain099:
Closing this - master has since replaced the `cloneCheckDecryptionKeys` path (which this PR modified) with the unified `cloneJoinTeam` / `Store.JoinTeam` flow, so the change no longer applies. If the recipient-vs-identity matching in the new JoinTeam path would still benefit from fingerprint normalization, that'd be a fresh change against that code. Thanks!

PR #3462, opened 2026-06-04
fscopy: derive copy direction from the source argument

arpitjain099:
Fixes #3340

`gopass fscopy <file> <dest>` refused to run whenever the destination
secret name happened to match a file of the same name in the current
directory. The reproduction from the issue:

```
$ touch test
$ gopass fscopy test test
Error: ambiguity detected. Only from or to can be a file. Use cp to copy between files
```

The disambiguation in `binaryCopy` decided the direction by inspecting
both arguments at once. Because `isFilePath` returns true for any real
file in the working directory, a perfectly valid filesystem-to-store
copy where the destination secret shares a basename with the source file
fell into the `isFilePath(from) && isFilePath(to)` branch and was
rejected.

This follows the direction agreed in the issue thread (if `from` is a
real file, treat `to` as a store entry, since fscopy is not meant to copy
between two files): the copy direction is now derived from the source
alone.

- If `from` is a file on disk, copy into the store and treat `to` as a
  secret name, even if a same-named file exists in the working directory.
- Otherwise, if `from` is an existing secret, copy out to a file.

The genuine guards are preserved:

- Both arguments are existing secrets still returns the "use `gopass cp`"
  error.
- A source that already lives inside the store still warns and requires
  `--force`.
- A source that is neither a file nor a secret now returns a clear
  message instead of the previous "unhandled case, please report a bug".

Before / after with the fixed binary:

```
$ gopass fscopy test test          # was: ambiguity detected ...
$ gopass ls
gopass
└── test
$ gopass fscopy test sub/test      # nested destination, also works now
```

Tests: added `TestBinaryCopyNameAmbiguity` covering the same-name and
nested-name copies succeeding, the secret-to-secret case still erroring,
and an unknown source erroring. `go test ./internal/action/...` passes,
`gofmt` and `go vet` are clean. The new test fails against the previous
code with the exact error string from the issue, and passes with this
change.

### DIFF
--- a/internal/action/binary.go
+++ b/internal/action/binary.go
@@ -194,14 +194,17 @@ func (s *binaryHandler) binaryCopy(ctx context.Context, cmd *cli.Command, from, 
 		return fmt.Errorf("usage: %s fs%s from to", cmd.Root().Name, op)
 	}
 
+	// The direction is determined by the source: if "from" is a real file on
+	// disk we copy from the filesystem into the store, otherwise we copy from
+	// the store out to the filesystem. We deliberately key off "from" so that a
+	// destination which happens to share a name with a file in the current
+	// directory (e.g. "gopass fscopy test test") is still treated as a store
+	// entry. fscopy does not support copying between two files, so once the
+	// source is known the destination side is unambiguous. See #3340.
 	switch {
-	case isFilePath(from) && isFilePath(to):
-		// copying from on file to another file is not supported.
-		return fmt.Errorf("ambiguity detected. Only from or to can be a file. Use cp to copy between files")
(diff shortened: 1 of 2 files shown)


dominikschulz:
Thanks, the reasoning sounds good. Please rebase onto master and look at the lint issues. Feel free to ignore the Grype Scan action. That builder seems to be broken right now and should be disabled on master for the time being.

arpitjain099:
Rebased onto master. golangci-lint v2.11.4 (the pinned version) now reports 0 new issues on the branch, and gofmt, go vet, and the internal/action tests all pass locally. Ready for another look, thanks!

dominikschulz (approved):

PR #3474, opened 2026-06-30
fix(age): expand ~/ in age.ssh-key-path

yzx9:
`age.ssh-key-path` (and the `--age-ssh-key-path` flag) was read raw and passed to `fsutil.IsDir/IsFile`, which do not expand a leading `~/`. A custom path like `~/custom-keys` was therefore silently ignored; only the default `~/.ssh` discovery ran. Expand it via `fsutil.ExpandHomedir` in New, the single constructor all callers flow through, matching how storage and mount paths already handle `~`.

Adds a unit test and a behavioral regression test; both fail on the un-fixed code.


dominikschulz (dismissed):
Thanks for including the model used.

dominikschulz (approved):

PR #3488, opened 2026-07-06
fix(age): recover agent on missing identities, not only when locked

### LINKED ISSUE #3467: Age agent not caching identities passphrase
Even with the latest stable (`v1.16.1 / 2025-12-13`) or nightly (`v1.17.0-rc.2 / 2026-06-07`) I see two issues that is supposed to work as describe in the backend/age documentation:

- The age agent does not start automatically.
- When the age agent is running, it does not cache the identities file passphrase as you would expect.

If I misconfigured something, please let me know, I would really like to use gopass with age and this feature is mandatory for me!

yzx9:
When the age agent is started ahead of time (e.g. by `launchd`), it is reachable but holds no identities. `decryptWithAgent` only reloaded and sent identities when the agent reported "agent is locked"; a fresh unlocked agent instead returns "no identities specified", which was treated as a hard failure. As a result `Age.Decrypt` silently fell back to direct decryption on every call: the caller loaded the identity for its own fallback use but never forwarded it, and the agent stayed empty and unused for its whole lifetime.

This PR treats "no identities specified" the same as "agent is locked" so the existing `unlock` + `send-identities` + `retry recovery` also covers a freshly started or freshly restarted agent.

Fundamentally, I believe this points to a design issue: the implementation treats the agent as if it only has two relevant states, `locked` and `unlocked`, while in practice there are three: `locked`, `unlocked with identities`, and `unlocked without identities`. The `locked` and `unlocked` states behave more like explicitly declared protocol states than descriptions of the agent’s actual readiness. Being in the unlocked state does not necessarily mean that the agent already holds an identity.

This conceptual mismatch is confusing for both users and developers. For example, when I first used the `gopass age agent unlock` command, I assumed it would force a password prompt, but it did not. In my view, the `unlocked` state should always imply that the agent holds a secret key. That change belongs to a larger redesign, so I have intentionally left it out of this PR.

Closes #3467

### DIFF
--- a/internal/backend/crypto/age/decrypt.go
+++ b/internal/backend/crypto/age/decrypt.go
@@ -41,7 +41,8 @@ func (a *Age) decryptWithAgent(ctx context.Context, ciphertext []byte) ([]byte, 
 		return plaintext, nil
 	}
 
-	if !strings.Contains(err.Error(), "agent is locked") {
+	if !strings.Contains(err.Error(), "agent is locked") &&
+		!strings.Contains(err.Error(), "no identities specified") {
 		debug.Log("failed to decrypt with agent: %s", err)
 
 		return nil, err
--- a/internal/backend/crypto/age/decrypt_agent_test.go
+++ b/internal/backend/crypto/age/decrypt_agent_test.go
@@ -1,0 +1,146 @@
+package age
+
+import (
+	"bytes"
+	"context"
+	"os"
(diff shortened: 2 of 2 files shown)


dominikschulz (approved):
Thanks!

yzx9:
@dominikschulz thank you for quick review!

PR #3492, opened 2026-07-15
fix: use imperative mood in generated commit messages

ivankovnatsky:
I am trying gopass as an enhanced alternative to pass and as a backup vault for my primary password manager. I used pass occasionally for many years and noticed that gopass uses the past tense in some generated commit messages.

Even though it is really a nit, I see the usual pass commit messages alongside gopass's slightly off past-tense ones in the same vault as I adapt to using gopass.

This aligns those messages with the imperative-mood convention accepted in #1222.

Most changes were edited and reviewed with AI assistance.


dominikschulz (approved):
That is a great idea, thanks!

dominikschulz (approved):

PR #3493, opened 2026-07-16
feat(create): support passwordless / SSO login profiles

1995parham:
## Summary

Adds a way to model accounts that have **no stored password** because login happens through an external provider (e.g. Google, Apple, SSO / social login).

The secret model already tolerates an empty password (first line), so this needs **no changes to the `Secret` interface or the storage format** — everything lives in the `gopass create` wizard.

## Changes

- **New default "SSO / passwordless login" template** (`internal/create/templates.go`). It shows up in `gopass create` and records the provider in a `login-method` field instead of prompting for a password:
  ```
  url: https://accounts.google.com
  username: user@example.com
  login-method: google
  ```
- **New `choice` attribute type** (`internal/create/wizard.go`) so a template can offer a fixed list of options. The SSO template uses it for the provider picker (`google`, `apple`, `github`, `microsoft`, `sso`, `magic-link`). Backed by the existing `cui.GetSelection` helper.
- **Optional / skippable password**: a `password` attribute marked `optional: true` asks *"Does this account have a password?"* and skips password entry entirely when declined (instead of storing a blank one). This lets a single template serve both password and passwordless accounts. Existing templates leave the flag unset, so their behaviour is unchanged.
- Tests covering the new default template, the `choice` type, and the `optional` flag.

## Notes

- The SSO template is filed under the existing `websites` prefix (`2-websites.yml`), so there is no collision with the `0-websites.yml` Website login template. Happy to move it under a dedicated prefix (e.g. `sso/`) if preferred.
- `gopass show` already handles empty-password secrets gracefully (it only errors when both password and body are empty, and guards clipboard/QR on a non-empty password), so passwordless entries display and behave correctly.

### DIFF
--- a/internal/create/templates.go
+++ b/internal/create/templates.go
@@ -59,6 +59,37 @@ attributes:
   - name: "comment"
     type: "string"
 `,
+	`---
+priority: 2
+name: "SSO / passwordless login (e.g. Google)"
+prefix: "websites"
+name_from:
+  - "url"
+  - "username"
+welcome: "🧪 Creating SSO / passwordless login"
+attributes:
+  - name: "url"
+    type: "hostname"
+    prompt: "Website URL"
+    min: 1
+    max: 255
+  - name: "username"
+    type: "string"
+    prompt: "Account / email"
+    min: 1
+  - name: "login-method"
+    type: "choice"
+    prompt: "Login via"
+    options:
+      - "google"
+      - "apple"
+      - "github"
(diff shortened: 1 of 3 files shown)


dominikschulz:
Sorry for the delay here. The idea sounds reasonable, I simply didn't get to look at the code, yet.

dominikschulz (approved):

PR #3496, opened 2026-07-25
feat(gpg): support v5/v6 post-quantum PGP keys with 64-character fingerprints

Q-er:
GnuPG 2.5+ can produce v5 keys (e.g. with PQC algorithms like `ky768_cv25519`) whose fingerprints are 64 hex characters instead of the traditional 40. This PR updates key parsing, ID extraction, and key lookup to handle both v4 and v5 fingerprint lengths.

### Changes

- **`Key.ID()`**: Add a branch for 64-char fingerprints, taking the first 16 characters as the short ID (matching GnuPG's own implementation).
- **`ReadNamesFromKey` / `GetFingerprint`**: Replace `ProtonMail/go-crypto` openpgp calls with `gpg --show-keys --with-colons`, since `go-crypto` does not support v5 key packets and fails to parse them.
- **`key_list.go`**: Match key IDs by prefix in addition to suffix, since v5 short IDs are derived from the start of the fingerprint.

### Tests

- `key_test.go`: Unit tests for v5 key ID extraction from 64-char fingerprints.
- `keyring_test.go`: Unit tests for v5 key parsing via `--show-keys --with-colons` output.

### DIFF
--- a/internal/backend/crypto/gpg/cli/keyring.go
+++ b/internal/backend/crypto/gpg/cli/keyring.go
@@ -3,13 +3,11 @@ package cli
 import (
 	"bytes"
 	"context"
-	"encoding/hex"
 	"fmt"
 	"os/exec"
 	"strings"
 	"text/template"
 
-	"github.com/ProtonMail/go-crypto/openpgp"
 	"github.com/gopasspw/gopass/internal/backend/crypto/gpg"
 	"github.com/gopasspw/gopass/internal/backend/crypto/gpg/colons"
 	"github.com/gopasspw/gopass/internal/out"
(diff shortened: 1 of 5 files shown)


dominikschulz (approved):
So we are back to shelling out and parsing gpg CLI responses ... but not supporting PQC isn't really an option anymore. LGTM.

Q-er:
I agree that falling back to the GPG CLI isn't ideal. It's annoying that we have to work around the broader ecosystem drama (GnuPG's v5 vs. IETF's v6 standard), but that's an issue for them to resolve. Thanks for the review!

The build is currently failing because of #3523

dominikschulz:
@Q-er Thanks. I think we can merge this as-is. #3523 is something I'll need to fix myself.

PR #3502, opened 2026-07-28
docs(adr): resolve the A-13 collision, zero-pad numbers and add an index

dantte-lp:
## What

Renumber one of the two records numbered A-13, zero-pad A-3 through A-9, and add `docs/adr/README.md` as an index.

## Why

Two records carried the number A-13:

```
docs/adr/A-13-expired-gpg-key-handling.md
docs/adr/A-13-screenshot-build-tag.md
```

An ADR number is a stable identifier, so a duplicate makes every citation ambiguous.

## Which record was renumbered, and why

`A-13-expired-gpg-key-handling.md` keeps the number. It is cited from three other documents:

```
docs/commands/recipients.md:137
docs/usecases/team-workflows.md:294
docs/adr/A-14-team-workflows.md:8
```

`A-13-screenshot-build-tag.md` has no inbound citations and is renumbered to A-15.

## Zero-padding

A-3 through A-9 previously used a single digit, which sorts them after A-10 in any lexical listing. None has an inbound citation from another document. The single reference in `internal/backend/storage/fs/rcs.go:22` is updated in this commit.

## The index

`docs/adr/README.md` records the naming rules, the status and authoring date of every record, and three facts that are otherwise only discoverable from git history:

- **A-01 and A-02 are reserved and have no records.** Both are cited from the `CHANGELOG.md` unreleased section ("Split Action handler into focused handler types (A-1)", "Replace context-key config system with typed structs (A-2)") but no file was ever written for either.
- The A-13 collision and which record was renumbered.
- `SECURITY_AUDIT_REPORT.md` and `CODE_QUALITY_REPORT.md`, cited as the `**Source:**` of A-03 through A-10, are not in the tree. Both were removed in `77894053` ("Clean up").

Statuses in the index are taken from each record's `**Status:**` line; dates from `git log --diff-filter=A --follow`.

## Verification

No record content is changed apart from the H1 lines, which must match the file names. `go build ./...` succeeds. Every link in the index resolves, and every record file is indexed. No dangling `A-N-` reference remains anywhere in the tree.

### DIFF
--- a/docs/adr/A-03-separate-storage-rcs.md
+++ b/docs/adr/A-03-separate-storage-rcs.md
@@ -1,4 +1,4 @@
-# A-3: Separate `Storage` and `RCS` interfaces
+# A-03: Separate `Storage` and `RCS` interfaces
 
 **Status:** deferred — current implementation intentionally keeps them merged  
 **Source:** CODE_QUALITY_REPORT.md § A-3
--- a/docs/adr/A-04-grep-match-error-counters.md
+++ b/docs/adr/A-04-grep-match-error-counters.md
@@ -1,4 +1,4 @@
-# A-4: Fix `grep` Match and Error Counters
+# A-04: Fix `grep` Match and Error Counters
 
 **Status:** open — not yet fixed  
 **Source:** SECURITY_AUDIT_REPORT.md § M-1, § Q-5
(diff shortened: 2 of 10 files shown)


dominikschulz (approved):
Thanks for cleaning this up.

Issue #3414, opened 2026-05-03
docs(adr): define API stability contract for pkg/gopass


dimension-zero:
## Background

`ARCHITECTURE.md` (lines 37–43) explicitly notes:

> "we have decided to apply semantic versioning only to the CLI tool, not the Go module. This is not ideal and might change with sufficient active contributors."

`pkg/gopass` is the documented integration point for external consumers. Several known integrations depend on it. There is currently no documented contract for what constitutes a breaking change or how consumers will be notified.

## Problem

A breaking change to `pkg/gopass.Store`, `pkg/gopass.Secret`, or related types will affect consumers silently — the module version does not increment, no changelog entry is mandated, and `go get` will pull the break without warning.

## What an ADR should decide

- **Option A**: Apply semver to the Go module as well. Breaking changes require a major version bump and a `v2` module path (`github.com/gopasspw/gopass/v2`). Requires coordination with existing integrators.
- **Option B**: Publish an explicit "best-effort stable" policy with a dedicated `CHANGELOG` section for `pkg/` breaking changes, and a minimum deprecation window before removal.
- **Option C**: Mark unstable packages explicitly (e.g., `// Unstable: API may change without notice.` doc comment on the package) so consumers understand the contract.
- **Option D**: Move `pkg/gopass` stability guarantees into a separate repository with its own release cycle.

## Affected packages

`pkg/gopass/`, `pkg/gopass/api/`, `pkg/gopass/apimock/`, `pkg/gopass/secrets/`

## Why a formal decision is needed

External integrations are already consuming these packages under an implied stability guarantee. The current situation — where CLI semver is guaranteed but module semver is not — should be a deliberate, documented choice rather than an oversight.

dominikschulz:
Option A is too much churn right now. This project is currently maintained by ~2 people with one of them doing ~100% of the releases and >60% of the reviews. I would love to have more active maintainers, but as things are right now Option A is infeasible.

Option B and C both sound good to me. I think we could do both, but if I'm missing something an they are mututally exclusive I'd prefer B over C.

Option D is my preferred long-term solution, but similar to A this seems to add quite a bit of overhead and I'm not confident we can properly manage that right now.

PR #3503, opened 2026-07-28
docs: add docs/conventions.md and adopt Conventional Commits

dantte-lp:
> Depends on #3502. This branch is stacked on it, so it currently shows two commits. Once #3502 merges, only the conventions commit remains.

## What

Replace the `[TAG]` commit-subject rule in `CONTRIBUTING.md` with Conventional Commits, and add `docs/conventions.md` as the single normative reference for commit messages, versioning, branch and tag names, and file naming.

## Why

`CONTRIBUTING.md:43-45` required a bracketed `[TAG]` prefix and listed nine valid tags. That has not matched practice for some time. Of the 344 commit subjects since 2025-01-01:

```
$ git log --since=2025-01-01 --format='%s' | grep -cE '^[a-z]+(\([^)]*\))?!?: '
225
$ git log --since=2025-01-01 --format='%s' | grep -cE '^\[[A-Za-z-]+\]'
67
```

The `CHANGELOG.md` unreleased section uses a third set including `[SECURITY]` and `[PKG-BREAK]`, neither of which `CONTRIBUTING.md` lists.

## What docs/conventions.md specifies

- **Commit types**, as a closed list, with the SemVer impact and changelog section of each. It documents two extensions to the set suggested by the specification: `security`, which is already used by five commits and populates the Keep a Changelog `Security` section, and `deps`, with `chore(deps)` remaining accepted because Dependabot emits it.
- **Scopes**, derived from the package layout: the 41 top-level subcommands, the backends, the subsystems and the public API packages.
- **The two compatibility surfaces.** A CLI break uses `!` and a `BREAKING CHANGE:` footer and forces a major release. A break confined to `pkg/gopass` uses a `PKG-BREAK:` footer, no `!`, and does not. The `PKG-BREAK:` footer is the machine-readable form of the changelog tag mandated by ADR A-12.
- **Semantic Versioning**, and which surfaces it does and does not cover.
- **Branch and tag names**, including the `release/` and `prep/` prefixes owned by the release automation.
- **File naming** for ADRs, documentation and Go sources.

It also records five values that appear as commit types in the history but are scopes: `otp` (2 commits since 2025-01-01), `age`, `bug`, `fscopy` and `openbsd` (1 each).

## The DCO is unchanged

Conventional Commits governs the subject line; the Developer Certificate of Origin adds a trailer. The two are independent and both continue to apply. `CONTRIBUTING.md:35-41` is untouched.

## Corrections to ARCHITECTURE.md

The API Stability section still described `pkg/gopass/doc.go` as carrying "an explicit instability warning" and instructed consumers to "treat any `pkg/` type or function change as potentially breaking". Both statements predate ADR A-12. `doc.go` now declares the package best-effort stable and permits additive changes in any release. The section also referred to issue #3414 as an open decision; that decision is recorded in A-12 with status `accepted`.

## AGENTS.md

The folder list did not mention five `pkg/` directories that exist: `otp`, `passkey`, `pinentry/cli`, `protect` and `qrcon`. Each is described using its own package doc comment. Note that `pkg/pinentry` itself holds no Go files; `cli` is the only package under it.

## Verification

Every relative link in the changed and added documents resolves. `make test` passes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -37,10 +37,23 @@ The project is specifically targeting users on all major platform, i.e. Linux, U
 - `/pkg/debug`: Contains a debug package with different verbosity levels. Use it to output debug information to a debug log.
 - `/pkg/fsutil`: Contains various helpers for interacting with the filesystem, e.g. checking for presence of files or directories. Prefer those over implementing these checks from scratch.
 - `/pkg/gopass`: Contains the public gopass API to interact with existing password stores. The `api` sub package contains the actual API and the `secrets` sub package the different secret types we support.
+- `/pkg/otp`: Contains functions to handle OTP secrets. It parses OTP secrets from various formats and generates QR codes for them.
+- `/pkg/passkey`: Implements support for WebAuthn credentials for authentication.
+- `/pkg/pinentry/cli`: Contains a pinentry client that uses the terminal for input and output. It is a drop-in replacement for the `pinentry` program, used to ask for a passphrase or PIN. Note that `/pkg/pinentry` itself holds no Go files; `cli` is the only package under it.
+- `/pkg/protect`: Provides an interface to the `pledge` syscall, used to limit the system calls the process can make. `pledge` exists only on OpenBSD; this package is a no-op everywhere else.
 - `/pkg/pwgen`: Contains a pure-Go implementation of the `pwgen` utility.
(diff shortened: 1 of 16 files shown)


dominikschulz (approved):
Very nice. Thanks! 👍

PR #3509, opened 2026-07-29
fix(age): send agent identities space-separated on a single line

### LINKED ISSUE #3507: age: agent loads at most one identity per identities send (newline framing) → intermittent agent decrypt
With `age.agent-enabled: true`, gopass sends its identities to the age agent as a single `Fprintln(conn, "identities " + ids)` where ids is newline-separated. The agent, however, reads its line-oriented protocol with a `bufio.Scanner` and parses only the tokens on the first line of the identities command. Every identity after the first is therefore read as a separate line, falls through to default: `ERR unknown command`, and is silently dropped. The agent ends up loading at most one identity per send — and zero when an SSH identity sorts first (see below).

Because the identity order comes from Go map iteration (randomized per process), which identity lands "first" is non-deterministic. This makes agent-backed decryption intermittent.

## Impact

- The agent effectively caches at most one identity. With multiple native identities, only whichever sorts first is available to the agent; secrets encrypted to any other identity fall back to (slower, pinentry-triggering) local decryption.
- With SSH identities present, the agent is empty whenever an SSH identity sorts first, so no secret is served from the agent cache in those runs.
- User-visible symptom: the same secret sometimes decrypts via the agent and sometimes via local fallback; the browser extension and CLI can appear to behave differently.

## Reproduction

1. `gopass init --crypto age`; ensure a native age identity exists and you have SSH keys in `~/.ssh/`.
2. Enable the agent (`age.agent-enabled: true`) and start it: 
   ```&h
   GOPASS_DEBUG=1 GOPASS_DEBUG_LOG=~/age-agent.log gopass age agent start
   ```
3. Repeatedly read a secret with debug logging:
   ```sh
   GOPASS_DEBUG=1 GOPASS_DEBUG_LOG=~/age-client.log gopass show <entry>
   ```
4. Inspect the log. Most runs show an SSH identity sorting first and no loaded N identities line; only when the native identity happens to sort first does the agent load anything:
   ```log
   received: identities &{[244 .. 169]}   # SSH identity first
   received: &{[161 .. 81]}
   received: AGE-SECRET-KEY-1R..VA
   received: set-timeout 3600
   received: decrypt.                     # no "loaded N identities" — bundle failed to parse
   ```
   vs.
   ```log
   received: identities AGE-SECRET-KEY-1R..VA
   loaded 1 identities                     # only this ordering loads anything
   ```

## Root cause

Framing mismatch. The client newline-joins the identities and writes them with one Fprintln:

https://github.com/gopasspw/gopass/blob/eb1368e493cc7f8484fe38e502a2e155596883a3/internal/backend/crypto/age/age.go#L229-L236

So the wire payload is identities `<id1>\n<id2>\n<id3>\n` — i.e. three lines. `tryStartAgent` does the same (`fmt.Sprintf("%s", id) + strings.Join(idStrs, "\n")`).

The agent reads line by line and parses only the current line:

https://github.com/gopasspw/gopass/blob/eb1368e493cc7f8484fe38e502a2e155596883a3/internal/backend/crypto/age/agent/agent.go#L111-L119

`args` is `strings.Fields` of the current line only. Line 2 (id2) and line 3 (id3) are parsed as standalone commands → default → ERR unknown command, discarded. Only the first identity reaches parseIdentities. The client's send() reads only the first response line and returns, so the trailing ERR replies are never observed — the loss is silent.

Note that the agent's identities handler already re-joins its args with newlines before parsing (strings.Join(args, "\n")) — i.e. it was designed to receive identities space-separated on a single line. The client simply never sends them that way.

SSH identities cannot be serialized at all. `filippo.io/age/agessh` identity types (`*Ed25519Identity`, `*RSAIdentity`, `*EncryptedSSHIdentity`) implement no `String()` method, so `fmt.Sprintf("%v", id)` emits a Go struct dump such as `&{[185 .. 233]}`, which `parseIdentities` rejects as unknown identity type. When an SSH identity sorts first, the entire identities command fails to parse and zero identities load.

Order is randomized. getAllIds iterates the map returned by getAllIdentities:

https://github.com/gopasspw/gopass/blob/eb1368e493cc7f8484fe38e502a2e155596883a3/internal/backend/crypto/age/decrypt.go#L125-L136

Each gopass / gopass-jsonapi invocation is a fresh process that rebuilds this map with a fresh random hash0 seed, so the order — and therefore which single identity (if any) the agent loads — differs per request.

## Suggested fix

Send identities space-separated on a single line, matching what the agent already parses (e.g. join with " " instead of "\n" in `identitiesToString`, and have `tryStartAgent` use the same helper):

- Space-joining places every identity into args, which the agent re-joins with \n and parses in full.
- SSH identities must be skipped (no serializable private-key form); decryption for them already falls back to the local path today.
- Validate that each encoding is whitespace-free (native age/plugin encodings are bech32 and contain no whitespace).

## Environment

- gopass 1.16.1
- age backend, `age.agent-enabled: true`
- 1 native (X25519) identity + several SSH keys under `~/.ssh/`

yzx9:
The age agent reads its line-oriented protocol with a bufio.Scanner and parses only the tokens on the "identities" command line (it re-joins args with newlines before parsing). The client, however, serialized identities newline-separated and wrote them with a single Fprintln, so every identity after the first landed on its own line and was discarded as an "unknown command". As a result the agent loaded at most one identity per send.

This was masked by Go's randomized map iteration order in `getAllIds`: each gopass process sent the identities in a different order. When a native X25519 identity happened to sort first the agent loaded it and decryption via the agent succeeded; when an SSH identity sorted first the whole bundle failed to parse and nothing loaded. Hence the browser extension and the CLI appeared to behave differently.

SSH identities compound this: their private-key types implement no String() method, so they formatted as an unparseable Go struct (e.g. `&{[185 .. 233]}`) and could never be sent to the agent at all.

Fix, client-side (the agent already expected this format):

  * identitiesToString now emits all identities space-separated on one line, matching what the agent parses.
  * Only natively serializable identity types (X25519, Hybrid, plugin and gopass plugin/wrapped identities) are included; SSH identities are skipped (decryption for them falls back to the local code path exactly as before — they never worked via the agent). Each encoding is validated to be whitespace-free.
  * `tryStartAgent` now uses `identitiesToString` too, replacing its own `fmt.Sprintf("%s", id)` + newline-join loop (which additionally mangled non-Stringer types as %!s(...)).
  * Callers skip the send when there is nothing serializable to avoid a noisy "missing identities" agent error.

closes: #3507

### DIFF
--- a/internal/backend/crypto/age/age.go
+++ b/internal/backend/crypto/age/age.go
@@ -152,20 +152,23 @@ func (a *Age) tryStartAgent(ctx context.Context) {
 	}
 
 	// send identities to agent
-	ids, err := a.getAllIdentities(ctx)
+	ids, err := a.getAllIds(ctx)
 	if err != nil {
 		debug.Log("failed to get identities: %s", err)
 
 		return
 	}
 
-	idStrs := make([]string, 0, len(ids))
-	for _, id := range ids {
-		idStrs = append(idStrs, fmt.Sprintf("%s", id))
-	}
+	sIds, err := a.identitiesToString(ids)
+	if err != nil {
+		debug.Log("failed to serialize identities: %s", err)
 
-	if err := client.SendIdentities(strings.Join(idStrs, "\n")); err != nil {
-		debug.Log("failed to send identities to agent: %s", err)
+		return
+	}
+	if sIds != "" {
+		if err := client.SendIdentities(sIds); err != nil {
+			debug.Log("failed to send identities to agent: %s", err)
+		}
 	}
 
 	// set timeout
@@ -226,13 +229,35 @@ func (a *Age) Lock() {
 	a.askPass.Lock()
 }
 
+// identitiesToString serializes the given identities into a single-line,
+// space-separated string for the age agent's line-oriented "identities"
+// command.
+//
+// The agent reads commands one line at a time and parses only the tokens on
+// that line (it re-joins them with newlines before parsing). Sending identities
+// newline-separated therefore turns every identity after the first into a
+// spurious "unknown command" that the agent discards. All identities must live
+// on a single line, i.e. space-separated.
+//
+// Only natively serializable identity types are included. SSH identities
+// (filippo.io/age/agessh) expose no String() form for the private key and
+// cannot be round-tripped through the wire format; they are skipped here, and
+// decryption for them falls back to the local code path exactly as before. All
+// of these encodings are bech32 and therefore whitespace-free, so separating
+// them with spaces is unambiguous to the agent's space-splitting parser.
 func (a *Age) identitiesToString(ids []age.Identity) (string, error) {
-	var sb strings.Builder
+	parts := make([]string, 0, len(ids))
 	for _, id := range ids {
-		fmt.Fprintln(&sb, id)
+		s, ok := identityToString(id)
+		if !ok {
+			debug.Log("skipping non-serializable identity %T for agent transfer", id)
+
+			continue
+		}
+		parts = append(parts, s)
 	}
 
-	return sb.String(), nil
+	return strings.Join(parts, " "), nil
 }
 
 // String implements fmt.Stringer.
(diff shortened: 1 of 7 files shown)


dominikschulz (approved):
LGTM. Thanks!

PR #3510, opened 2026-07-29
age: don't assumes `os.Args[0]` implements the `age agent start` subcommand

### LINKED ISSUE #220: Age agent auto-start causes recursive fork bomb when agent isn't pre-started
When the age crypto backend's agent auto-start is enabled (`age.agent-enabled = true`) and no `gopass` age agent is already running, triggering the browser extension causes `gopass-jsonapi` to enter a recursive fork bomb: it repeatedly spawns copies of itself (`gopass-jsonapi age agent start`), each of which spawns another, until the process count explodes and CPU is pinned at 100%, making the machine unresponsive.

The root cause is in the embedded `gopass` library's `startAgent`, which re-execs `os.Args[0]` with `age agent start`. Because the running binary is `gopass-jsonapi` (which has no `age` subcommand), the spawned process never starts an agent — instead it re-initializes the gopass store (`api.New`), which re-triggers agent auto-start, which spawns yet another copy. There is no recursion guard, so the cascade is unbounded.

Notably, if the agent is **pre-started by the real `gopass` binary** before the extension talks to `gopass-jsonapi`, everything works fine (the very first `Ping()` succeeds and no spawn happens).

## Steps To Reproduce

1. Use an age-encrypted gopass store and enable the agent:

   ```sh
   gopass config age.agent-enabled true
   ```
2. Make sure no agent is running yet:

   ```
   gopass age agent status   # -> Age agent is not running
   ```
3. Trigger the browser extension `gopassbridge` to talk to `gopass-jsonapi`.
4. Observe: the process count climbs rapidly (`pgrep -fl gopass-jsonapi` shows many gopass-jsonapi age agent start children), CPU saturates, the machine becomes unresponsive.

## Expected behavior

`gopass-jsonapi` should be able to start `gopass` age agent.

### Environment

- OS: `macOS (Darwin 25.5.0 Darwin Kernel Version 25.5.0: Mon Apr 27 20:39:42 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T6031 arm64 arm Darwin)`
- `gopass-jsonapi` version: `1.16.1`
- `gopassbridge` version: `2.1.0`
- Installation method: `nixpkgs-unstable`

### Root-cause analysis

1. `startAgent` try to spawns `<gopass-jsonapi> age agent start`.

https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/agent_starter_unix.go#L12-L23

2. But there is no `age` command in `gopass-jsonapi`. The spawned process never runs an agent server, instead it runs `api.New` again → `tryStartAgent` again → spawns another child → unbounded recursion.

### Suggested fixes

- In `gopass-jsonapi`: avoid running the full `api.New` store initialization when the process is being invoked as an agent starter. Add a mechanism to prevent infinite restarts, such as implementing a delay.
- In `gopass` core, `startAgent` should not blindly re-exec `os.Args[0]`. Instead, it should locate the actual gopass binary. Unfortunately, this approach does not apply to `gopass-jsonapi`, because this repository uses gopass as a library and does not involve a standalone binary.

## Workaround

Start the agent wilogin (e.g. via a systemd / launchd) before the browser uses it:

```sh
gopass age agent start &
gopass age agent unlock
```

or disable the feature until fixed: `gopass config age.agent-enabled false`.

### LINKED ISSUE #3495: age: `startAgent` assumes `os.Args[0]` implements the `age agent start` subcommand`
When `age.agent-enabled` is true and no age agent is running, the age backend spawns a new process to start the agent by re-executing `os.Args[0]` with the arguments age agent start:

https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/agent_starter_unix.go#L12-L23

This assumes `os.Args[0]` is a binary that implements the age agent start subcommand — i.e. the standalone gopass binary. That holds for gopass itself, but gopass is also distributed as a Go library, and consumers that embed it without exposing the age CLI subcommand hit an unbounded recursive fork bomb:

1. `startAgent` execs `<host-binary> age agent start`.
2. The host binary has no age subcommand, so the spawned process never starts an agent. Instead it re-runs its normal entry path, which re-constructs the age backend (`age.New` → `tryStartAgent`).
3. `tryStartAgent` pings, sees no agent, and spawns another copy. There is no recursion guard, so the cascade runs until the machine is saturated (100% CPU, unresponsive).

This was originally filed against the consumer (gopasspw/gopass-jsonapi#220), but the root cause is the `os.Args[0]` re-exec in gopass core, with no guard against the spawned process not being gopass.

## To reproduce

The standalone gopass binary is not affected (its os.Args[0] is gopass, which owns age agent start). The bug reproduces in any library consumer that constructs the age backend without exposing the age subcommand. The known case is gopass-jsonapi:

1. Use an age-encrypted store and enable the agent:
   ```
   gopass config age.agent-enabled true
   ```
2. Ensure no agent is running:
   ```sh
   gopass age agent status   # -> Age agent is not running
   ```
3. Trigger the gopassbridge browser extension to talk to gopass-jsonapi.
4. Observe: `pgrep -fl gopass-jsonapi` shows many `gopass-jsonapi age agent start` children multiplying rapidly, CPU pinned at 100%, machine unresponsive.

If the agent is pre-started by the real gopass binary before the extension talks to gopass-jsonapi, the very first `Ping()` succeeds and no spawn happens — confirming the spawn path is the culprit.

Full consumer-side repro and traces: gopasspw/gopass-jsonapi#220.

## Expected behavior

`tryStartAgent` should not fork-bomb when embedded in a binary that isn't gopass. At minimum, the spawned process must not re-enter `tryStartAgent` unbounded.

## Root cause

-  Calls `a.tryStartAgent(ctx)` from `New`.
   https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/age.go#L76
- `tryStartAgent` pings; on failure (line 136) calls `startAgent`.
  https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/age.go#L121-L125
  https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/age.go#L136-L140
- `startAgent` execs `os.Args[0]` age agent start with no recursion guard.
  https://github.com/gopasspw/gopass/blob/f25fc7b4e19a66ede32e05a073e24fc04c0fdbc8/internal/backend/crypto/age/agent_starter_unix.go#L12-L16

  (The Windows counterpart in internal/backend/crypto/age/agent_starter_windows.go is identical.)

## Suggested fixes

- Recursion guard (minimal, stops the fork bomb in all cases): set a private env var (e.g. GOPASS_AGE_AGENT_SPAWNING=1) on the spawned command, and have tryStartAgent short-circuit when it is already set, so a mis-targeted spawn cannot cascade.
- Don't assume os.Args[0] is gopass: resolve the real gopass binary via exec.LookPath("gopass") and fall back gracefully (skip auto-start) if not found. Version skew is a concern, but better than a fork bomb.
- Or start the agent in-process using the existing agent package (agent.New() + Run() in a goroutine) instead of re-exec'ing — removes the os.Args[0] dependency entirely and is friendlier to library consumers.
- Or make the agent starter injectable (a context value or an Age option) so embedders can supply their own binary path / start strategy; keep the current os.Args[0] behavior as the default for the standalone binary.

## Environment

- OS: macOS (Darwin 25.5.0, arm64)
- gopass: master (bug present in current internal/backend/crypto/age/agent_starter_unix.go and agent_starter_windows.go); reproducer uses gopass-jsonapi 1.16.1 + gopassbridge 2.1.0 (nixpkgs-unstable).

## Workaround

Pre-start the agent with the real gopass binary before the consumer needs it:

```sh
gopass age agent start &
gopass age agent unlock
```

or disable the feature: `gopass config age.agent-enabled false`.

yzx9:
This PR introduces two main changes:

1.  It adds an environment variable to prevent infinite attempts to spawn a process. Even after our second commit fix, users could still potentially trigger this bug by using gopass in a wrapper, leading to an accidental activation.
2.  It introduces an `AgentLauncher` interface and moves the agent start code from the backend to the CLI. I've temporarily placed it in a new package under `internal`, following existing conventions, but I'm not sure if it's the best location. If you have other suggestions, please let me know.

Additionally, after the second commit fix, the client (e.g. `gopass-jsonapi`) will no longer automatically start our agent. This could be a issue that might require a broader fix, and I'm currently unsure how we should proceed. There are two possibilities:

* For services like the `gopass-jsonapi`, We could potentially follow the gopass pattern and create a similar command to start the age agent.
* Additionally, we could provide a lightweight `gopass-age-agent` helper to trigger this function.
* Alternatively, we could recommend that users manage the agent using methods like `systemd` or `launchd`, rather than relying on our code for automatic startup.

closes #3495

closes gopasspw/gopass-jsonapi#220


yzx9:
@dominikschulz Feel free to ping me with any questions on this PR, especially around how the API should support age agents going forward.

My current idea is for `gopass-jsonapi` to start the agent directly, similar to the CLI.

dominikschulz (approved):
LGTM. Thanks!

dominikschulz:
@yzx9 Thanks a lot. Sorry for the delay, I've been very busy.

yzx9:
@dominikschulz No worries! Thanks for taking the time to look into the PR. Much appreciated!

PR #3525, opened 2026-08-11
fix: require --age-sshkeys (or age.sshkeys config) to access ~/.ssh

### LINKED ISSUE #3522: Docs and code conflicting about age.ssh-key-path; cannot make gopass not read identities from ~/.ssh
I think I am running into https://github.com/gopasspw/gopass/issues/3507 which is indeed a flaw in itself. But I was originally trying to make gopass not read any sshkeys from ~/.ssh at all; i have all sorts of things there that I don't want gopass to access at all.

The docs in https://github.com/gopasspw/gopass/blob/master/docs/config.md says that `age.ssh-key-path` (only) will be used if set, otherwise `~/.ssh` will be used. So I was trying to set `age.ssh-key-path` to `~/.non-existant`, to make gopass not load *any* SSH identities, but that did not work. Because [the code](https://github.com/gopasspw/gopass/blob/master/internal/backend/crypto/age/ssh.go#L41) uses both dirs to find SSH identities!

I really think there should be a way to not have gopass access ~/.ssh -- I think making it to what the docs says would make sense. Or possibly some other way to turn off SSH identities entirely. What do you say @dominikschulz

quite:
This makes actual functionality follow status quo documentation. Also does not access ~/.ssh unless user ir explicit about that.

If age.ssh-key-path is set, it is loaded in addition to age.sshkeys, if that is enabled, or instead of those, if not.

Fixes #3522


quite:
The revised documentation tries to make it clear that `age.ssh-key-path` can be used no matter if `age.sshkeys` is `true` or `false`. If it's `true`, `age.ssh-key-path` is for loading additional keys, if `false`, it's effectively for loading alternative keys. But not sure it is really understandable?

quite:
Confusing that some tests fail like this:
```
# github.com/gopasspw/gopass/internal/backend/crypto/age [github.com/gopasspw/gopass/internal/backend/crypto/age.test]
Error: internal/backend/crypto/age/age_test.go:46:21: not enough arguments in call to New
	have (context.Context, string)
	want (context.Context, bool, string)
Error: internal/backend/crypto/age/age_test.go:68:22: not enough arguments in call to New
	have (context.Context, string)
	want (context.Context, bool, string)
Error: internal/backend/crypto/age/age_test.go:89:22: not enough arguments in call to New
	have (context.Context, string)
	want (context.Context, bool, string)
     FAIL	github.com/gopasspw/gopass/internal/backend/crypto/age [build failed]
FAIL
```

The calls to `age.New()` in `age_test.go` really have been revised for the added bool. Also in the above, only the first of the line numbers (46, 68, 89) actually point to a line with a `New(...)`. Hm.

internal/backend/crypto/age/age.go
```diff
@@ -46,13 +38,14 @@ type Age struct {
...
-	sshKeyPath      string // custom SSH key or directory path
+	sshKeys         bool   // load (or not) SSH keys from default SSH dir (~/.ssh), or dir in GOPASS_SSH_DIR env var
+	sshKeyPath      string // custom SSH key or directory path; if set, loaded even if the above sshKeys is false
 	pwCallback      func(string, bool) ([]byte, error)
 	pwPurgeCallback func(string)
 }
 
 // New creates a new Age backend.
-func New(ctx context.Context, sshKeyPath string) (*Age, error) {
+func New(ctx context.Context, sshKeys bool, sshKeyPath string) (*Age, error) {
```
dominikschulz: `sshKeys` isn't very explanatory. Consider renaming this to `loadSshKeys`.

internal/backend/crypto/age/age_test.go
```diff
@@ -14,117 +11,39 @@ import (
...
-	a4, err := New(t.Context(), "")
+	a4, err := New(t.Context(), false, "")
 	require.NoError(t, err)
 	assert.Empty(t, a4.sshKeyPath)
 }
 
 func TestInitialized(t *testing.T) {
 	ctx := t.Context()
-	a, err := New(ctx, "")
+	a, err := New(ctx, false, "")
```
dominikschulz, line 46: I don't see what's wrong here right now, but we need to fix this before we can merge the PR.

internal/backend/crypto/age/ssh.go
```diff
@@ -40,13 +40,17 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
...
 	sshDirs := make([]string, 0, 2)
 
-	sshDir, err := getSSHDir()
-	if err != nil {
-		debug.Log("no .ssh directory found at %s.", sshDir)
-	}
-	if sshDir != "" {
-		debug.Log("found .ssh directory at %s", sshDir)
-		sshDirs = append(sshDirs, sshDir)
+	if a.sshKeys {
```
dominikschulz: Optional: the `getSSHIdentities` method is getting pretty long and nested. We should consider some light refactorings.

internal/backend/crypto/age/ssh.go
```diff
@@ -75,7 +79,7 @@ func (a *Age) getSSHIdentities(ctx context.Context) (map[string]age.Identity, er
 		debug.Log("searching for SSH identities in %s", sshDir)
 		files, err := os.ReadDir(sshDir)
 		if err != nil {
-			debug.Log("unable to read .ssh dir %s: %s", sshDir, err)
+			debug.Log("unable to read SSH keys dir %s: %s", sshDir, err)
```
dominikschulz: nit: Add a "from" after "SSH Keys".

dominikschulz (changes requested):
Looks pretty good, just some small nits. Thanks!

internal/backend/crypto/age/age.go
quite: You mean just for this argument name, right?

internal/backend/crypto/age/ssh.go
quite: I won't address this here. I think one complicating factor is that ssh-key-path accepts both a single file, and a dir. Can't as easily break out 2 functions (1 for sshKeys, 1 for sshKeyPath) that only return a dir to be added to sshDirs. Otherwise that could have made sense.

internal/backend/crypto/age/ssh.go
quite: Done

internal/backend/crypto/age/age_test.go
quite: Smells some cache somewhere. But that also does not make sense to me in this case.

quite:
CodeQL CI step also fails:
```
Run github/codeql-action/analyze@e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81
Error: Loaded a configuration file for version '4.36.1', but running version '4.37.3'
``

internal/backend/crypto/age/age.go
quite: Or user-facing as well?

dominikschulz: I meant only this instance (and the places where it's used in this class, of course). The user facing value can stay as is if you want. But maybe it would make sense to rename that as well - I'm not sure right now. Your call.

dominikschulz:
@quite The CodeQL CI failure shouldn't be your concern. But I'm worried about the other builds failing. Please don't spend too much time on this. I've merged over a dozen pending PRs yesterday. Maybe some GitHub Action broke something. I'll try to take a look at that soon. But if you find the root cause, that would help :smile:

internal/backend/crypto/age/age.go
quite: Done (changed all variables named sshKeys -> loadSSHKeys)

quite:
Last forcepush fixes in commands.go so that `age.sshkeys = true` in config remains true also when cmdline does NOT have `--age-sshkeys`.

dominikschulz:
I can't reproduce the test failures locally and most of the other failing actions should have been fixed by other PRs. So I'm going ahead and merge this.

PR #3529, opened 2026-08-14
fix(age): raise agent scanner buffer limit to handle large decrypt payloads

### LINKED ISSUE #3508: age: agent `bufio.Scanner` 64 KiB line limit silently drops large `decrypt` commands
The gopass age agent reads its line-oriented protocol with `bufio.NewScanner(conn)` and never raises the default token cap (`bufio.MaxScanTokenSize` = 64 KiB), nor checks `scanner.Err()` after the read loop. Any single command line longer than 64 KiB makes `scanner.Scan()` fail with `bufio.ErrTooLong`; `handleConnection` exits silently, the connection is closed, and the client receives EOF.

This is reachable today via the `decrypt` command, which carries the base64-encoded ciphertext of a secret on a **single line**. A secret whose plaintext is ≳ ~48 KiB produces a `decrypt` line exceeding 64 KiB.

## To reproduce

```bash
gopass config age.agent-enabled true
gopass age agent start

# store a secret larger than ~48 KiB
head -c 50000 /dev/urandom | base64 | gopass insert -f big/secret

GOPASS_DEBUG=1 gopass show big/secret
gopass age agent unlock   # -> "failed to send identities to agent: EOF"
```

**Expected**: the secret decrypts through the agent.

**Actual**: the decrypt line exceeds 64 KiB → bufio.ErrTooLong → the connection is dropped. Age.Decrypt falls back to local decryption (works, but with a confusing debug log); gopass age agent unlock errors with failed to send identities to agent: EOF and never calls Unlock, leaving the agent un-unlocked.

## Root cause

https://github.com/gopasspw/gopass/blob/eb1368e493cc7f8484fe38e502a2e155596883a3/internal/backend/crypto/age/agent/agent.go#L111-L117

The client sends the whole ciphertext as one line, so the decrypt command has always been subject to this limit. It is independent of any other change.

https://github.com/gopasspw/gopass/blob/eb1368e493cc7f8484fe38e502a2e155596883a3/internal/backend/crypto/age/agent/client.go#L90-L97

## Impact

- `decryptWithAgent` / `tryStartAgent`: the error is logged and swallowed; decryption falls back to the local path. Not a crash, but the agent silently stops helping and the debug output is confusing.
- gopass age agent unlock (commands.go): the `SendIdentities` error is returned to the user as failed to send identities to agent: EOF (exit Unknown) and client.Unlock() is never reached — the agent is left un-unlocked. This is the user-visible failure.
- Because scanner.Err() is never checked, any scanner failure (over-long line or I/O error) ends connection handling with no diagnostic.

## Suggested fix

There are two possible ways to address this issue.

### Option 1: Increase the `bufio.Scanner` token limit

In `handleConnection`:

```go
scanner := bufio.NewScanner(conn)
scanner.Buffer(make([]byte, 0, 64*1024), 1<<20) // raise cap to 1 MiB
for scanner.Scan() {
    // ...
}
if err := scanner.Err(); err != nil {
    debug.Log("agent connection scan error: %s", err)
}
```

1 MiB covers secrets up to ~750 KiB and is consistent in scale with the inner parseIdentities reader (which already uses a 16 MiB io.LimitReader). Alternatively, chunk large payloads instead of raising the cap.

### Option 2: Separate the command header from the payload

A more robust solution would be to avoid encoding the entire encrypted payload as part of one newline-delimited command.

Instead of sending:

```text
decrypt <base64-encoded-ciphertext>\n
```

the protocol could send a small command header, followed by exactly that many payload bytes:

```text
decrypt-start\n
decrypt-chunk <chunk bytes>\n
decrypt-chunk <chunk bytes>\n
decrypt-end\n
```

Therefore, increasing `Scanner.Buffer` is likely the appropriate short-term fix, while a length-prefixed binary payload would be the more scalable long-term protocol design.

However, there is a potential issue here: we have no way to detect whether a agent supports a particular command. This could lead to incompatibility issues between versions.

## Environment

- gopass master (and 1.16.1)
- age backend, age.agent-enabled = true

yzx9:
The age agent read its line-oriented protocol with a default
bufio.Scanner, which caps lines at 64 KiB (`bufio.MaxScanTokenSize`).
The decrypt command carries the base64-encoded ciphertext on a
single line, so any secret larger than ~48 KiB made scanner.Scan()
fail with `bufio.ErrTooLong`, silently dropping the connection and
leaving the client with EOF (e.g. "failed to send identities to
agent: EOF" from gopass age agent unlock).

Raise the scanner buffer cap to 16 MiB, matching the
`privateKeySizeLimit` already used when parsing identities. No buffer
is preallocated: the scanner starts at bufio's small initial size and
only doubles as the longest line actually received requires, so
this does not allocate anything upfront. Also log
scanner.Err() so future scan failures are no longer silent.

Implements Option 1 from #3508. A chunked or length-prefixed
protocol (Option 2) requires agent protocol versioning first.

Fixes #3508

### DIFF
--- a/internal/backend/crypto/age/agent/agent.go
+++ b/internal/backend/crypto/age/agent/agent.go
@@ -114,6 +114,13 @@ func (a *Agent) handleConnection(ctx context.Context, conn net.Conn) {
 	}()
 
 	scanner := bufio.NewScanner(conn)
+	// the decrypt command carries the base64-encoded ciphertext on a single
+	// line, so we need to raise the default 64 KiB token limit to handle
+	// larger secrets (see issue #3508). No buffer is preallocated: the
+	// scanner starts at bufio's small initial size and only doubles as the
+	// longest received line requires, up to this limit. The 16 MiB maximum
+	// matches the privateKeySizeLimit used when parsing identities.
+	scanner.Buffer(nil, 1<<24) // 16 MiB max line size
 	for scanner.Scan() {
 		line := scanner.Text()
 		debug.Log("received: %s", line)
@@ -216,6 +223,9 @@ func (a *Agent) handleConnection(ctx context.Context, conn net.Conn) {
 		default:
 			fmt.Fprintln(conn, "ERR unknown command")
 		}
+	}
+	if err := scanner.Err(); err != nil {
+		debug.Log("agent connection scan error: %s", err)
 	}
 }
 
--- a/internal/backend/crypto/age/agent/agent_test.go
+++ b/internal/backend/crypto/age/agent/agent_test.go
@@ -229,3 +229,48 @@ func TestAgentMultipleIdentities(t *testing.T) {
 	// cleanup
 	require.NoError(t, c.Quit())
 }
+
+func TestAgentLargePayload(t *testing.T) {
+	ctx := t.Context()
+	ctx = termio.WithPassPromptFunc(ctx, func(ctx context.Context, prompt string) (string, error) {
+		return "test", nil
+	})
+
+	// start agent
+	a, err := New()
+	require.NoError(t, err)
+
+	go func() {
+		_ = a.Run(ctx)
+	}()
+	defer a.Shutdown(ctx)
+
+	// wait for it to be ready
+	time.Sleep(time.Second)
+
+	// create client
+	c := NewClient()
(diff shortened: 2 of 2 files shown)


dominikschulz (approved):
LGTM. Thank you!

PR #3538, opened 2026-08-20
ci: restore security workflow reliability

dantte-lp:
## Summary

- keep all CodeQL phases on the same pinned v4.37.3 commit
- correct stale CodeQL version annotations
- run the updater key-expiry canary weekly or on demand instead of making every push and PR fail during the documented key rollover window

## Evidence

The current CodeQL job fails because init/autobuild load 4.36.1 while analyze runs 4.37.3. The updater canary correctly predicts that the 2025 signing key will be expired six months from now; adding the 2026 public key does not change the signature embedded in the canary. Keeping that signal scheduled preserves the warning without making unrelated changes permanently red.

## Validation

- `actionlint -shellcheck= .github/workflows/codeql-analysis.yml .github/workflows/key-canary.yml`
- verified `v4.37.3` peels to `e4fba868fa4b1b91e1fdab776edc8cfbe6e9fb81`

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -68,7 +68,7 @@ jobs:
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@87557b9c84dde89fdd9b10e88954ac2f4248e463 # v3.29.5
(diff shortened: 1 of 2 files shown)


dominikschulz (approved):
Thanks!

Issue #3118, opened 2025-04-17
exit status 2: |gpg: Fatal: can't create directory '/home/runner/.gnupg': File exists


dominikschulz:
Please report instances of flaky tests due to `exit status 2: |gpg: Fatal: can't create directory '/home/runner/.gnupg': File exists` here.

We should try to weed out these flakes and I think we might be able to, but we need to collect instances of these failures first.

dantte-lp:
This flake has recurred on current `master`-based PRs on 2026-08-20 in two independent jobs:

- Linux Go 1.25.13: `TestReadNamesFromKey` failed because `/home/runner/.gnupg` already existed ([run 32411941858](https://github.com/gopasspw/gopass/actions/runs/32411941858/job/96564135070)).
- macOS arm64 / GnuPG 2.5.21: `TestGetFingerprintV5` failed for the same reason ([run 32411949137](https://github.com/gopasspw/gopass/actions/runs/32411949137/job/96564157897)).

Both branches only changed CI configuration, so this is unrelated to their diffs. Please consider reopening. The likely durable direction is to give GPG CLI tests an isolated `GNUPGHOME`/`GOPASS_HOMEDIR` per test or package instead of racing on the runner home.

dominikschulz:
I believe #3557 has fixed this. I have not seen this kind of error since then.

PR #3557, opened 2026-08-20
test(gpg): isolate package GnuPG home

dantte-lp:
## Summary

- create a package-scoped temporary `GNUPGHOME` before parallel GPG CLI tests start
- pre-create it with mode `0700` to remove the concurrent GnuPG home-creation race
- preserve existing parallel test execution and production behavior

Fixes the recurring `gpg: Fatal: can't create directory ~/.gnupg: File exists` failure reported in #3118.

## Validation

- `go test -count=20 ./internal/backend/crypto/gpg/cli`
- `go test -race -shuffle=on -count=5 ./internal/backend/crypto/gpg/cli`
- `make test` reaches the pre-existing root-only `TestIsUpdateable` failure after the GPG package passes
- master codequality has the pre-existing `internal/notify/icon.go` nlreturn finding addressed by #3498

### DIFF
--- a/internal/backend/crypto/gpg/cli/main_test.go
+++ b/internal/backend/crypto/gpg/cli/main_test.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


dominikschulz (approved):
Thanks! This is great 👍

PR #3560, opened 2026-08-21
docs: replace retired Go Report Card badge with golangci-lint CI badge

yzx9:
image

Go Report Card has been retired, so the badge no longer renders useful status. Replace it with a GitHub Actions workflow status badge for our [golangci-lint](https://github.com/gopasspw/gopass/actions/workflows/golangci-lint.yml) workflow, styled like the existing build status badge.

---

New:

image


dominikschulz (approved):
Thanks!

Issue #3074, opened 2025-03-02
Export function?


formula-spectre:
I would like to export my password to Bitwarden. can an official export function be implemented? something that outputs a json or csv that can be read from bitwarden

AnomalRoil:
In principle any `pass` -> bitwarden exporter should work. Have your tried e.g. https://github.com/philsnow/pass-to-bitwarden ?

dominikschulz:
In case that doesn't work: Adding support for some kind of export functionality is something that I wouldn't object to, but it's not anywhere near the top of my priority list. So feel free to send a PR, otherwise it won't happen.

PR #3569, opened 2026-08-25
feat: add export command to dump secrets as JSON

ChrisJr404:
This adds a `gopass export` command for #3074. It walks the store, decrypts each secret and writes them out as a JSON array with the password, body and attributes, so nothing gets dropped on the way out. You can pass one or more folder prefixes to narrow it down, and `--out` to write to a file (0600) instead of stdout. The output is plaintext so it prints a warning to stderr first.

I went with a generic JSON shape rather than a Bitwarden-specific one, figuring it is easier to transform into whatever target format than to bake one importer in. Happy to change the shape or add a confirmation gate if you would rather have something different.

Tests cover the whole-store export, prefix filtering and the `--out` file path. `make test` and golangci-lint are clean.

### DIFF
--- a/docs/commands/export.md
+++ b/docs/commands/export.md
@@ -1,0 +1,47 @@
+# `export` command
+
+The `export` command decrypts secrets and writes them to a portable JSON
+document. It is meant to help migrate a store into another password manager:
+the JSON keeps the password, the body and all key/value attributes so nothing
+is lost, and you can transform it into whatever format the target expects.
+
+By default the whole store is exported. Provide one or more folder prefixes to
+export only those subfolders.
+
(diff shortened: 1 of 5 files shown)


dominikschulz:
While I understand that some users might appreciate this feature, I'm a little bit concerned because it's a also a pretty big vector for (accidental) mis-use. A single command could lead to the entire store being stored in plain files.

I don't think this is something I want to add to `gopass` by default.

ChrisJr404:
That's a fair concern. A single command that can write the entire store out to plaintext JSON is a big footgun, and keeping that surface out of gopass by default makes sense. Closing this out; thanks for taking the time to look at it.

PR #3576, opened 2026-08-27
Prevent REPL command dispatch loop

### LINKED ISSUE #3491: Bug: Interactive REPL shell (`gopass>`) deadlocks completely (ignores Ctrl+C) on OTP fallback error
### Description

When running `gopass` in its interactive subshell mode (`gopass>`), executing the `otp` command against a secret that contains no OTP/TOTP data causes a complete terminal deadlock.

While running the exact same command directly from the native system shell (`zsh/bash`) gracefully exits with an error code, executing it inside the interactive `gopass>` environment freezes the interface permanently. Once frozen, `Ctrl+C` is completely ignored, forcing the user to hard-kill the terminal process.

### Steps to Reproduce

1. Launch the interactive gopass shell:

```bash
gopass

```

2. From the `gopass>` prompt, query an entry that has no OTP token configured:

```text
gopass> otp github.com/marekkowalczyk

```

3. Observe the exact-match fallback message, followed by the base32 decoding failure, followed by an absolute application lockup. Try pressing `Ctrl+C` to interrupt.

### Expected Behavior

The interactive subshell should catch the command failure, print the error message, and display a new `gopass>` prompt for the next command.

### Actual Behavior / Logs

```text
gopass> otp github.com/marekkowalczyk
Entry "github.com/marekkowalczyk" not found. Starting search...
✅ Found exact match in "websites/github.com/marekkowalczyk"
❌ Failed to compute OTP token for websites/github.com/marekkowalczyk: Decoding of secret as base32 failed."
<--- INTERACTIVE SHELL DEADLOCKS HERE — CTRL+C IS COMPLETELY IGNORED --->

```

### Environment

* **Gopass Version:** `1.16.1`
* **Go Version:** `go1.25.5`
* **OS:** macOS (`darwin arm64`)
* **Terminal Emulator:** Apple Terminal / Alacritty

jvasile:
The problem: The interactive shell recursively invokes the root command for each line. The active command is stored in a private context value, so reusing the outer context without filtering makes the nested command inherit itself as its parent and causes command dispatch to loop indefinitely after input is submitted.

The fix: This patch wraps the REPL dispatch context to hide active-command while preserving the application context. This avoids the infinite loop and allows command dispatch to execute normally.

This fix might also fix #3491.  On my system, I can use the patched code to do otp as expected.

### DIFF
--- a/internal/action/repl.go
+++ b/internal/action/repl.go
@@ -151,7 +151,7 @@ type completionSpec int
 // nested command invocation. This prevents infinite loops as the nested
 // invocation would otherwise inherit the command currently being executed.
 type replContext struct {
-	context.Context
+	context.Context //nolint:containedctx
 }
 
 func (c replContext) Value(key any) any {
--- a/internal/action/repl.go
+++ b/internal/action/repl.go
@@ -86,7 +86,7 @@ READ:
 		default:
 		}
 
-		if err := cmd.Root().Run(ctx, append([]string{"gopass"}, args...)); err != nil {
+		if err := cmd.Root().Run(replContext{Context: ctx}, append([]string{"gopass"}, args...)); err != nil {
 			continue
 		}
 	}
@@ -146,6 +146,21 @@ func unescapeEntry(s string) string {
 
 // completionSpec describes what candidates a command should complete against.
 type completionSpec int
+
+// replContext filters out urfave/cli's private active-command value from a
+// nested command invocation. This prevents infinite loops as the nested
+// invocation would otherwise inherit the command currently being executed.
+type replContext struct {
+	context.Context
+}
+
+func (c replContext) Value(key any) any {
+	if fmt.Sprint(key) == "cli.context" {
+		return nil
+	}
+
+	return c.Context.Value(key)
+}
 
 const (
 	completeNone       completionSpec = iota


dominikschulz:
This sounds reasonable, but would still like to understand the problem better. I haven't noticed any issues so far. Is this a recent regression?

dominikschulz (approved):

dominikschulz:
Ok, reproduced locally. The fix seems to work. Thanks a lot. Please add a "//nolint:containedctx" to make the linter happy.

internal/action/repl.go
```diff
@@ -147,6 +147,21 @@ func unescapeEntry(s string) string {
 // completionSpec describes what candidates a command should complete against.
 type completionSpec int
 
+// replContext filters out urfave/cli's private active-command value from a
+// nested command invocation. This prevents infinite loops as the nested
+// invocation would otherwise inherit the command currently being executed.
+type replContext struct {
+	context.Context
```
dominikschulz: Please add "//nolint:containedctx" to ignore this. I think we need the embedded context here.

dominikschulz (dismissed):
Thanks!

internal/action/repl.go
jvasile: Added.  Thanks.

jvasile:
It looks like this bug is an effect of upgrading urfave/cli from v2 to v3 on May 16 in commit 7c63ba09.

dominikschulz (approved):

PR #3577, opened 2026-08-27
fix(gpg): make identity selection deterministic

jvasile:
## The Problem

When selecting a user ID for displaying GPG info, a key might have multiple user identities.  We currently select the newest, but in the case of multiple identities with the same timestamp we select one according to map iteration order, which has no deterministic order.  This makes identity non-deterministic, which can trigger crude audit tooling looking for diffs in the output.

Multiple IDs with the same creation time does happen. Encountering this in the wild is how we discovered the bug.

One way this occurs is when a key has two UIDs added in rapid succession:

```
gpg --quick-add-uid KEYID 'Alice <alice@example.com>'; gpg --quick-add-uid KEYID 'Alice <alice@work.example.com>'
```

GPG gives us time down to the second.  If both operations occur within the same second, GPG can emit:

```
uid:u::::1700000000::...::Alice <alice@example.com>:
uid:u::::1700000000::...::Alice <alice@work.example.com>:
```

Both UIDs end up with the same parsed CreationDate.

## The Fix

Retain each parsed user's ID so identity selection has a stable tie-breaker when multiple UIDs share a creation timestamp.

### DIFF
--- a/internal/backend/crypto/gpg/colons/parse_colons.go
+++ b/internal/backend/crypto/gpg/colons/parse_colons.go
@@ -150,6 +150,7 @@ func parseColonIdentity(fields []string) gpg.Identity {
 
 	id := fields[9]
 	ni := gpg.Identity{
+		UID:            id,
 		Name:           id,
 		CreationDate:   parseTS(fields[5]),
 		ExpirationDate: parseTS(fields[6]),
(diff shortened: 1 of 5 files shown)


dominikschulz (approved):
Thanks!

dominikschulz:
Appreciate the steps to reproduce. Another one for the "should not happen, but does happen" book of bugs :)

PR #3578, opened 2026-08-27
Faster, more specific recipients command

jvasile:
This is three commits that add up to two changes that build on each other.

In the first commit, I added `FormatKeys`, which does what `FormatKey` does but replaces multiple calls by batching.  Then the second commit adjusts the `recipients` command to use `FormatKeys`.  That yielded an instant 10x speed using my `.password-store` and keyring!  

The third commit implements prefix selectors so you can `gopass recipients list path/to/subtree`.  Now you don't have to wade through a ton of stuff to get to the bit you want.

See the invididual commit messages for details.

(I also have an unpushed commit that uses this in other places FormatKey might get called multiple times, but those are going to be much smaller improvements than with recipients so doing the adjustment is more about code consistency than performance.  I can push it here or separately or not at all, as you prefer.)

### DIFF
--- a/docs/commands/recipients.md
+++ b/docs/commands/recipients.md
@@ -11,6 +11,7 @@ For the full team workflow reference, see
 
 ```
 $ gopass recipients
+$ gopass recipients list [prefix]
 $ gopass recipients add [--store=<store>] <recipient-id>...
 $ gopass recipients remove [--store=<store>] <recipient-id>...
 $ gopass recipients update [--store=<store>] [<recipient-id>...]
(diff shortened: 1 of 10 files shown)


dominikschulz:
@jvasile Sounds great, thanks. Feel free to push the pending commit as well.

dominikschulz (dismissed):

jvasile:
I pushed the additional uses of FormatKeys and added a couple blank lines to pacify the linter.  Thanks.

dominikschulz (approved):

PR #3610, opened 2026-09-15
fix(tempfile): avoid hdiutil deprecation warning on macOS 27

1995parham:
## Summary

On macOS 27, every gopass command that needs a secure tempdir prints a deprecation warning to stderr:

```
hdiutil: WARNING: 'hdiutil attach -nomount ...' is deprecated. Please use 'diskutil image attach --noMount ...' instead.
```

`pkg/tempfile` creates its ramdisk with `hdid -drivekey system-image=yes -nomount ram://32768`. `hdid(8)` is a thin wrapper around the hdiutil disk image interface, which macOS 27 deprecated. The attach still succeeds, but `mount` sets `cmd.Stderr = os.Stderr`, so the warning goes straight to the user's terminal on every `gopass edit`, `gopass env`, and anything else that goes through `tempfile.New`.

## Fix

Prefer `diskutil image attach --noMount --nobrowse ram://32768`, and keep `hdid` as a fallback for releases that predate the `image` verb. The probe's stderr is captured rather than passed through, so users on older macOS see no output from an attempt that did not end up mattering, and the outcome is memoized so the failed exec happens at most once per process.

The two commands format their output differently, so the device-node parsing had to change. `hdid` pads the node with spaces and tabs (`/dev/disk4          \t                               \t\n`) while `diskutil` emits it bare (`/dev/disk4\n`). The existing `strings.Split(out, " ")[0]` happens to work for `hdid` but would have kept the trailing newline on `diskutil` output and produced an invalid device node; `strings.Fields` handles both. `parseDev` is covered by a new table test using the real output of both commands.

Everything after the attach is unchanged — `newfs_hfs` and the `diskutil mount` / `unmountDisk` / `eject` calls are not deprecated and work as before.

Also corrects the ramdisk size comment: 32768 sectors of 512 bytes is 16 MiB, not the 32 MB the comment claimed. `newfs_hfs` confirms it — `Initialized /dev/rdisk4 as a 16 MB case-insensitive HFS Plus volume`.

## Testing

Verified on macOS 27.0 (build 26A428, arm64) with a small program calling `tempfile.New` and capturing stderr.

Before, on `master`:

```
TMPFILE: /var/folders/.../gopass-check-442827772/secret
OK
--- stderr: ---
hdiutil: WARNING: 'hdiutil attach -nomount ...' is deprecated. Please use 'diskutil image attach --noMount ...' instead.
```

After, with this change — stderr is empty:

```
TMPFILE: /var/folders/.../gopass-check-776066011/secret
OK
--- stderr: ---
```

`go test ./pkg/tempfile/...` and `golangci-lint run pkg/tempfile/...` both pass, and the full mount/write/unmount cycle was exercised end to end.

I could not test the `hdid` fallback path on a pre-27 machine — I do not have one to hand. The fallback is the previous code path, reached whenever the `diskutil` attempt returns an error, so a reviewer on an older release confirming that `gopass edit` still works there would be worth having.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_011xDya1SH2B8fmV7Mx8nDkd


dominikschulz:
Thanks. Code looks reasonable, but I want to test this before merging.

dominikschulz (approved):
Works for me. Thanks!

